PR #562, opened 2014-06-04
Set IE11 to "yes" for DeviceOrientation

dstorey:
Says partial refers to lack of `compassneedscalibration`. IE11 supports this: http://msdn.microsoft.com/en-us/library/ie/dn629487(v=vs.85).aspx


Fyrd:
Note has been added to explain limitations to support.

PR #715, opened 2014-09-03
Touch Events now supported IE11 mobile

jacobrossi:
http://blogs.msdn.com/b/ie/archive/2014/07/31/the-mobile-web-should-just-work-for-everyone.aspx

Note that status.modern.IE reflects this as "In Development" at the moment. I'll be updating that soon, but just haven't yet figured out how to best differentiate in our site that this feature is currently only supported on mobile.

### DIFF
--- a/features-json/touch.json
+++ b/features-json/touch.json
@@ -204,7 +204,7 @@
     },
     "ie_mob":{
       "10":"p",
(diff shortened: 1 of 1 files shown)


Fyrd:
Set manually to partial support with note. Once Windows Phone 8.1 Update is used widespread it can be changed to full support.

PR #753, opened 2014-10-02
HTML imports ready for Firefox 35

waldyrious:
See https://bugzilla.mozilla.org/show_bug.cgi?id=877072#c83


waldyrious:
From https://hacks.mozilla.org/2014/12/mozilla-and-web-components/:

> Mozilla will not ship an implementation of HTML Imports. We expect that once JavaScript modules — a feature derived from JavaScript libraries written by the developer community — is shipped, the way we look at this problem will have changed. We have also learned from Gaia and others, that lack of HTML Imports is not a problem as the functionality can easily be provided for with a polyfill if desired.

For back-reference, #822 reverts this.

PR #863, opened 2015-01-24
Fix resolution media query

ai:
According [MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/resolution) I add support data for old Opera and old Firefox.


ai:
@Fyrd Autoprefixer 5.1 will need this PR to clean this old prefixes.

Fyrd:
@ai These Firefox versions support the standard min-resolution: dpi format, which trumps their support for the non-standard `min--moz-device-pixel-ratio`. Can you change those back in your PR? Thanks.

ai:
@Fyrd only Firefox >= 16 supports `min-resolution: dpi` so I left `y` there. As I understand Firefox < 16 supports only `min-device-pixel-ratio` so I add `a x` with note there.

I didn’t understand what versions I need to fix :(.

Fyrd:
@ai In my tests Firefox 3.6+ support min-resolution: dpi. Do you have any test that indicates otherwise?

ai:
Very strange. MDN page says:

> -moz-device-pixel-ratio may be used for compatibility with Firefox older than 16

https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Media_queries

As I understand it means, that < 16 didn't support min-resolution (also this is new spec, I think it doesn't even exits in 3.6).

Fyrd:
Looking at https://developer.mozilla.org/en-US/docs/Web/CSS/resolution I'm seeing 3.5+. And, as mentioned, it's working in my tests. :)

ai:
Resolution page is a page of units, not the query. And units are exists in 3.5. Can I see a test, maybe there is some issue there? (Very strange that FF 3.5 supports `min-resolution` even without a prefix, because there is no spec in that time).

Fyrd:
http://tests.caniuse.com/?feat=css-media-resolution - The second test should be lime if `@media (min-resolution: 1dpi) ` matches, which according to BrowserStack screenshots works in Firefox 3.6:
http://www.browserstack.com/screenshots/f6508589b83765e44029978f00a0cf73f6d61523 (and I've confirmed it stays red in a browser without support)

Fyrd:
(sorry, didn't mean to close)

ai:
Awesome tests! :) What means yellow and green support? Can I look at tests source code?

Fyrd:
If you're seeing yellow under the browser column, that just means the current support for that browser is set to "partial". Currently for the source code just...view source. ;) At some point I plan on putting all the data on GitHub too so it's easier to see.

ai:
Yeap, seems like you are right. Firefox really supports `min-resolution` since 3.5 O_o

ai:
@Fyrd done, I update Firefox data.

Fyrd:
Thanks!

PR #864, opened 2015-01-25
Add crisp-edges value

ai:
I fill data aacording [MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#bc1)


ai:
@Fyrd this PR is important for Autoprefixer 5.1 too https://github.com/postcss/autoprefixer/issues/367

Fyrd:
@ai Just FYI I recently wound up making a number of changes to this based on my tests and research. In particular it appears that Chrome does not really support `-webkit-optimize-contrast` (like Safari does) and instead will support the `pixelate` value in upcoming versions.
https://github.com/Fyrd/caniuse/blob/master/features-json/css-crisp-edges.json

ai:
@Fyrd thanks for research

stuartpb:
html5please is looking at a description for this too: h5bp/html5please#229

PR #866, opened 2015-01-26
Link to whatwg for the xhr spec

MayhemYDG:
http://www.w3.org/TR/XMLHttpRequest2/:

>Work on this document has been discontinued and it should not be referenced or used as a basis for implementation. However, the Web Applications Working Group continues to work on XMLHttpRequest Level 1 and the WHATWG continues to work on XMLHttprequest.


Fyrd:
Thanks!

PR #873, opened 2015-01-30
Fetch

MayhemYDG:
http://www.google.com/moderator/#9/e=ae425&g=fetch

Chrome:
https://code.google.com/p/chromium/issues/detail?id=436770
I'm not entirely sure since which version we can enable fetch in the window via the platform flag. @jakearchibald might know better.
https://jakearchibald.github.io/isserviceworkerready/#fetch(request)
https://src.chromium.org/viewvc/blink?revision=186770&view=revision

Firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1039846
Available behind a flag since version 34 afaik:
https://developer.mozilla.org/en/Firefox/Releases/34


Fyrd:
Looking good, thanks!

PR #880, opened 2015-02-02
New feature: Referrer Policy

cramforce:
The feature has been in Chrome for a while and was recently adopted by Firefox. I'm currently lacking precise info about support in Safari. Will add later.


Fyrd:
Nice work, thanks!

PR #881, opened 2015-02-02
Note about the need for the lang attribute to make CSS hyphenation work

graste:
To make hyphenation work it is advisable to set the `lang` attribute on the HTML element as suggested [on quirksmode.org](http://www.quirksmode.org/blog/archives/2012/11/hyphenation_wor.html) and [adrianroselli.com](http://blog.adrianroselli.com/2015/01/on-use-of-lang-attribute.html).

```html
<html lang="en">
…
</html>
```

### DIFF
--- a/features-json/css-hyphens.json
+++ b/features-json/css-hyphens.json
@@ -209,7 +209,7 @@
       "9.9":"a x"
     }
   },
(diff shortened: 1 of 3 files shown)


Fyrd:
Thanks!

PR #882, opened 2015-02-04
Position:sticky is now "official", in WD

iamvdo:
Spec published yesterday: http://www.w3.org/TR/2015/WD-css3-positioning-20150203/


Fyrd:
Thanks!

PR #889, opened 2015-02-13
Update fetch.json

MayhemYDG:
Fetch shipped in Chrome canary 42 in the window context by default:
https://code.google.com/p/chromium/issues/detail?id=436770#c24
Also matched the support in Opera.


Fyrd:
Thanks!

PR #892, opened 2015-02-18
MutationObserver Chrome Android Updates

zhaoz:
https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
According to this, mobile support is aligned with the desktop support for mutation observer.


zhaoz:
Seems like chrome android browsers are not supported right now? https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#unsupported-changes

Tell me if this should be dropped and sent via feedback or if browser changes are coming.

MayhemYDG:
Caniuse only lists the current stable version of Chrome for Android for any features.

Fyrd:
Yeah, @MayhemYDG is correct.

zhaoz:
Why is that? Why the different policy here?

PR #894, opened 2015-02-19
Ruby enabled in Firefox 38.0

mpopp75:
Ruby is enabled by default in Firefox 38.0:

https://bugzilla.mozilla.org/show_bug.cgi?id=1039006


Fyrd:
Thanks!

PR #903, opened 2015-03-05
Chrome 41 supports mix-blend-mode

MidnightDesign:
See https://www.chromestatus.com/feature/6362616360337408


Fyrd:
Updated manually. Thanks!

peter-mouland:
I'm not sure if this is 100% accurate.... 

The site now says that chrome supports mix-blend mode, but i found a number of property values to be lacking.  Here is a full example :

http://codepen.io/peter-mouland/details/MwdbMY/

Am I misunderstanding something??

peter-mouland:
I just realised that this is not supported for SVG elements within chrome, maybe worth mentioning on caniuse.com?

peter-mouland:
Without a background colour set on the parent container, the results between firefox and chrome are different.

The example given on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode) and [w3.org](http://www.w3.org/TR/compositing-1/#mix-blend-mode) doesn't specify that background colour (behind the 3 circles being blended) is required.

Plus, as a side, firefox displays artifacts around the png for a few of the properties where as chrome does not.

Are these points worth mentioning on caniuse.com?

PR #912, opened 2015-03-12
Svg files in IE9 don't animate with an example.

ReptarAzar:
To test: Open IE9 and check out [this example](http://samherbert.net/svg-loaders/).  cc @macgyver

### DIFF
--- a/features-json/svg.json
+++ b/features-json/svg.json
@@ -46,7 +46,7 @@
       "6":"p",
       "7":"p",
       "8":"p",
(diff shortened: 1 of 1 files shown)


macgyver:
nor IE10 for that matter

Fyrd:
Animated SVGs are already covered under [SVG SMIL animation](http://caniuse.com/#feat=svg-smil), I don't feel like this needs additional mentioning under the basic SVG support.

PR #913, opened 2015-03-15
Update css3-tabsize.json

SRGOM:
Win10 + IE11 (slow ring) has white-space working, and even completing in the F12 developer tools style editor


Fyrd:
This feature is for the `tab-size` property, not the `white-space` one. And `tab-size`does not appear to be supported in the IE technical preview.

dstorey:
Yes, we’re not working on tab-size right now.

PR #918, opened 2015-03-18
EdgeHTML.dll supports Touch Events in March tech preview

dstorey:
(caveat is touch screen is required, but that is kinda obvious)


Fyrd:
Whee!

PR #921, opened 2015-03-19
Update Web Animations status

birtles:


Fyrd:
Can you add some comment in the notes or numbered notes why this support is partial?

birtles:
For which platform? For Firefox the details are at: https://birtles.github.io/areweanimatedyet/ I can add a note pointing to that.

For Chrome some of the limitations are:
* Element.animate doesn't support single keyframes
* Animation.ready and Animation.finished Promises not available
* Animation.effect and Animation.timeline not exposed
* Animation constructor not implemented
* None of the effect interfaces are available
* document.timeline not implemented
* I don't think it supports animating attributes or additive animation but I haven't checked
* Uses old AnimationPlayer interface name

I could add a note saying "Partial support corresponds to basic support of playback control and element.animate interfaces"?

Fyrd:
Yeah, those changes sound good.

Fyrd:
Thanks much!

PR #926, opened 2015-03-26
font-smooth (based on MDN article and google research)

pkuczynski:


Fyrd:
@pkuczynski Looks good, though why do you have the "should not be used" messages in the notes? I don't see any problem with devs using those properties.

Fyrd:
Actually I'll just merge for now, though I intend to remove those comments unless there's a good reason for them. Thanks!

pkuczynski:
Well, I believe the comment should stay, as it is highlighted in MDN, this is non standard extension. Although it was part of a CSS spec, it was intentionally removed given the issues which has been encountered with it..

PR #930, opened 2015-03-28
Make WebP unofficial

dstorey:
WebP is unofficial and not on any standards track. It is controlled by Google.


Fyrd:
Fair enough.

PR #933, opened 2015-03-29
Add MQ4 Interaction media features, IME API and CSS Exclusions

dstorey:
pointer, any-pointer, hover, and any-hover.

Android stock browser needs checked if it is supported in latest versions, as I don't have access to Android.


Fyrd:
Very nice, thanks! They'll still be hidden by default at first as I do some verification, but hopefully can make them visible within the next few days.

dstorey:
Great thanks. Is there a way I can add test cases to tests.caniuse.com?

PR #937, opened 2015-03-30
text-size-adjust supported with prefix in Project Spartan

dstorey:
…at least according to the test.caniuse.com page.


Fyrd:
Hmmm...technically this is supposed to be for mobile browsers only. Some desktop browsers (like Firefox) acknowledge the property but afaict don't actually do anything with it.

dstorey:
Yeah, I didn't test too much to see how it handles it as wasn't sure when it suppose to kick in. Gathered it would be used for tablets et al on Windows 10 as we have a touch browser by default.

Fyrd:
Would you happen to have access to a tablet with Windows 10 on it? If so, could you try out http://tests.caniuse.com/tsa.html and see if clicking the button affects the font size?

dstorey:
I have a touch screen laptop at the office, Will run in tomorrow AM and let you know

dstorey:
Looks like it isn’t working on desktop, at least with a touch screen laptop, so this can be closed won't fix.

Fyrd:
Cool, thanks for checking.

PR #939, opened 2015-03-31
Project Spartan now supports filter functions (behind flag)

dstorey:


Fyrd:
Actually this needs to be `n d #2`. Currently there's not really a way to define partial support behind a flag.

dstorey:
Updated it

Fyrd:
Thanks!

PR #945, opened 2015-04-08
Spartan has partial srcset support

zcorpan:
https://twitter.com/gregwhitworth/status/582685630154440704


Fyrd:
@dstorey I believe you set this to `y`, do you think it's safe to say `srcset` will have complete support on launch or should we go with partial support for now?

dstorey:
@Fyrd  we’ve having some issues with srcset as the picturefill polyfill breaks, assuming supporting srcset means supporting picture too :/  https://github.com/scottjehl/picturefill/issues/478 Breaks major sites. So we're looking into what to do internally.  

We support srcset and currentSrc, but not sizes and picture, which I believe are separate items here? 

It looks like we have to regress out support and remove currentSrc due to Picturefill being broken https://github.com/scottjehl/picturefill/issues/478#issuecomment-90306153

If partial here means not supporting sizes and (now) currentSrc then we should set it to partial.

zcorpan:
Partial here means not supporting w/sizes.

@dstorey you could mark currentSrc as `[Replaceable]` until you also support `HTMLPictureElement` to fix the picturefill issue I think.

Fyrd:
Yeah, `sizes` is more srcset-related, `picture` is its own feature.

zcorpan:
@gregwhitworth FYI (see my comment above about `[Replaceable]`)

gregwhitworth:
@zcorpan We considered that option.

PR #947, opened 2015-04-10
Added an IE11 flexbox bug

runarberg:
IE11 will not apply flexbox on pseudo-elements
https://connect.microsoft.com/IE/feedbackdetail/view/1058330/ie11-will-not-apply-flexbox-on-pseudo-elements


Fyrd:
Thanks!

PR #948, opened 2015-04-13
Add Audio Tracks and Video Tracks

dstorey:
Cool way to add DVD-like extras and to improve accessibility for both
audio and video.

Basic tests: http://jsfiddle.net/dstorey/ny7hzgmm/
http://jsfiddle.net/dstorey/Lvhvzvv6/


Fyrd:
Cool, thanks!

PR #950, opened 2015-04-13
Better Portable Graphics (BPG) image format

pkuczynski:


Fyrd:
@pkuczynski My criteria for caniuse additions include support in at least one browser version. So since that doesn't seem to be the case for BPG files I'm afraid I can't accept this.

pkuczynski:
I understand. This is very fresh thing and I expect it will be picked up by browsers pretty quickly. I will resubmit my updated PR then :)

Issue #867, opened 2015-01-27
WebGL in Safari 8


dhritzkiv:
I noticed that WebGL in Safari 8 is listed as partial support, potentially because not all users had it enabled by default (it was enabled by an option). While this was true in Safari 7, AFAIK, all users in 8+ have WebGL enabled by default. Though, I very well may be mistaken.

pkuczynski:
Good point. Created a pull request for this change...

Fyrd:
Yup, this was merged.

PR #951, opened 2015-04-13
WebGL supported in Safari 8 by default (#867)

pkuczynski:
* https://en.wikipedia.org/wiki/WebGL#Desktop_browsers
* http://blog.ludei.com/webgl-ios-8-safari-webview/


Fyrd:
Thanks!

PR #953, opened 2015-04-15
IE10/11 don't support calc() in transform

tomsterritt:
https://connect.microsoft.com/IE/feedback/details/814380/


Fyrd:
Thanks!

PR #957, opened 2015-04-16
add notes for mathml on firefox

patrickkettner:
I am not sure if this warrants downgrading support to partial as well. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/math), firefox does not support the html5 notation for mathml until v 4

### DIFF
--- a/features-json/mathml.json
+++ b/features-json/mathml.json
@@ -22,7 +22,7 @@
     }
   ],
   "bugs":[
-    
+
   ],
   "categories":[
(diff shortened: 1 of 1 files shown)


patrickkettner:
from the same page, "MathML [on pre-firefox 4 browsers] must be served [with the mimetype] application/xhtml+xml"

Fyrd:
Thanks!

PR #962, opened 2015-04-22
Update css-clip-path.json

dubrod:
Detailed Example of clipping an image using clip-path polygon css and svg coordinates url(#) style.


Fyrd:
Thanks!

PR #963, opened 2015-04-23
Update bugs section of high-resolution-time.json

PublicStartup:
The High Resolution Time API has been found to enable a "Spy in the Sandbox" class of security vulnerability. Full PDF of security research paper is at http://arxiv.org/pdf/1502.07373v2.pdf

### DIFF
--- a/features-json/high-resolution-time.json
+++ b/features-json/high-resolution-time.json
@@ -22,7 +22,10 @@
     }
   ],
   "bugs":[
-    
+    {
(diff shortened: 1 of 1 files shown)


Fyrd:
@PublicStartup Can you reformat the "bug" object to consist just of a description? See [the sample file](https://github.com/Fyrd/caniuse/blob/master/sample-data.json) and note that you can use markdown to include a link. Thanks.

PR #964, opened 2015-04-24
Update bugs section of high-resolution-time.json

PublicStartup:
The High Resolution Time API has been found to enable a "Spy in the Sandbox" class of security vulnerability. Full PDF of security research paper is at http://arxiv.org/pdf/1502.07373v2.pdf
Also found an old Temp2 fork created last year by Fyrd containing Chrome and Opera about older versions of Chrome and Opera on Windows. Copied and pasted into this proposed update, then revised to "Some versions of ..."

### DIFF
--- a/features-json/high-resolution-time.json
+++ b/features-json/high-resolution-time.json
@@ -22,7 +22,10 @@
     }
   ],
   "bugs":[
-    
+    {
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

Issue #902, opened 2015-03-03
firefox 38 defaults to true for unicode-range support


patrickkettner:
`layout.css.unicode-range.enabled` is true on Firefox 38, but listed as unsupported (but works behind flag) on the site.

hebbet:
not sure if firefox 38 will really ship with unicode-range support as it is still an early release.

patrickkettner:
sure, but as of now its `y`. data can always be updated if something changes

patrickkettner:
also  -thanks for opening a PR!

patrickkettner:
closed since #911 merged

PR #973, opened 2015-05-10
update unicode-range support in firefox

patrickkettner:
continuing #902...

38 ended up actually turning off unicode-range in. I apologize for creating
data churn.

Testing 38, 39, and 40 on the test suite actually shows failures

image

as a result, I think the data should be reverted back


Fyrd:
Thanks!

PR #1569, opened 2015-05-14
Update contentsecuritypolicy.json

zack-lin:
Test on UCBrowser, make sure it uses #2, Supported through the `X-Webkit-CSP` header


Fyrd:
Thanks!

PR #1573, opened 2015-05-15
Added 'Upgrade Insecure Requests'

TonyZTan:
I have added the "Upgrade Insecure Requests" feature, which allows sites to request the transparent upgrade of HTTP resources to HTTPS using a HTTP header or a meta HTML tag. The Chromium link is https://www.chromestatus.com/feature/6534575509471232, and the Firefox documentation is at https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#upgrade-insecure-requests. Thanks.


Fyrd:
Thanks!

Issue #1564, opened 2015-05-12
Canvas .toDataUrl()


Fyrd:
Canvas .toDataUrl()

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+3__
Issue added by __peppe__ on __2014-12-04__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

pkuczynski:
Shouldn't that be closed too if you cancel my pull request?

Fyrd:
Yeah, this support is now included in http://caniuse.com/#feat=canvas

PR #1579, opened 2015-05-19
Canvas .toDataUrl() fixing #1564

pkuczynski:


Fyrd:
As this seems to be widely supported, almost matching support for `canvas` itself, I'd suggest just including support there and adding some notes. I'll use your info and take care of that. Thanks!

pkuczynski:
Well, I tend to disagree, especially when it comes to mobile. I wanted to help you and worked on a ticket from the Issues list. Feel a bit like I wasted the time. How shall we proceed in the future to avoid that?

Fyrd:
@pkuczynski Why is it wasted time if I included most of the information anyway? Besides it seems your information wasn't all entirely accurate, you had old Opera Mobile and UC browser with no support though they seem to support it just fine. In the future if you don't know and are unable to test, just set the value to `u` for unknown.

For future issues if it seems it may already be part of another feature I'd suggest just checking with me first, otherwise it should be fine. As I don't have the time to look into every single suggestion in advance there may be the occasional one that doesn't make sense to add.

pkuczynski:
Fair enough :)

Most of my commits is based on the Google research, not actual testing. In the future if I won't be able to find any valuable information I will fill it up with "u".

Issue #1550, opened 2015-05-12
'caniuse.com' should have small-caps (and other font variants?) broken out as a separate fea...


Fyrd:
&#39;caniuse.com&#39; should have small-caps (and other font variants?) broken out as a separate feature.

iOS versions &lt;= 4.3.5 don&#39;t render .ttf &amp; .svg web fonts when the CSS property &#39;font-variant:  small-caps&#39; is used.

---
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)

Moderator votes: **+3**
Issue added by **A fate  worse than Beth** on **2011-09-06**
To vote this issue up or down, simply include `+1` or `-1` in your comment.

vladimirmartsul:
+1

jensimmons:
Yes, all of these should be separate CSS properties:
```
font-variant-ligatures
font-variant-caps
font-variant-east-asian
font-variant-alternates
font-variant-numeric 
font-variant-position
```

RoelN:
+1

Fyrd:
These are all available now, e.g. https://caniuse.com/#feat=mdn-css_properties_font-variant-caps

PR #1588, opened 2015-05-21
font-feature-settings - updating links, adding some notes and keywords

pkuczynski:
In the light of adding font-variant-* #1550


Fyrd:
Looks mostly good, though I think note 1 should be put under the general "Notes" section since it isn't browser-specific and 3 is unnecessary as caniuse doesn't track supported prefixed properties after the prefix is dropped. Thanks!

pkuczynski:
Applied suggested changes.

Don't you think the file name should be font-feature-settings.json instead of font-feature.json?

In overall the file naming is quite odd and gives me headache :) Sometimes files are prefixed with css- like css-font-stretch, while in other cases its just font-feature. It would make life easier if you would create some file-naming conventions for contributors :)

Fyrd:
Great, thanks!

As for the file naming, yeah, it's a bit annoying that it's inconsistent. I don't want to just change the current ones though since that would break links pointing at support tables. At some point I'll make sure the old ones work as aliases, then it should be safe to rename them.

PR #1589, opened 2015-05-21
CSS font-variant-alternates

pkuczynski:
Part of adding all font-variant-* #1550


Fyrd:
Nice work, thanks! For future reference, support behind a flag should have a `d` (disabled by default). I'll go ahead and set that for Firefox 24-33.

PR #1592, opened 2015-05-21
Update notifications.json

AurelioDeRosa:
Web Notification API is now a CR: http://www.w3.org/TR/2015/CR-notifications-20150519/


Fyrd:
Thanks!

PR #1595, opened 2015-05-23
Add form-attribute.json

jakezatecky:
Added a new feature for the `form` attribute that can be utilized on various form input and buttons:

``` html
<form id="a">
  ...
</form>
...
<input type="password" form="a">
...
<button type="submit" form="a">Submit</button>
```

I was surprised to see this was not already recorded, so I decided to create a PR. Of course, if this should be added in a different manner, I can make the appropriate changes.

Great website!

http://www.w3.org/TR/html-markup/button.submit.html#button.submit.attrs.form
http://www.w3.org/TR/html5/forms.html#attr-fae-form


Fyrd:
Looks good, thanks!

PR #1601, opened 2015-05-27
Fixed some truncated text.

phistuck:
Speculatively continued it. Alternatively, it can just be removed.


Fyrd:
Yeah, looking at the [original](https://github.com/Fyrd/caniuse/commit/68ded018ca2d0449f25bbdb8940acb30e286d7b2) that is indeed correct (not sure what happened)

Thanks!

PR #1605, opened 2015-05-28
Add note about Chrome using old version of spec

dstorey:
There is site compat issues due to Blink supporting old version of spec that devs are coding to. Calling out difference here.

Not sure about older phones like Blackberry, but suspect they use old version of spec too.


Fyrd:
Hm, okay. Thanks!

PR #1606, opened 2015-05-28
::placeholder and :placeholder-shown not the same selector

yisibl:
1. Add css-placeholder-shown.json
2. Fixed ::placeholder descriptions

See also: 
https://github.com/postcss/autoprefixer-core/commit/54dab9f3cd045c89d3e965f7d7d61f810ea513f7


Fyrd:
Hmm, I normally prefer not to include features that don't have any support at all, but I suppose I'll allow it in this case to make things clearer. Also good to know ::placeholder has been specified now, I know it didn't used to. Thanks!

yisibl:
:+1: I think as long as there is a browser to support, we can increase.

PR #1608, opened 2015-05-29
Scroll Snap Points enabled by default in Fx39+

mpopp75:
In Firefox 39+, CSS Scroll Snap Points are enabled by default.


Fyrd:
Cool, thanks!

Issue #1248, opened 2015-05-12
HPKP is draft for security 'certificate pinning' through a new HTTP header.  This protection...


Fyrd:
HPKP is draft for security &#39;certificate pinning&#39; through a new HTTP header.  This protection reduces the likelihood of successful MITM attacks against TLS through rogue or compromised CAs.

https://tools.ietf.org/html/draft-ietf-websec-key-pinning-21

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+6__
Issue added by __s__ on __2014-10-30__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

hthole:
+1

Fyrd:
Now available at http://caniuse.com/#feat=publickeypinning

PR #1612, opened 2015-05-30
Add HTTP Public Key Pinning, Change Upgrade Insecure Requests Status

TonyZTan:
I have added data for HPKP (Public Key Pinning) and changed the status for Upgrade Insecure Requests. The data for HPKP support for Safari, Opera, and Android is limited. This should be related to issue #1248.


Fyrd:
Looks good, thanks!

PR #1641, opened 2015-06-05
Add "custom properties" as keyword for CSS variables

tzi:
Hi!

I often looking for "custom properties" instead of "variables".
Perhaps we could add it to the keywords?

Cheers,
Thomas.


Fyrd:
Good idea, thanks!

PR #1650, opened 2015-06-09
Added a list of unsupported `calc()` properties for Firefox

jonkoops:
A number of CSS properties in Firefox do not support using the CSS
`calc()` inside of them. This commit adds the exceptions to the list of
known bugs for `calc()`. The related bug report can be found
[here](https://bugzilla.mozilla.org/show_bug.cgi?id=594933).


Fyrd:
Thanks!

PR #1651, opened 2015-06-09
Added an .editorconfig file for consistent code style

jonkoops:
Added an .editorconfig file to ensure editors adhere to the standards
used for JSON files. This is mostly to prevent contributors from
accidentally adding JSON indented with tabs or adding newlines to the
end of a file.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*.json]
+charset = utf-8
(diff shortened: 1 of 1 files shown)


Fyrd:
I'll merge this it may be slightly helpful, but note that the JSONs are often overwritten by my build process which will always automatically use the same format.

PR #1652, opened 2015-06-09
Strict Transport Security is now enabled for IE11

mkurz:
See here: https://blogs.windows.com/msedgedev/2015/06/09/http-strict-transport-security-comes-to-internet-explorer-11-on-windows-8-1-and-windows-7/


Fyrd:
Thanks!

PR #1657, opened 2015-06-11
Safari still does not support CORS for <video> in <canvas>

dconnolly:
Early testing of iOS 9 is not hopeful either:
https://bugs.webkit.org/show_bug.cgi?id=135379


Fyrd:
Thanks! I don't think this is a common enough use case to make it worth partial support though, so I'm going to change it to just being a bug.

dconnolly:
Works for me :+1:

PR #1668, opened 2015-06-18
Chrome & Opera supports "word-break: keep-all"

kiding:
[_Chromium_ added support for `keep-all` property](https://codereview.chromium.org/1094863007), and that makes _Chrome_ and _Opera_ a full support on `word-break`.


Fyrd:
Confirmed, thanks!

PR #1671, opened 2015-06-19
Remove note about Chrome for iOS bug now that it has been resolved.

anthonyryan1:
Opposite of commit 861c59c11b95aac856f7, since this was a bug rather than a case of partial support it seems like it should be removed now.

See https://code.google.com/p/chromium/issues/detail?id=396076 for more information.


Fyrd:
Thanks!

PR #1684, opened 2015-06-25
Add Content Security Policy Level 2

dylantack:
The source for this data is mostly detailed testing using browserstack.  Everything shown in "current aligned" browsers should be covered - I have not attempted to test rare browsers such as Blackberry.


Fyrd:
Thanks! May I ask what the tests are you used for this data? If could could share them that would be much appreciated.

dylantack:
Thanks for merging! The tests I used are at http://csp.metaltoad.com/.  They are pretty crude, unfortunately. I stumbled into http://tests.caniuse.com/, if there's a way to contribute there I'd be happy to improve the tests.

BTW I got really inconsistent results when generating screenshots on browserstack, I had to use the interactive browsers to get consistent behavior.

Fyrd:
Great, thanks! I currently don't have a proper way to accept tests unfortunately, I actually like to write them myself to improve my own understanding of the feature. :) I'll take a look at what you have there and see if I can include those myself if you don't mind. Thanks again!

Issue #1545, opened 2015-05-12
i would like to see the comparation of ZOOM style between browsers. Thank you so much!


Fyrd:
i would like to see the comparation of ZOOM style between browsers. Thank you so much!

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+3__
Issue added by __thaodns__ on __2012-12-06__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

pkuczynski:
This is duplicate of #1453, isn't it? So it should be closed...

Issue #1453, opened 2015-05-12
zoom  Originally IE only, not recommended to use elsewhere (it's better to use CSS Transforms). ...


Fyrd:
zoom

Originally IE only, not recommended to use elsewhere (it&#39;s better to use CSS Transforms). Still would be nice to list.

Sources:
http://msdn.microsoft.com/en-us/library/ie/ms531189(v=vs.85).aspx
http://css-tricks.com/almanac/properties/z/zoom/

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+3__
Issue added by __diggie__ on __2014-04-20__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

pkuczynski:
I am working on it...

Fyrd:
Now available at http://caniuse.com/#feat=css-zoom

PR #1691, opened 2015-06-27
Adding CSS zoom property

pkuczynski:
Fixing #1453, #1545, #1285, #1053


Fyrd:
There's a few issues here, the spec link is wrong, the spec status is wrong and there's a link entry without values.

If there's not a specification that includes this property I suggest using the MSDN link there.

pkuczynski:
I pushed this too early, sorry! Fixes applied...

Fyrd:
Thanks! I'm currently on vacation so it may take a week or so before I'm able to verify the information and publish it to the sit.

pkuczynski:
No rush! Enjoy your holiday :)

Fyrd:
@pkuczynski Okay, it's on the site now. I should note that a number of corrections were needed, seems `zoom` is still supported in all newer IE and mobile IE as well as any other mobile WebKit browser. As I believe I have mentioned before, if you do not know the support, use the `u` value, not `n`.

pkuczynski:
Regarding IE, I read about it somewhere, but can't find it now. I should check it more deeply, sorry. Will try better next time. However which webkit browsers I skipped? I thought I marked all of them correctly?

Fyrd:
The BlackBerry browsers and UC browser are all WebKit-based (actually I think BB10 might use Blink). Again, I don't expect everyone to test those browsers, but they're more likely to match say a slightly older version of Safari than anything else.

pkuczynski:
Got it.

PR #1698, opened 2015-06-30
Add backdrop-filter data.

### LINKED ISSUE #1555: backdrop-filter
backdrop-filter

---------
Moderator votes: __+3__
Issue added by __rale__ on __2015-02-18__

jonnyscholes:
Note that the spec for this is in the Editors Draft phase - I couldn't see an 'ed' state for the status so I've made it 'wd'. Is that the done thing? 

If merged will fix #1555


Fyrd:
Thanks! Generally I consider editors drafts as unofficial, so "unoff" is the appropriate value to use. But I'll go ahead and take care of that.

jonnyscholes:
Ah cool good to know - will remember that in the future :)

PR #1701, opened 2015-07-03
Fix location-search json filename

MoOx:
This make crash the caniuse-api npm module :/

Can you please release this as a patch? Thanks in advance!


MoOx:
Note: I also just fixed the json which were invalid.

Fyrd:
@MoOx Thanks! Looks I (and the validator script) missed that one due to the missing file extension.

PR #1715, opened 2015-07-09
document.execCommand('copy'), tested on CBT

dandv:


dandv:
Opera Mobile 26 is legit:

image

Fyrd:
Actually cut, copy & paste commands are already covered by http://caniuse.com/#feat=clipboard

As for Opera Mobile 26, thanks, I'll update to that.

PR #1717, opened 2015-07-09
Create devicepixelratio.json

dandv:


Fyrd:
Only the latest Blink-based Opera Mobile is included, FYI. You can't include others.

dandv:
> Only the latest Blink-based Opera Mobile and Chrome for Android is included, FYI. You can't include others.

Which ones did you mean? Is this a new change? I saw in the latest commits apparently a global search and replace for op_mob 24 with 30. On the other hand, you [said](https://github.com/Fyrd/caniuse/pull/1715#issuecomment-120222707) you'll update "to" Opera 26 (not sure what you meant).

Anyway, is the removal of versions a technical limitation, or a business decision? I imagine a certain market segment of Android users don't update their apps automatically, so they might keep around the *pre*-latest Opera Mobile or Chrome for Android. (I know I don't, and my older Opera runs just fine, without some of the unneeded bells introduced in later versions.)

Anyway, I've spent the time testing the feature on extra devices, so I'm kindly asking you guys to patch the file and remove the unwanted versions. IMO, not showing certain versions should be a presentation decision, rather than a data model one, because if data about versions is present, it can be ignored by UIs or bots like [browserlist](https://github.com/ai/browserslist); if it's missing, it's missing.

Fyrd:
Sorry, I meant to say Opera 30, not 26.

There's a few reasons why only the latest versions of these browsers is displayed. The first is because unlike most desktop browsers it's near impossible to re-download an older version for testing. Secondly, the usage data from StatCounter doesn't include version numbers for mobile browsers, so I wouldn't have usage information to use for the older versions. Finally, less versions to include means less to keep track of. 

Admittedly, none of those are great reasons and I've been thinking of changing it to maybe have just the latest 3 or 4 versions rather than just the one. But it would take some time and work.

The reason you can't include versions is because the JSONs are synced with a database which only allows specific versions. So even if I were to merge them into the repo they'd be excluded next time the JSONs were generated. That's why the validator script ensures no extra versions are added.

Fyrd:
Gonna go ahead and merge this before it goes stale. Will make the fixes myself. If at some point the other versions can be included we'll refer to this original PR.

dandv:
Thanks!

PR #1726, opened 2015-07-14
More complete table for CSP 2.

dylantack:
I finished testing the remaining browsers. Please let me know if there's anything else needed to toggle "shown:true". The tests are run against http://csp.metaltoad.com/. Thanks!

### DIFF
--- a/features-json/contentsecuritypolicy2.json
+++ b/features-json/contentsecuritypolicy2.json
@@ -10,7 +10,7 @@
     }
   ],
   "bugs":[
-    
+
   ],
   "categories":[
     "Other"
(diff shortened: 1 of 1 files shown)


Fyrd:
Nice! I'll look through it and let you know.

Fyrd:
@dylantack Okay, I've gone ahead and published it on the site. For future reference, numbered notes should be combined when possible (I did this for the Chrome & Opera ones) and including the browser names & version numbers is not strictly necessary since the UI already highlights them. Sometimes I do include them for clarity though, so I'm fine leaving them.

Oh, one question...it seems Firefox for Android (38) should match Firefox 38 on desktop, when I run the tests it seems it does. Can you confirm?

Thanks again!

dylantack:
Whoops, you're right about FF38. Thanks for combining the notes. Some of those intermediate versions appear to have implemented CSP2 piecemeal, so things got a little messy.

Issue #998, opened 2015-05-12
Please add compatibility information on  "srcdoc" attribute of IFRAME.  Link to the spec


Fyrd:
Please add compatibility information on  &quot;srcdoc&quot; attribute of IFRAME.

Link to the spec: http://www.whatwg.org/specs/web-apps/current-work/multipage/the-iframe-element.html#attr-iframe-srcdoc

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+25__
Issue added by __Ivan__ on __2013-12-25__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

KevinMartin:
+1

Fyrd:
Now available at http://caniuse.com/#feat=iframe-srcdoc

PR #1730, opened 2015-07-15
Add iframe srcdoc attribute

KevinMartin:


KevinMartin:
Pertains to: #998

Fyrd:
Thanks!

PR #1747, opened 2015-07-22
data update

### LINKED ISSUE #1746: Incorrect known issue - svg data uri encoding
http://caniuse.com/#feat=datauri

Known issue - For SVG data URIs IE only supports base64 encoding.

This is not quite correct - see http://dbushell.com/demos/svg/utf8uri/test2.html

IE does support non-base64 encoded, but unlike other browsers it requires the characters to be uri encoded.

more detail: https://github.com/Compass/compass/issues/1460#issuecomment-65132746

lukeapage:
Fixes #1746

### DIFF
--- a/features-json/datauri.json
+++ b/features-json/datauri.json
@@ -23,7 +23,7 @@
   ],
   "bugs":[
     {
-      "description":"For SVG data URIs IE only supports base64 encoding."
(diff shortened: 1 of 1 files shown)


lukeapage:
See later comments on issue, I don't think this is complete.

PR #1754, opened 2015-07-25
In Chrome 44 text-overflow: ellipsis is ignored

arleytriana:


Fyrd:
@arleytriana The bug report you mention doesn't seem to actually match the issue you describe. One refers to box-sizing, the other to font-size. Can you either update the bug description or include a test case the font-size issue?

arleytriana:
@Fyrd the issue is also reproducible without setting the CSS property "box-sizing" to "border-box". That is why i didn't mention it in the description.

A minimized jsbin example of the issue without box-sizing: border-box: 
http://jsbin.com/vozuwo/1/edit?html,css,output

Fyrd:
@arleytriana Ah, okay. Looking at that test case though it doesn't really seem like a bug...and I'm seeing it happen in Firefox too. With that much space there's either room for just a character or two or only ellipses (which I'm seeing in Safari). Frankly having the characters there seems more useful.

Fyrd:
Closing this as I feel it's a non-issue.

PR #1758, opened 2015-07-27
Web speach API reorganisation

yukulele:
Separate Web Speech API in 2 parts:
- Speech Recognition API
- Speech Synthesis API


Fyrd:
@yukulele Thanks. Ordinarily I try to avoid renaming the ID (filename) to prevent broken links, but I figured this change made enough sense. I've also ensured the old link still works: http://caniuse.com/#feat=web-speech now redirects to http://caniuse.com/#feat=speech-recognition

PR #1761, opened 2015-07-28
Add direct link to Web Animations polyfill

samthor:
It's linked to via the related blogposts, but this brings it front and center.


Fyrd:
Thanks!

PR #1771, opened 2015-07-30
Update WebPlatform Docs visibilityState URL

robcolburn:
[WebPlatform Docs moved it](https://docs.webplatform.org/wiki/apis/timing/properties/visibilityState)


Fyrd:
Thanks!

PR #1778, opened 2015-08-03
Add feature for page transition event

### LINKED ISSUE #1777: Add page transition events
https://html.spec.whatwg.org/multipage/indices.html#event-pagehide
https://html.spec.whatwg.org/multipage/indices.html#event-pageshow

wturnerharris:
Fixes #1777


Fyrd:
Please use one of the categories according to the [contirbuting](https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md) file. (see also the Travis CI error). Probably this should be DOM & JS API.

wturnerharris:
@Fyrd fixed failing tests. Thanks for pointing me in the right direction.

Fyrd:
Thanks! Note that this will stay as shown:false for a while until I've verified the data and updated most of the mobile data.

PR #1789, opened 2015-08-09
Changing dpcm back to yes again

dstorey:
dpcm is supported, as per a previous pull request I made. The feature test fails as there is a bug (listed here) where values <1 don't fire. But in reality, as screens with less than 1dpcm don't exist, this isn't a bug you'd hit in the wild except for feature detects like this.

### DIFF
--- a/features-json/css-media-resolution.json
+++ b/features-json/css-media-resolution.json
@@ -11,7 +11,7 @@
   ],
   "bugs":[
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Ah, sorry about that. Must've forgotten that previous PR. Have updated my test now and can confirm that it passes.

PR #1796, opened 2015-08-10
Fix backdrop-filter and add filter() function

iamvdo:


Fyrd:
Thanks! Good correction and nice addition.

PR #1800, opened 2015-08-11
Add CSS element() function

### LINKED ISSUE #1407: CSS element() (part of CSS image level 4). That should be included since the site includes CSS Canva...
CSS element() (part of CSS image level 4).
That should be included since the site includes CSS Canvas Drawings which isn&#39;t even official (while CSS element() is).

---------
Moderator votes: __+4__
Issue added by __ntim__ on __2015-02-24__

iamvdo:
(fix #1407)


Fyrd:
Looking good, thanks!

PR #1818, opened 2015-08-18
Update pointer.json

mattbrundage:
Diction; remove reference to Nightly, as this feature has long since propagated to all release channels.


Fyrd:
Thanks!

PR #1828, opened 2015-08-23
Update font-loading.json

parasyte:
Font-loading events are enabled by default in Firefox starting with version 41
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1149381


yisibl:
Nice! :ghost:

PR #1857, opened 2015-08-30
Safari 9 beta doesn’t support unset currently

dstorey:


Fyrd:
Thanks, already covered.

PR #1858, opened 2015-08-30
Firefox only supports mouse input with pointer events.

dstorey:
touch can be enabled with flags, but only on Windows. https://hacks.mozilla.org/2015/08/pointer-events-now-in-firefox-nightly/ , https://wiki.mozilla.org/Gecko/Touch


Fyrd:
Thanks!

PR #1868, opened 2015-08-31
IE9 support for max-width with workaround

jmarceli:


features-json/minmaxwh.json
```diff
@@ -31,7 +31,7 @@
       "description":"Firefox does not support `min-height` on table elements or elements with `display: table`"
     },
     {
-      "description":"IE10 and IE11 do not appear to support overriding `min-width` or `max-width` values using the `initial` value."
+      "description":"IE9, IE10 and IE11 do not appear to support overriding `min-width` or `max-width` values using the `initial` value. Possible workaround is to use `none` instead of `initial`."
```
dstorey: Yeah, IE doesn‘t support Initial. There is another entry for initial on caniuse. I wouldn't call it a bug of max-width, just an unsupported post CSS2.1 value.

Fyrd:
Yeah, I agree with @dstorey. Since there's info on "initial" support now there's no need to expressly include it here, so I'll just remove the whole thing.

PR #1870, opened 2015-08-31
Correct URL of `<input type=tel>`

annevk:
Everywhere else the data files use "(type=...)" and using percent-encoding actually breaks the HTML Standard. See https://github.com/whatwg/html-build/issues/6.


Fyrd:
Thanks!

annevk:
Why does this not result in an update to https://raw.githubusercontent.com/Fyrd/caniuse/master/data.json?

Fyrd:
It will, but it doesn't happen as an automatic result of the PR being accepted. Should be updated soon.

annevk:
Thank you!

PR #1882, opened 2015-09-02
Update IE/Edge support

dstorey:
* writing-mode
* filters


Fyrd:
Thanks!

ai:
@dstorey you remove `x` from `writing-mode`, but here is [MSDN page](https://msdn.microsoft.com/en-us/library/ms531187(v=vs.85).aspx) with `-ms-wiriting-mode`. Also [MDN page](https://developer.mozilla.org/ru/docs/Web/CSS/writing-mode) tells that prefix is required.

Does this edit is correct? Our `x` was just missed accidentally?

I hope for soon answer and fix, because it lock Autoprefixer release, which is promised for tomorrow.

dstorey:
@ai Here is the info "Internet Explorer 8. The -ms-writing-mode attribute is an extension to Cascading Style Sheets (CSS) and can be used as a synonym for writing-mode in IE8 Standards mode."

When it was originally added prefixes were not a thing, so added prefixless. IE8 added the prefix too as the recommended way of writing it, while still supporting the unprefixed.

ai:
@dstorey I tested right in IE 11 on CodePen

This CSS does not work:

```css
body {
  -ms-writing-mode: vertical-rl;
  writing-mode: vertical-rl;
}
```

This CSS works:

```css
body {
  -ms-writing-mode: tb-rl;
}
```

dstorey:
@ai vertical-rl was only added in Edge.

you need writing-mode: tb-rl;

ai:
@dstorey ouh. What you think if Autoprefixer anyway will set `-ms-` prefixes, because of non-standard value and for easy next cleaning?

dstorey:
@ai. works for me. Only thing is it won't work for IE5.5–IE7, but you may not care about those at this point.

PR #1891, opened 2015-09-03
Added rel=preconnect resource hint

igrigorik:
Preconnect is available in FF39+ and Chrome 46+.


Fyrd:
@igrigorik Please check the integration errors on your PRs. Looks like you have additional versions included, version numbers must match the sample file exactly.

igrigorik:
@Fyrd I'm a bit confused.. The validator complains that `and_chr` (Android Chrome, correct?) has an invalid key for '38', and the only key I see in sample-data is 44? How do I indicate that something was added in M38 Chrome for Android?

Fyrd:
@igrigorik You can't. Only the latest is included. I'm considering changing that to be the latest 5 or 10 or something at some point but right now it's just the one.

igrigorik:
@Fyrd ah, I see. It'd be great to get more granularity on mobile browsers... since that's where much of the growth is happening nowadays. Updated all three pulls, lmk if there is anything else.

Fyrd:
Thanks!

PR #1892, opened 2015-09-03
added Client Hints: DPR, Viewport-Width, Width

igrigorik:
Shipped in Chrome 46.


Fyrd:
Thanks!

Fyrd:
Hey @igrigorik, I was working on this test for client hints but it doesn't seem to be working in Chrome 46 (currently Chrome Dev)
The test is here: http://tests.caniuse.com/client-hints.php

Basically just a page with the `<meta http-equiv="CH-Accept" value="DPR, Viewport-Width, Width">` meta tag and an iframe which will displays its request headers. I would have expected the DPR, Viewport-Width and Width headers to appear in that list, but they're not there. Am I doing something wrong?

igrigorik:
@Fyrd doh, just realized that my article had the wrong syntax: it should be content="..." instead of value="...". Fixed in https://github.com/google/WebFundamentals/commit/ac0fe163cac50bee747128c56c4a91a4a1024a3f.

That said, your example still won't work because we treat iframes as separate navigation contexts: each iframe needs to opt-in on its own. Here's an example link with images echoing back the advertised values: http://tests.caniuse.com/client-hints.php

Fyrd:
@igrigorik Ha, okay glad I could help there.

I would have thought the iframe to have its own headers work though...not looking at resources inside it, it *is* the resource.

Anyway, have made the change to "content" on that page and used an image and checked its headers in dev tools and still not seeing the expected headers... (and your example link is just my page again ;))

igrigorik:
@Fyrd doh, I meant to link to: https://client-hints.herokuapp.com/meta-optin.html

The correct syntax is: `<meta http-equiv="Accept-CH" content="DPR, Width, Viewport-Width">`

PR #1917, opened 2015-09-09
referer shipped in Edge

dstorey:
https://dev.modern.ie/platform/status/metareferrer/


Fyrd:
@dstorey Hmm, it does not appear to pass my test:
http://tests.caniuse.com/meta-referrer.php?content=no-referrer
The page has the `<meta name="referrer" content="no-referrer">` tag so following the link the resulting page should not include a referrer header, but in Edge 10240 it does.

dstorey:
Will check on who set it to shipped, and what the details are. I know it was listed as in development for a long time.

dstorey:
@fryd messed you with details TL;DR: we support the WHATWG version rather than the ED W3C spec.

Fyrd:
Thanks, I'll update it manually based on that info. I see the test passes when using the "never" value.

PR #1935, opened 2015-09-14
Add EC-3 and AC-3 codecs

dstorey:
…I wasn’t sure if they should be combined or separate. EC-3 is the successor in the same way AAC is a successor to MP3, which are listed separately here, but VP8 and 9 in the video world are combined (and H.265 is not included yet).

I’ve kept them together for now, but i can separate them if needed. Might be easier to describe the differences that way.


dstorey:
Test case: http://jsfiddle.net/9f3chfcf/2/ (or try demo at https://dev.modern.ie/testdrive/demos/dolbyaudioexperience/)

Fyrd:
Yeah, let's keep them combined for now. If other browsers eventually adopt one without the other than that may be a point at which they can be split. Thanks!

PR #1936, opened 2015-09-14
Safari should be partial for scrolling snap points

dstorey:
See failed tests on css3test (ignoring the properties no longer in the spec). Also failing tests on test.caniuse.com (but it isn't 100% clear to me if one value should be accepted for those properties or has to be two, so not left a note about those. Firefox passes the tests though.)


Fyrd:
Sounds fair. Thanks.

PR #1942, opened 2015-09-15
Hyphens property unprefixed in Firefox 43

mpopp75:
https://www.fxsitecompat.com/en-US/docs/2015/hyphens-property-has-been-unprefixed/


Fyrd:
Thanks!

PR #1943, opened 2015-09-15
Add SCSS workaround for iOS issue in viewport-units

BenMorel:
The [viewport-units](http://caniuse.com/#feat=viewport-units) page links to a [workaround](https://gist.github.com/pburtchaell/e702f441ba9b3f76f587) by Patrick Burtchaell for iOS issues.

I've created SCSS [mixins](https://gist.github.com/BenMorel/e9e34c08360ebbbd0634) based on his idea, that include a comprehensive list of iOS devices.

This basically allow you to support iOS 7 devices by just replacing:

    height: 50vh;

with:

    @include vh(50);

I thought it could be useful to link it on caniuse!


Fyrd:
Thanks!

PR #1944, opened 2015-09-15
maxlength: Add info about @Serabe's iOS 8 bug

cvrebert:
See https://github.com/Fyrd/caniuse/issues/1770#issuecomment-140523871


Fyrd:
Thanks!

PR #1945, opened 2015-09-16
appearance is now standardised, with the supported `none` and unsuppo…

dstorey:
…rted `auto` values.


Fyrd:
Looks good, thanks!

PR #1951, opened 2015-09-18
Update cryptography.json

vibornoff:
Add link to [_webcrypto-shim.js_](https://github.com/vibornoff/webcrypto-shim) — project aiming to fix bugs in prefixed webcrypto implementations in IE11 and Safari.


Fyrd:
Thanks!

PR #1957, opened 2015-09-21
Added bug description about selecting the text, styled with css ::first-letter pseudo-element

universome:
Added this bug: [https://bugs.webkit.org/show_bug.cgi?id=6185](https://bugs.webkit.org/show_bug.cgi?id=6185)


Fyrd:
Thanks!

PR #1958, opened 2015-09-21
Add more keywords to SVG fragment data

jelmerdemaat:
Item could not be found by the terms `svg`, which is even in the title, or `use`, which is an important part of this functionality.
Added these terms in keywords.

Cheers!

### DIFF
--- a/features-json/svg-fragment.json
+++ b/features-json/svg-fragment.json
@@ -234,7 +234,7 @@
   "usage_perc_a":0.12,
   "ucprefix":false,
   "parent":"",
-  "keywords":"fragments,sprite",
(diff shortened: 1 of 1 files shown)


Fyrd:
It actually looks at the title first...just need to scroll down and click the "show 5 more" or "show all 8" buttons. I'll merge this for the "use" entry, though I doubt it'll matter much since the word "use" is in a lot of descriptions already.

jelmerdemaat:
Ah, I see. I hadn't even noticed these buttons before. Maybe this can also be input for an improvement to the visibility of these buttons.

Thanks!

PR #1963, opened 2015-09-22
Hidden is supported on ie9

billrob:
My company uses ie9 as our company standard browser and "hidden" does work as expected on ie9.  We are running
Version 9.0.8112
Update Version 9.0.38


Fyrd:
Not in my tests. See for example http://codepen.io/Fyrd/pen/dYpobx

 If you have a specific testcase you can show me where it's working, please share.

billrob:
Thanks for your consideration.  I see the javascript that runs is making:

<div hidden=""> This should be hidden</div> 

element when it renders.  It should be 

<div hidden>This should be hidden</div>

It might be a ie9 DOM/JS interaction issue, but the straight hidden attribute does work.

I will create a full stand alone web page and send the link that shows hidden attribute working as expected.

billrob:
rats, it erased all my html snippets, forgot about git hub and markdown.  I'll be back later, but basically the js interaction is creating 'hidden=""' rather than just 'hidden' attribute.

PR #1968, opened 2015-09-23
Update fullscreen data to reflect latest spec

andyearnshaw:
The latest version of the spec states that a Promise instance should be returned, which is fulfilled on success and rejected on error.  None of the current implementations have this, however, so IE is demoted to only having partial support :wink:.


Fyrd:
Thanks!

PR #1974, opened 2015-09-25
update text-align-last support table

myakura:
* update Chrome as it will be enabled by default in Chrome 47
  (see https://chromium.googlesource.com/chromium/src/+/87ca955695 )
* remove IE notes from Edge
* add more Android Chrome support status
  (marked it as `n d #2` as there hasn't been public CrMo 47 build yet)


Fyrd:
Note that it is not possible to add versions. You'll need to remove them for the Travis CI build to pass.

myakura:
oops, missed the "Unsupported Features" section. my bad.
(hoping doing that soonish ;p)

Fyrd:
Thanks!

PR #1977, opened 2015-09-25
Mention Safari's lack of fullscreen stacking in Full Screen API known issues

andyearnshaw:
Just ran into this today when trying to set a video full screen inside an already – fullscreen element.  God dammit, Safari, as if your full screen stretching animation isn't bad enough!

(╯°□°)╯︵ ┻━┻


Fyrd:
Thanks for submitting!

PR #1983, opened 2015-09-27
Use container-based Travis infrastructure

cvrebert:
Should slightly increase the speed of the build.
See http://docs.travis-ci.com/user/workers/container-based-infrastructure/


Fyrd:
Cool! Thanks for finding that.

PR #1998, opened 2015-10-01
Update battery-status.json

alexeyten:
Firefox 43+ supports `navigator.getBattery()`

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getBattery


Fyrd:
Confirmed, thanks!

PR #2019, opened 2015-10-08
css-grid: Use correct note for Firefox

bfrohs:
The note listed for Firefox is currently one meant for IE.


Fyrd:
Nice catch, thanks!

PR #2023, opened 2015-10-09
note partial support for Service Workers in Firefox 44

mykmelez:
Firefox is currently listed as not supporting Service Workers in any version, including upcoming ones, but Mozilla's platform engineering team is actually planning to ship Service Workers in either 43 or 44 (for both desktop and Android).

The relevant bugs are:

* [Bug 1059784](https://bugzilla.mozilla.org/show_bug.cgi?id=ServiceWorkers-v1) - (ServiceWorkers-v1) [meta] Ship Service Workers on desktop on the release channel
* [Bug 1207262](https://bugzilla.mozilla.org/show_bug.cgi?id=1207262) - (ServiceWorkers-Android) [meta] Ship Service Workers on Android on the release channel

Note: those bugs suggest that it's going to ship in 43, but I attended the most recent Service Workers meeting two days ago, and it sounds like it will slip into 44, as there are some issues with the implementation on the 43 branch (currently on Firefox's "Developer Edition" channel) that may not be resolved in time for it to remain enabled when that version gets released.

So this pull request conservatively changes only the entry for Firefox 44 to indicate partial support for Service Workers.

It might also be worth noting that you don't have to set `dom.serviceWorkers.enabled` on Nightly or Developer Edition builds, as Service Workers are enabled on those builds, and they'll remain so, even if Service Workers get disabled for version 43 once it moves from Developer Edition to Beta.

But I'm not sure how to explain that without confusing users who are looking at the version-centric waterfall chart of support, so I haven't tried to do so. (Perhaps it could be a note on the last few versions, something like "Service Workers is enabled on the latest Nightly and Developer Edition builds.")


Fyrd:
Thanks! Yeah, maybe a note as you mention is worthwhile, I'll think about it.

PR #2032, opened 2015-10-14
Added bug with details element and box-sizing in Chrome

jochienabuurs:
When using the common box-sizing fix using "box-sizing: inherit;" on
details elements, children of the details element inherit a value of
"box-sizing: content-box;" even if the box-sizing value is explicitly
set to "box-sizing: border-box"


Fyrd:
Thanks!

PR #2036, opened 2015-10-16
safari only partial supports audio-api, fixes #2035

### LINKED ISSUE #2035: Support for Web Audio API in Safari should be marked 'partial'
Safari 8 and 9 offer only partial support for the Web Audio API (mainly just the playback portions). If you search for 'Web Audio API', Safari 8 and 9 incorrectly have green boxes, with small yellow minus signs indicating that it requires a 'webkit' prefix. The whole box should be yellow because Safari 8 and 9 don't support large portions of the Web Audio API, including getUserMedia() which is necessary to record audio.

hebbet:


Fyrd:
As mentioned at #2035 I'd like more information on what's not supported. Also, partial support should be accompanied with a note describing why it's partial.

Fyrd:
Closing this PR based on my comment at #2035. If other information about missing support in Safari comes to light I'll re-open (though it'll still need the notes to describe why it's partial)

Issue #1406, opened 2015-05-12
scroll-behavior


Fyrd:
scroll-behavior

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+4__
Issue added by __Janek__ on __2014-12-16__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

nikkyandrei:
+1

youngroger:
+1

shavounet:
+1

zgreen:
+1

joeybaker:
+1

fxedel:
+1

Swatinem:
+1

danez:
+1

Fyrd:
Now available at http://caniuse.com/#feat=css-scroll-behavior Thanks @danez !

PR #2044, opened 2015-10-21
Add CSSOM scroll-behavior

danez:
I added the initial file for scroll-behavior.

It is currently only supported in Firefox 36+.
In Chrome it is behind a flag and not yet fully implemented.

This relates to #1406


Fyrd:
Looks good, thanks! Can you change the "a d" values to be "n d" instead? Bit confusing, I know, but since there's no support by default it needs to be an "n".

danez:
I updated the PR to use "n d" instead of "a d". I also tested with opera and it is the same as chrome so i also added "n d" for opera.

Fyrd:
Thanks!

PR #2049, opened 2015-10-23
preconnect is supported by Chrome 46+

igrigorik:
Support is enabled on all platforms, see: https://www.chromestatus.com/feature/5560623895150592


Fyrd:
Thanks!

PR #2059, opened 2015-10-30
Edge dropped support for EOT fonts

miguelsousa:


Fyrd:
Oh wow, totally missed that. Interesting, thanks!

PR #2075, opened 2015-11-06
Firefox 39 and 42 don't prevent alpha chars input

joseluisq:
Firefox 39 in Mac OSX by default does not prevent input alpha characters and Firefox 42 validates only but It doesn't disable alpha keys input.

https://jsfiddle.net/fht19Loe/1/


Fyrd:
Thanks! This might actually be allowed by the spec, not sure, but I agree it's good to have as a known issue.

joseluisq:
Yeah I think the same. In fact as known issue this should be to provided by the spec.

PR #2103, opened 2015-11-22
Add data for cursor: grab and grabbing too

ai:
We have `css3-cursors` for common values and `css3-cursors-newer` for zoom values. But `grab` and `grabbing` values have different support table.

I used [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) and recheck its data in current browsers.


ai:
@Fyrd I need this PR for Autoprefixer, because it generates wrong prefixes right now https://github.com/postcss/autoprefixer/issues/563

PR #2124, opened 2015-12-03
Update multicolumn.json

mattbrundage:
Changed Chrome platform status ID, as it's more representative of the work that Blink needs to get to full support for this feature.


Fyrd:
Thanks, I went ahead and just added the ID to the existing one (the list can be comma-separated)

PR #2125, opened 2015-12-03
Update http2.json

ADNewsom09:
Edge and IE also only support https http/2 traffic, see https://msdn.microsoft.com/library/dn905221(v=vs.85).aspx


Fyrd:
Thanks!

PR #2137, opened 2015-12-08
Update Chrome support

ahmadalfy:
I just found that CSS variables works on Chrome v48.
This is the complete information from my Chrome://version:
48.0.2564.22 (Official Build) dev-m (64-bit).
It is working there fine


Fyrd:
Thanks, confirmed. Updated manually to avoid dealing with merge conflict

PR #2141, opened 2015-12-11
Added requestAnimationFrame Android Chrome data

phistuck:
The data was previously automated to always show the last Chrome for Android version (huh?), so I fixed it since Chrome for Android should have the same support for this as the desktop versions.
Also added a missing backtick in a note.

### DIFF
--- a/features-json/requestanimationframe.json
+++ b/features-json/requestanimationframe.json
@@ -225,7 +225,33 @@
       "33":"y"
     },
     "and_chr":{
-      "47":"y"
+      "18":"a x #1",
+      "25":"y",
+      "26":"y",
+      "27":"y",
+      "28":"y",
+      "29":"y",
+      "30":"y",
+      "31":"y",
+      "32":"y",
+      "33":"y",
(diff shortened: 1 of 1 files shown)


phistuck:
*always show only the last

The Travis CI error is not clear to me. Chrome for Android never shipped versions 19 - 24, but it did ship version 18 and versions 25+, so how am I supposed to represent that?

Fyrd:
You can't actually add versions to caniuse data, as mentioned in the [contributing docs](https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md). I plan to include older Chrome for Android versions at some point, but it's not currently possible via PR.

phistuck:
Will a note be acceptable ("Contrary to what is shown above, this is supported since Chrome for Android 25.")?

Fyrd:
I don't think that would make any sense, unlike some other sites, caniuse just shows which browser versions have support, it doesn't make any claims as to what versions support is started with.

PR #2168, opened 2015-12-22
Safari 9 word-break: keep-all Support

kiding:
Safari 9 and iOS Safari 9 ([bug 123782](https://bugs.webkit.org/show_bug.cgi?id=123782)) started supporting word-break: keep-all, making it fully supported by all major browsers.


Fyrd:
Doesn't seem to work in my test...do you have a demo that shows it working?

kiding:
screen shot 2015-12-24 at 10 01 26 am

I believe it's one of the properties intended only for CJK(Chinese-Japanese-Korean.) You can check out [the example on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break).

yisibl:
Yes! Safari supported!

Fyrd:
Alright, thanks! Can confirm now, guess I need to update my test.

yisibl:
@Fyrd Thanks!

PR #2180, opened 2015-12-29
10up Open Sources IE 8 and 9 Support for Flexbox

maheshwaghmare:
Added - 10up Open Sources IE 8 and 9 Support for Flexbox


Fyrd:
Thanks!

PR #2181, opened 2015-12-30
Edge 12 and 13 do not support touch events

Compynerd255:
Can verify that Edge 12 and 13 do not support touch events. By default, if "touch-action" CSS property is left at default, Edge eats all touch manipulation of the page and sends little to any of it to the browser (except for the first split second, so clicks still work). If "touch-action" is set to "none", the pointer events work, and touches are turned to mouse events, but touches themselves do not respond. Could not verify the same for Edge 14


Studio384:
Touch Events are supported by Edge. On Edge for Windows 10 Mobile they are enabled by default, on Edge for Windows 10, they are disabled. But support is there.

Fyrd:
Yeah, touch events are supported, as per https://dev.windows.com/en-us/microsoft-edge/platform/status/touchevents

Compynerd255:
How do you enable touch events, then? I tested it on [this test page](https://mdn.mozillademos.org/en-US/docs/Web/API/Touch_events$samples/Example?revision=979019) and the touches were not recognized, even when I used DevTools to edit the "touch-action" CSS property of the hitbox. I was on a desktop computer using a Perceptive Pixel touch display.

Studio384:
about:flags

PR #2220, opened 2016-01-19
Partial support in Chrome 47, no support on iOS 9

fetis:
I suddenly discovered that caniuse data is incorrect. Bad surprise for me :(
So, I used this public link https://css-tricks.com/svg-fragment-identifiers-work/ for testing

My result:
iOS 9 no fragment identifiers at all!
Chrome 47 partial support

I attached screenshot with render results
img_0179
img_0180
how svg fragment identifiers work css-tricks - google chrome_022


Fyrd:
Have updated this manually. Apparently it seemed fragments in CSS backgrounds were working in my tests, but with further investigation this would only be the case if the same image had already been set by an image on the same page (which was a separate test). That actually may be a workaround to making it work, haven't really investigated further, but for now have updated the data and added a known issue with more info. Thanks!

fetis:
@Fyrd can you share your tests?
According my tests and link above theres's currently no support for fragment identifier at iOS at all. 

So, I included icon as
```
<img src="images.svg#icon" width="12" height="12">
```
and all sprite was shown, not the single icon

PR #2260, opened 2016-01-29
Web Animation API enabled by default in Fx47

mpopp75:
Web Animation API enabled by default in Firefox 47 as of January 29th.


Fyrd:
Nice, thanks!

PR #2288, opened 2016-02-07
menu: +popup type in Chrome 41 to 47, +partial support in Firefox for Android

cassiodoroVicinetti:
[About popup type in Chrome](http://stackoverflow.com/a/28858365/1267803)

About nested menus in Firefox for Android:

1. go to http://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_menuitem
2. hold pressed inside the yellow box <-- the context menu appears
3. press "Share on..." <-- the sub menu does not appear

### DIFF
--- a/features-json/menu.json
+++ b/features-json/menu.json
@@ -134,13 +134,13 @@
       "38":"n",
       "39":"n",
       "40":"n",
-      "41":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #2295, opened 2016-02-10
Update audio-api.json

vprigent:
We are using the audio api on android with devices from 4.0 to 5.1, it's working on every android >=4.4.4

EDIT > My bad, the project worked on Chrome, not as a native application


Fyrd:
That would be in "Chrome for Android" then. Android 4.4.* actually refers to the included browser engine which is separate from Chrome, which does not support the web API in my tests.

PR #2302, opened 2016-02-14
Normalize ECMAScript "spec" url

takenspc:
Some of "spec" urls of ECMAScript entries are inconsistent or outdated.

This PR normalises "spec" urls of ECMAScript entries.
More specifically, it makes all "spec" urls reference official spec either

- http://www.ecma-international.org/ecma-262/5.1/
- http://www.ecma-international.org/ecma-262/6.0/


Fyrd:
Thanks!

PR #2303, opened 2016-02-14
Normalise IETF "spec" url 

takenspc:
Some of "spec" urls of IETF entries are inconsistent.

This PR normalises "spec" urls of IETF entries.
More specifically, it makes all "spec" urls reference html version

- https://tools.ietf.org/html/


Fyrd:
Thanks!

PR #2321, opened 2016-02-25
XHR: add issue about setting timeout in IE

jalan:
In Internet Explorer the `timeout` property may only be set after calling `open` and before calling `send`. Chrome and Firefox each support setting `timeout` before `open` and after `send`.


Fyrd:
Thanks!

PR #2331, opened 2016-02-29
 Noted lack of support for nested media queries

Ceremony64:
Part of the official standard: https://www.w3.org/TR/css3-conditional/#processing
Works in Firefox and Chrome.
Does not work in IE11 and below.
Not sure if Edge and other browsers support this.


Fyrd:
This is already noted under http://caniuse.com/#feat=css-mediaqueries so it seems unnecessary to mention it under every other media query subfeature.

PR #2336, opened 2016-03-01
Properties don't work on Chrome

madvik:
I test on v48 but not before


Fyrd:
Seem to work fine to me. Can you provide a specific example of a property not working?

madvik:
Have you an inline exemple ? Thanks

Fyrd:
My tests are here: http://tests.caniuse.com/?feat=background-img-opts

madvik:
OK, thanks, I don't see what I talk about.
Here an exemple works on FF and not in Chrome : 
```
<select id="companytype" class="selecticon" name="companytype">
<option value="all">--- All Types ---</option>
<option value="1" style="background: transparent url(farm.png) no-repeat 0 0;">Farm</option>
<option value="3" style="background: transparent url(packhouse.png) no-repeat 0 0;">Packhouse</option>
```

Fyrd:
Ah, so those are being applied to `<option>` elements in a `<select>` list. I believe there's no defined way for browsers to apply styles on such elements, so that doesn't need to be considered partial support.

madvik:
The code I pasted works fine on Firefox !

PR #2338, opened 2016-03-04
Updated known bugs

jo-soft:
Extended known bug  for link elements.


Fyrd:
A version number would be nice, but for bugs I don't mind not knowing exactly.

PR #2348, opened 2016-03-09
Firefox workaround for no background-position-x/y

andyearnshaw:
Many people don't realise that background-position-x/y is achievable in Firefox, so I thought I'd add a note and some self-promotion by linking to one of my Stack Overflow answers from a while ago.  If that's an issue, feel free to remove it but I think it's helpful in any case.


Fyrd:
Cool, thanks!

PR #2377, opened 2016-03-22
multicolumn layout is unprefixed since Chrome 50

4thana:
Chrome 50 is currently in beta, the change was committed on Feb 18:
https://bugs.chromium.org/p/chromium/issues/detail?id=492297


4thana:
It still doesn't support break-before, break-after, break-inside though, but that will be fixed later in this bug: https://bugs.chromium.org/p/chromium/issues/detail?id=223068

4thana:
I also checked Opera and updated the PR.

Fyrd:
Thanks!

PR #2380, opened 2016-03-24
Update referrer-policy.json for Safari 8

public:
Safari 8 (possibly others, can't test) only supports the old set that Edge uses too.


joshma:
Safari 9 only supports the old set as well

Fyrd:
Thanks, updated manually to include other versions.

PR #2383, opened 2016-03-26
Fixed Incorrect CS Grid Layout Data

phistuck:
Chrome and Firefox still have not enabled it by default. I wonder where you found this data...

For Chrome -
https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/platform/RuntimeEnabledFeatures.in&q=grid%20file:WebKit%20file:%5C.in&sq=package:chromium&type=cs&l=49
"=experimental" means you must use the `--enable-experimental-web-platform-features` flag.

It might be ready later this year, see -
https://groups.google.com/a/chromium.org/d/msg/blink-dev/y221wJxoh84/qU_el0JRAwAJ

Mozilla bug and platform status -
https://bugzilla.mozilla.org/show_bug.cgi?id=1217086
https://platform-status.mozilla.org/#grid-layout


Fyrd:
Thanks!

You're right about Chrome...someone informed me it was enabled by default in upcoming builds but I guess I failed to verify the truth to that.

For Firefox, originally https://developer.mozilla.org/en-US/Firefox/Releases/45 simply stated that it was enabled by default, only later did they amend that to point out it's only in nightly & dev editions.

PR #2459, opened 2016-04-19
Add Same Site Cookie attribute as feature

LukasReschke:
This will be shipped with Chrome 51 as per the [Intent to Ship](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/csCtW3M3-wg). Firefox also [voiced public support](https://groups.google.com/forum/#!msg/mozilla.dev.platform/yEqC74IgnqQ/wIVQh4W2EAkJ). So it may be handy to have this to track which browsers support this security hardening over the time.


Fyrd:
Thanks! Would you happen to know how to easily test support for this?

LukasReschke:
Thanks for the merge. 🚀 

Easily is always such a thing when it comes to security features. At the moment I'm adding support for this to @ownCloud (https://github.com/owncloud/core/pull/24092), but since this is only targeted for master it won't be possible to use our demo instance to verify this.

In Chrome the feature can be quickly checked for existence by going to the Cookies resource overview and checking whether the `SameSite` column is in there:



I don't know whether other browsers will also add an overview like that. Another somewhat reliable way to test this is to `iframe` a page that uses either `lax` or `strict` cookies (make sure to not have any `X-Frame-Options` that may block this) and verify that in the iframe the user is indeed not logged-in and no cookies are sent.

Anyways, I'll keep closely track on adoption to this one as we intend to add this feature to our next release. So once other browsers keep adding support for this as well (something that we can easily check using an installed ownCloud) I'll update this entry 😄

DamirSvrtan:
Is there a reason why I can't find this feature on caniuse.com? I may be looking at it somewhat in a wrong manner, not sure. Thnx!

Fyrd:
@DamirSvrtan The feature data had not yet been reviewed, but it has now and is available at http://caniuse.com/#feat=same-site-cookie-attribute

DamirSvrtan:
@Fyrd thnx!

PR #2462, opened 2016-04-20
Add ie_id for FIDO U2F

nextPrime:
Microsoft tracks this feature in https://developer.microsoft.com/en-us/microsoft-edge/platform/status/fidou2f (Not currently planned).


Fyrd:
Thanks!

PR #2487, opened 2016-04-26
Use correct Firefox bug for CSS masks

flying-sheep:
The current one is about implementing only, the new one about implementing and shipping (Meta bug)


Fyrd:
Thanks!

flying-sheep:
NP :smile:

PR #2490, opened 2016-04-28
Update "known issue"

n3dst4:
The known issue with the popstate event not being fired when the hash changes affects Edge as well as IE. The official public-facing bug report has a new URL.


Fyrd:
Thanks!

PR #2496, opened 2016-05-05
Update online-status.json

edmondchui:
The following browsers support navigator.onLine but not online/offline events:

- Chrome 50.0
- Safari 9.1
- iOS Safari 9.3
- Chrome for Android 50


edmondchui:
My mistake. Was listening the events on document. Should listen on window.

Issue #1144, opened 2015-05-12
Add supported https protocols (like SSL 2, SSL 3, TLS 1.0, TLS 1.1 TLS 1.2) and Supported Cipher sui...


Fyrd:
Add supported https protocols (like SSL 2, SSL 3, TLS 1.0, TLS 1.1 TLS 1.2) and Supported Cipher suits.

More info see
- https://www.ssllabs.com/ssltest/viewMyClient.html 
- https://www.ssllabs.com/ssltest/clients.html

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+14__
Issue added by __andreus__ on __2015-03-07__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

tjconcept:
+1

ximex:
+1

nicdumz:
+1

nicdumz:
In particular, I came here to figure out which browser were supporting SHA2 certificates.

florianjacob:
+1

rugk:
Well SHA2 support is another thing, but the specific SSL/TLS protocols would be nice to add so :+1:.

frkay:
+1 especially since end of the road for TLS 1.0 has been set to June 2016 by PCI DSS 3.1, check:
https://www.pcisecuritystandards.org/documents/Migrating_from_SSL_Early_TLS_Information%20Supplement_v1.pdf
Safari desktop before 7, Internet Explorer 6, 7, 8, 9 (and even 10 when not properly configured), a lot of Android browsers do not handle TLS 1.1 and will be left in the dark.

frkay:
To check SHA-2 compatibility:
https://support.globalsign.com/customer/portal/articles/1499561-sha-256-compatibility

Koelli91:
+1

lorddaedra:
:+1:

RoelVliegen:
+1

Naddiseo:
+1

phanbu:
+1

TaoK:
+1

Fale:
+1

codyn329:
+1

equalsJeffH:
+1

smasher164:
+1

PeterDaveHello:
:+1:

CAfromCA:
+1

tyscorp:
+1

frkay:
[...]

PR #2500, opened 2016-05-06
Add ChaCha20-Poly1305 cipher suites

nextPrime:
This adds ChaCha20-Poly1305 TLS cipher suites to caniuse.com. This feature is part of #1144.
Spec: https://tools.ietf.org/html/draft-ietf-tls-chacha20-poly1305-04
This feature is supported in Chrome, Opera, and definitely not supported in IE or Edge. I am not sure about the support status on some platforms. To check if a browser supports this feature, visit https://www.ssllabs.com/ssltest/viewMyClient.html


Fyrd:
Thanks!

PR #2503, opened 2016-05-08
Update text-stroke.json

Yaffle:
Firefox 48 will support -webkit-text-stroke according to https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-stroke and https://developer.mozilla.org/en-US/Firefox/Releases/48


Fyrd:
Thanks!

PR #2515, opened 2016-05-17
Update svg-html.json

zoltan-dulac:
IE11 don't support `foreignObject` and the current json file does not reflect that fact.   This fixes that.


Fyrd:
Please fix the JSON: The string "part" is included. Also, for the newline please use `\n` instead of `<br>`

stevenvachon:
@zoltan-dulac ^

PR #2516, opened 2016-05-18
upgrade insecure requests status to cr

nextPrime:
Upgrade Insecure Requests is now a W3C Candidate Recommendation.
https://www.w3.org/TR/upgrade-insecure-requests/


Fyrd:
Thanks!

PR #2522, opened 2016-05-23
child-src support was added in FF 45

danez:
taken from here: https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#Browser_compatibility


Fyrd:
Thanks!

PR #2525, opened 2016-05-24
Brotli should now be enabled by default on Chrome 51+

MayhemYDG:
https://bugs.chromium.org/p/chromium/issues/detail?id=452335#c83


Fyrd:
Thanks!

PR #2526, opened 2016-05-24
Mark brotli as enabled for Chrome M51+

eustas:
https://chromium.googlesource.com/chromium/src/+/52b672b2462d6a3751a13187a1003e6fffdfbfbd


Fyrd:
Already added from another PR, but thanks!

PR #2529, opened 2016-05-25
Android Webview supports H2 on V50.

Garbee:
Fixed by https://bugs.chromium.org/p/chromium/issues/detail?id=329681 upstream.


Fyrd:
Thanks!

PR #2532, opened 2016-05-26
Added note about IE11 flex-basis bug.

damienmckenna:
IE11 requires a unit to be added to the 'flex-basis' attribute and to the third argument to the 'flex' attribute.


Fyrd:
Thanks!

PR #2533, opened 2016-05-27
More information about brotli experiment

eustas:
Brotli was enabled for all M50 users.


Fyrd:
What exactly does "Enabled via Chrome-Variations (92fb4fc2)" mean?

eustas:
"Chrome-Variations" is a technology that allows to control chrome flags.
Open "chrome://version" page and it will show "Variations" section with a list of XXXXXXXX-YYYYYYYY values. XXXXXXXX is a hash of flag and YYYYYYYY represents the flag setting.

eustas:
Ping?

Fyrd:
Okay, got it. Wonder if there's a better way to express that though since I think it would confuse most people like that. Also if it's not enabled by default the value still needs to be `n`. If it is, the value should be `y` without the `d`. (sorry for the delay, on vacation)

eustas:
Oh, sorry, didn't mean to disturb you on vacation.
So, it is now enabled by default, but still some users might need to restart their browser to apply this default.
Going to update PR soon.

Fyrd:
lgtm, thanks!

PR #2538, opened 2016-05-31
Update css-grid.json

mattbrundage:
Latest TP removes vendor prefix from all CSS Grid Layout properties: https://trac.webkit.org/changeset/200510


Fyrd:
Thanks!

PR #2553, opened 2016-06-06
Correct initial version of support on Android.

jpmedley:
https://www.chromestatus.com/features/5572795866021888


Fyrd:
These values are not the latest values of support, they just register the support of the latest version.

PR #2565, opened 2016-06-10
Added not for HTTP2 ALPN requirement

Toflar:
Chrome 51 only supports HTTP/2 over ALPN. There are many sources that state that, even nginx blogged about it: https://www.nginx.com/blog/supporting-http2-google-chrome-users/


Fyrd:
Thanks!

PR #2575, opened 2016-06-15
Flexbox: fixed spec definition of `flex` property

taylankasap:
`Initial` value in the latest [spec definition](https://www.w3.org/TR/css-flexbox-1/#flex-property) for `flex` property is `1 0 auto` not `0 1 auto`.


Fyrd:
Thanks!

PR #2577, opened 2016-06-16
iOS supports the pseudo class :placeholder-shown

BenjaminPoulain:
Someone forgot to add iOS WebKit when adding Safari.
The code is the same, iOS also passes all the tests for :placeholder-shown.


Fyrd:
Thanks!

PR #2580, opened 2016-06-17
IE11 review

SayTen:
* Removes child element focusing, testcase appears the same across all browsers.
* Removes pseudo-element as linked bug claims a fix.
* Removes long paragraph wrapping as test case renders like .top "flex-direction: row" and that works.
* Moves two height elements together as they appear to be the same issue.

### DIFF
--- a/features-json/flexbox.json
+++ b/features-json/flexbox.json
@@ -46,9 +46,6 @@
       "description":"In IE10 the default value for `flex` is `0 0 auto` rather than `1 0 auto` as defined in the latest spec."
     },
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Looks good. Thanks!

platosha:
> Removes child element focusing, testcase appears the same across all browsers.

This has to be reverted, as the issue is still present. If you click the span in the testcase, the console log says `"HTMLSpanElement"` for IE11 and `"HTMLDivElement"` for the other browsers.

PR #2584, opened 2016-06-17
Fix link to pinned tabs

rockhopper72:
For some reason the link to pinned tabs on Apple's developer website resulted in a 404. I corrected the link to pinned tabs so that it goes to the corresponding category on Apple's website


Fyrd:
Thanks!

PR #2589, opened 2016-06-19
Edge 14 allows user in increment/decrement number element

dstorey:
...using the value of the step attribute. Uses up/down arrow. There are no visual arrows as that is not part of the spec (ui is up to user agent) and that would be a terrible experience on touch based devices, of which Windows 10 is.


Fyrd:
Thanks, have manually updated the info on number inputs with more details.

PR #2591, opened 2016-06-19
Ambient light in Edge 14

dstorey:
https://developer.microsoft.com/en-us/microsoft-edge/platform/status/ambientlightevents


Fyrd:
Cool, thanks!

PR #2598, opened 2016-06-21
Edge 14 has added `none` back. cursor: url() never went away

dstorey:
All versions of Edge support cursor: url() just like IE did http://samples.msdn.microsoft.com/workshop/samples/author/dhtml/refs/cursor_a.htm (second format has been dropped, presumably because other browsers don't support it.)


Fyrd:
Thanks!

PR #2601, opened 2016-06-23
Update upgradeinsecurerequests.json

mattbrundage:
Support added in latest Safari TP:
https://trac.webkit.org/changeset/201679/trunk/Source
https://trac.webkit.org/changeset/201753/trunk/Source


Fyrd:
Thanks!

PR #2610, opened 2016-06-27
[ChaCha20-Poly1305.json] Update spec link

nextPrime:
ChaCha20-Poly1305 is now [RFC7905](https://tools.ietf.org/html/rfc7905). Also when will this feature be displayed on the website?


Fyrd:
Thanks. Currently based on the caniuse search logs there's very little interest in this feature so it hasn't been a priority, but I'll get to it eventually.

PR #2617, opened 2016-07-01
Update mediasource.json

pingyen:
Chrome started to support MediaSource 

by window.WebKitMediaSource from Chrome 23

and by window.MediaSource from Chrome 31.


Fyrd:
Thanks!

PR #2624, opened 2016-07-06
css-clip-path: Removed fixed bug in chrome

bfred-it:
It has been fixed 9 months ago https://code.google.com/p/chromium/issues/detail?id=350724


Fyrd:
@bfred-it Can you change it to where you leave an empty `bugs` array rather than removing it entirely? Thanks!

bfred-it:
Done

Fyrd:
Thanks!

PR #2629, opened 2016-07-07
push-api.json: add note #2 to Opera Mobile 37

PascalReintjens:
This adds note #2 ("Requires browser to be running to receive messages") to Push-Api: Opera Mobile 37 which I missed in my previous pull request


Fyrd:
Thanks!

PR #2635, opened 2016-07-11
Correct note about Push on Android

jpmedley:
I forgot, for mobile browsers are you showing the current stable version or the first version to get the feature? 

The version number you have is wrong either way. 50 is the version that got data payloads. 42 was the first version to support push. 51 is the current stable, and we're about two weeks away from 52 being stable.


Fyrd:
Thanks!

It's the current stable version. Sometimes I get a bit behind on that, thanks for pointing it out.

PR #2636, opened 2016-07-12
Corrected IE10 issue description

LayZeeDK:
Initial value of `flex` is `0 1 auto` as per [Flexbox Spec](https://www.w3.org/TR/css-flexbox/#flex-initial).

### DIFF
--- a/features-json/flexbox.json
+++ b/features-json/flexbox.json
@@ -43,7 +43,7 @@
   ],
   "bugs":[
     {
-      "description":"In IE10 the default value for `flex` is `0 0 auto` rather than `1 0 auto` as defined in the latest spec."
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks...an earlier PR had actually changed it from `0 1 auto` to `1 0 auto` based on what's at https://www.w3.org/TR/css-flexbox/#propdef-flex but did a bit more reading and it does seem that `0 1 auto` is correct as you mention. And it seems the latest [editor's draft](https://drafts.csswg.org/css-flexbox/#flex-property) does away with the confusing `1 0 auto` value.

LayZeeDK:
That is really confusing, @Fyrd. Let's go with the Editor's Draft for now :-)

PR #2643, opened 2016-07-13
Update push-api - Remove #2 on Opera Mobile & make meaning of note #2 clearer

PascalReintjens:
Since the merged pull request #2635 removed #2 on Chrome for Android, this means that Opera for Android 37 shouldn't have this note too, because it behaves exactly the same. Both need their app to be running but not necessarily visible for the user - they are both using a background service.

Note #2 wasn't really clear in which cases it applied. You could argue that running means the full visible browser or running already means a background-service.


Fyrd:
Thanks!

PR #2670, opened 2016-07-25
Broadcast Channel is implemented now

alippai:


Fyrd:
@alippai Thanks! Can you leave the Chrome ID? Leaving it results in a link under Resources, which may be of some value to some people.

alippai:
I thought it results in "flag enabled only". How can I change that?

Fyrd:
Nah, that happens when you use `n d` ("d" is "disabed by default").

Fyrd:
Fixed by @cvrebert's PR

PR #2691, opened 2016-08-03
[ChaCha20-Poly1305.json] Edge 14 doesn't support it

nextPrime:
Edge 14 still lacks the support of ChaCha20-Poly1305.


Fyrd:
Thanks!

PR #2708, opened 2016-08-09
Android browser 51+ does HTTP/2 only over TLS

scop:
As the Android browser 51+ is chromium based, I think it's a fair assumption that it supports HTTP/2 only over TLS. I've tested this locally with an Android 6.0.1 phone's browser and did not find any evidence of it doing cleartext HTTP/2 (no `Upgrade` header present, direct connection is HTTP/1.1). The browser identifies itself as:

Mozilla/5.0 (Linux; Android 6.0.1; LG-D802 Build/MMB29X; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/52.0.2743.98 Mobile Safari/537.36


Fyrd:
Sounds right, thanks!

PR #2717, opened 2016-08-12
Update calc bug link on IE / Edge

danielgamage:
The original issue at [connect.microsoft.com](https://connect.microsoft.com/IE/feedback/details/814380/) provided a dead link in the comments, and I'm unable to post clarification at that url. The updated issue is at [developer.microsoft.com](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/104773).

I specified `Edge < 14` in the notes as well, as I was able to reproduce up to v13 but not in v14.

For anyone happening across this PR looking for a fix for those older versions, [here's a fix for simple cases](https://stackoverflow.com/questions/21142923/ie-10-11-css-transitions-with-calc-do-not-work/21381301#21381301).

### DIFF
--- a/features-json/calc.json
+++ b/features-json/calc.json
@@ -25,7 +25,7 @@
       "description":"IE 9 - 11 don't render `box-shadow` when `calc()` is used for any of the values"
     },
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #2726, opened 2016-08-16
Adding Payment Request API

agektmr:
Adding Payment Request API entry.


Fyrd:
Thanks! Looks like another fairly straightforward one, so hopefully I'll get this one up soon. :)

PR #2729, opened 2016-08-16
Update text-emphasis.json

kkairri-jawp:
Firefox version 46 or later are supported “text-emphasis” as default.


Fyrd:
@kkairri-jawp Thanks! If there's no need for the prefix, it shouldn't have the `x` value. Can you fix that?

kkairri-jawp:
I didn't understand well what "x" means.
I fixed that errors, so please check it out.

Fyrd:
Great, thanks! FYI there's more details on those characters at the [CONTRIBUTING](https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md) docs.

PR #2735, opened 2016-08-18
touch-action property support is limited in Safari

bethge:
The `touch-action` support in Safari is limited to one out of four properties (not counting the default `auto`). The other properties are just as relevant such that I would recommend treating the support for touch-action as partial.


Fyrd:
Can you include a numbered note for this? Otherwise it will be unclear to people why the support is partial.

bethge:
_Updated_

Fyrd:
Thanks!

PR #2753, opened 2016-08-24
Update details.json

Yaffle:
info about toggle event support for UC Browser


Yaffle:
May be Chrome < 36 and Safari also do not support this event...

Fyrd:
Thanks!

PR #2767, opened 2016-08-30
Safari 10 does not support CSS Grid Layout

apple-web-evangelist:
Thank you for making this repo available so that we can provide updates.


Fyrd:
Thanks!

PR #2794, opened 2016-09-12
Node.innerText has been standardized in HTML

cvrebert:
See https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute


Fyrd:
Cool, thanks!

PR #2796, opened 2016-09-12
Correct the support status of text-emphasis of Firefox

upsuper:
FIrefox fully supports all values of text-emphasis-position from its initial implementation of text-emphasis. The note doesn't reflect the truth.


Fyrd:
Verified, thanks! However since there's full support the values can be `y` instead of just `a`. I have manually made this change now, so closing this PR.

PR #2819, opened 2016-09-19
Safari 9.2+ appears to support CSS transform

geelen:
Can't find any information about this fix being applied, but it seems to work.

Testcase: http://codepen.io/geelen/pen/ALWNdJ

Works on Safari 10
image

Works on iOS 9.2 in simulator (also confirmed 9.3 on my own device)
image

Failure case looks like:
image
(css changes colour but doesn't slide the toggle to the right)

Anyone with older Safari able to see the last version it was actually broken?


Fyrd:
Thanks! Looks like Safari 8 may have been the last version with this issue, as it also works in 9.1. Will update that.

PR #2825, opened 2016-09-20
`border-image-repeat: space;` supported in FF 50+

Schweinepriester:
See https://developer.mozilla.org/de/Firefox/Releases/50 =>
https://bugzilla.mozilla.org/show_bug.cgi?id=720531


Fyrd:
Thanks!

PR #2830, opened 2016-09-22
Add link to new MDN article about :focus-within

cvrebert:
https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within


Fyrd:
Thanks!

PR #2847, opened 2016-09-28
correct rellist information

patrickkettner:
its supported in a number of browsers not currently listed


Fyrd:
Hmm, I think this support should be partial, it's supported for `link` elements but not `a` elements. (Firefox & Safari support it for both)

patrickkettner:
You're the boss, but Id disagree/consider that a separate  data point. link rels are much, much more commonly used (and generally useful) than any of the anchor tag ones. But happy to modify as you wish.

Fyrd:
Yeah, I totally see where you're coming from...back in the day I would have agreed. However that's often lead to people pointing out the clear lack of support in a certain area (regardless of how much it's used) and it's easier to mark it as "partial" for technical reasons then to try to make arguments on how much something's used and where the threshold lies. 

Besides, this is Chrome we're talking about, it's probably just a matter of time before they get it on anchor elements anyway. :)

Thanks!

patrickkettner:
updated per request

Fyrd:
Thanks!

PR #2849, opened 2016-09-28
Update the link to the Edge bug for currentColor

stof:
The previous link points to a page marking the bug as closed because of the migration of Edge issues to their new dedicated issue tracker. This updates the link to point to the new issue tracker instead.


Fyrd:
Thanks!

PR #2851, opened 2016-09-28
Custom Elements (v1) supported in Safari TP

Schweinepriester:
See <https://webkit.org/blog/6969/release-notes-for-safari-technology-preview-release-14/>.

Was defensively and did not delete 'The API is also available in WebKit Nightly.', but that probably can be removed?

Furthermore I'm wondering about all the `"p"` in the file - isn't `"a"` for partial and `"n"` for no support?


Fyrd:
Yeah, agree that can be removed.
"p" technically means it can work via polyfill, though it's currently treated as "n" on the site (but used to have special treatment). Mostly it's a legacy thing that's proven too much work to keep up to date.

Schweinepriester:
I see!

The text has been removed :)

Fyrd:
Just updated my Safari TP and I notice the support is disabled by default, so this should change to a `n d #1`with a note pointing out it can be enabled via the "Experimental Features" developer menu.

Schweinepriester:
Ok, can do!

But before: Should we do the same for all "Experimental Features"?

ctdox1avyaas2br
(Stolen from <https://twitter.com/addyosmani/status/781179957384876033>)

Other than [Custom Elements (v1)](http://caniuse.com/#feat=custom-elementsv1) on caniuse:

- Grid: <http://caniuse.com/#feat=css-grid> with `y`
- WebGL 2.0: <http://caniuse.com/#feat=webgl2> with `u`

Fyrd:
Yup, was thinking the same thing.

Schweinepriester:
Okay, will you change those or should I open a pull request for both (or one for each)?

Fyrd:
Either way. If you don't do it I'll get around to it some time. :)

Schweinepriester:
I fucked up, will do a new pull request, including those other features.

PR #2855, opened 2016-09-29
Fixed link for customElements github polyfill

siggysamson:
The following link is currently broken: https://github.com/webcomponents/webcomponentsjs/tree/v1/src/CustomElements/v1

I found it when searching for Custom elements on [CanIUse](http://caniuse.com/#search=custom%20elements) (`Resources`: "customElements.define polyfill (supersedes document.registerElement)")

I replaced it with a new link to the current folder of the polyfill: https://github.com/webcomponents/webcomponentsjs/tree/v1/src/WebComponents

It might be a good idea to remove this link altogether though or just link to the Github repository itself instead of a folder in the tree (or a branch) as this is most likely going to break again in the future.

Let me know if I should remove the link from the list

### DIFF
--- a/features-json/custom-elements.json
+++ b/features-json/custom-elements.json
@@ -37,7 +37,7 @@
       "title":"Google Developers - Custom elements v1: reusable web components"
     },
     {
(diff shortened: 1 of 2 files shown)


Fyrd:
This is fine for now, will consider the other options later

PR #2857, opened 2016-10-02
Update flex support for UCBrowser

hexalys:
U3 0.8 / UCBrowser 10 (checked at 10.7) and now 11+ support unprefixed `flex` including Flexwrap. \o/


Fyrd:
@hexalys Just tried 10.10.8.820/0.80 and it doesn't seem to work unprefixed for me...

hexalys:
@Fyrd Ok, I misread it. You are correct. The 10.7.5 I have re-tested is prefixed. I corrected.

Fyrd:
Thanks, however it's not possible to add versions via PR. I'll see if I can update the browser to the latest version (11?) some time soon, then 9.9 will be replaced by it.

hexalys:
Ok, I wasn't sure how to handle that, sorry. I thought perhaps you'd add the versions first and then do a merge on it. Pretty new at this. Can't you make 9.9 a 9 perhaps and add UCWeb 10 and 11?
The newest UCWeb is at 11.0.5.x.

Fyrd:
@hexalys Have updated UC browser to the latest version (11.0.5.850) and support is still the same as before...

hexalys:
Argh, sadly you are right. I initially looked at feature detect mostly based on flexwrap, and 10 and 11 indicates those respective changes of support. I thought I did a test of one of my site without prefixes but I must have missed it. It's a false positive then, confirmed [here](http://stackoverflow.com/questions/38413091/detect-uc-browser-as-well-as-browsers-with-partial-support-of-flexbox). Which is even worse than no support. :(

Should we mention the false positive in known issues?

NekR:
Feature detection, sadly, doesn't work for UC. e.g. it says it supports unprefixed transitions, but fires prefixed transition end event. This kind of stuff.

hexalys:
@NekR It appears they are adding recent CSS webkit properties without the features... I really don't what to make of it. But I guess we have to assume UCWeb's so called U3 engine has strictly nothing new since Safari 5 / Webkit 534.40 then. UCWeb CSSRegion support is wrong also then and perhaps other things.

Fyrd:
Ugh, yeah that's too bad. Thanks for looking into it.

PR #2859, opened 2016-10-03
Add/update to list of browsers not supporting <use xlink:href>

hexalys:
See http://codepen.io/hexalys/pen/epErZj/ for insider details of tests and my polyfill methods.


Fyrd:
Thanks, though as mentioned at #2857 you cannot add new browser versions.

Fyrd:
Thanks!

PR #2862, opened 2016-10-04
Adjust "Experimental Features" in Safari TP

Schweinepriester:
As discussed in #2851.
With the addition of <http://caniuse.com/#feat=gamepad>, which I missed there.


Fyrd:
Looks great, thank you!

PR #2865, opened 2016-10-06
Updating Firefox's webgl status

potch:
Firefox has had full WebGL support without `experimental-webgl` since version 24.


potch:
I somehow missed the note about video drivers and partial support! Sorry.

PR #2871, opened 2016-10-07
upgradeinsecurerequests: add upgrade-insecure-requests alias

thejh:
Allow searching for `upgradeinsecurerequests` as `upgrade-insecure-requests` since that's the name that's actually used in the CSP header.


Fyrd:
Thanks!

PR #2873, opened 2016-10-08
Update @supports for Opera Mini at Presto 12.1 level

hexalys:
Now that Opera Mini it at Presto 12.1 level. It supports @supports like Opera 12.1 in the table.


Fyrd:
Good catch, thanks!

PR #2877, opened 2016-10-09
position:fixed partially supported in Android 4.0-4.3

mattiacci:
CSS transforms and margin:auto are ignored on position:fixed elements in Android 4.0-4.3.
Example: https://codepen.io/mattiacci/pen/mPRKZY

Note: Couldn't find a way to test Android 3, so not sure about compatibility there.

### DIFF
--- a/features-json/css-fixed.json
+++ b/features-json/css-fixed.json
@@ -230,9 +230,9 @@
       "2.2":"a #2",
       "2.3":"a #2",
       "3":"y",
-      "4":"y",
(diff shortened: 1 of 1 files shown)


Fyrd:
I would mark this as a bug rather than partial support...it's a bad side-effect but position:fixed itself is still supported.

mattiacci:
Updated to "full support with note" and updated note description.

Fyrd:
Cool, thanks!

PR #2878, opened 2016-10-10
textencoder: Add ie_id

cvrebert:
See https://developer.microsoft.com/en-us/microsoft-edge/platform/status/encodingstandard/


Fyrd:
Thanks!

PR #2883, opened 2016-10-10
data update for selection API

gritchou:
There a was a typo on the selectionchange event.
It's selectionchange and not selectionchanged.
[see mdn documentation for the selectionchange event here](https://developer.mozilla.org/en-US/docs/Web/Events/selectionchange)


Fyrd:
Thanks!

PR #2886, opened 2016-10-11
css-supports-api: Update Opera Mini to 12.1 Presto level

hexalys:
Like my recent similar @support update. The window.supportsCSS is also supported in Opera Mini. I did a couple tests to verify it.


Fyrd:
Thanks! I think I'll switch it to partial support though since they only supports the older API.

PR #2892, opened 2016-10-12
Fx49 supports Multiple column without prefix

mpopp75:
Prefixed CSS3 multi-column properties will be removed:

https://www.fxsitecompat.com/en-CA/docs/2016/prefixed-css3-multi-column-properties-will-be-removed/

It is implemented as of October 12 Nightlies.


Fyrd:
Thanks!

PR #2896, opened 2016-10-13
Chromium 54+ supports unprefixed user-select

citrusui:
See https://www.chromestatus.com/features/5722186220306432.


Fyrd:
Thanks! For future reference, note that you cannot add versions in pull requests.

PR #2897, opened 2016-10-13
Update status of CSS Grid to CR

tomalec:
> W3C Candidate Recommendation, 29 September 2016
https://www.w3.org/TR/css3-grid-layout/


Fyrd:
Thanks!

PR #2903, opened 2016-10-16
Clarify speech recognition support in FF

Daniel-Hug:
It's kind of disappointing when after enabling the flag, it still doesn't work. For reference see the FF footnote to the compatibility table on MDN: https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API


Fyrd:
Thanks!

PR #2906, opened 2016-10-17
update ambientlight information

patrickkettner:
it is working on devices with ambient light sensors


Fyrd:
Hmm, only seems to pass my test with the experimental web platform flag enabled. Can you update accordingly (or provide more info) ?

Fyrd:
@patrickkettner ^

PR #2910, opened 2016-10-18
Mark svg-html as partially supported on IE and Edge.

### LINKED ISSUE #2333: Edge still doesn't support SVG filter effects on HTML - should not be marked "Supported"
(svg-html)

Specifically, SVG `<filter>` elements referenced using `filter: url("#filter_id")` are not supported. Only the inline CSS phrases like `filter: blur(10)` seem to be supported.

Corresponding Edge bug: ~~https://connect.microsoft.com/IE/feedback/details/2379118~~ https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/6618695/

Arnavion:
Due to https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/6618695/

Fixes #2333

### DIFF
--- a/features-json/svg-html.json
+++ b/features-json/svg-html.json
@@ -31,14 +31,14 @@
       "6":"n",
       "7":"n",
       "8":"n",
-      "9":"a #1",
-      "10":"a #1",
-      "11":"a #1"
+      "9":"a #1 #2",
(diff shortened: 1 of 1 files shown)


Fyrd:
Confirmed, thanks!

PR #2912, opened 2016-10-19
Update xml-serializer.json

mattbrundage:
Note #3 (used by Firefox) was curiously missing, so I reconstructed it, based on https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML


Fyrd:
Thanks!

PR #2913, opened 2016-10-19
Update push-api.json: Opera dev 42 supports push

PascalReintjens:
The opera developer 42.0 version now seems to fully support push messages. While previous versions internally always returned a "push service not available" as soon as you tried to subscribe, dev 42 seems to be able to successfully subscribe and deliver messages. 

Tested with different sites, for example: https://gauntface.github.io/simple-push-demo/
While it now seems to work, it doesn't seem to be 100% stable and working on every implementation, but as it is a dev version, I am confident that this will change until the release of 42.

Browser support is getting better and better :)


Fyrd:
Thanks!

PR #2917, opened 2016-10-21
Added animation event bug for IE10/IE11

dogoku:
When animating a pseudo element, IE10 and IE11 will not fire the relevant Animation Events


Fyrd:
Thanks!

PR #2924, opened 2016-10-24
Chrome 55+ unprefixed clip-path

Schweinepriester:
<https://blog.chromium.org/2016/10/chrome-55-beta-input-handling.html>:

> CSS Clipping Path properties no longer require the webkit prefix.

Questions:

- What about Opera (41)?
- Should we update <http://caniuse.com/#feat=css-masks>?


Fyrd:
Thanks! Yes, I'll update Opera too, I don't think I've ever seen a CSS change in Blink that didn't make it to Opera.

What kind of change would you feel is needed for `css-masks`?

Schweinepriester:
I see, will keep that in mind for the future!

I thought maybe they removed the prefix for `css-masks` too (as `clip-path` is listed as a subfeature), but just checked and they didn't, so no change needed.

PR #2934, opened 2016-10-26
Safari TP 16

Schweinepriester:
<https://webkit.org/blog/7030/release-notes-for-safari-technology-preview-16/>:

> - Enabled the Gamepad API by default, allowing gamepads already configured for use with macOS to work on web pages using the Gamepad API
> - Supported CSS Shapes Level 1 without a -webkit- prefix
> - Ensured a toggle event is fired whenever details.open is changed


Fyrd:
Thanks!

PR #2939, opened 2016-10-28
Safari TP fixes: WOFF 2 & unicode-range

Schweinepriester:
- http://caniuse.com/#feat=font-unicode-range
- http://caniuse.com/#feat=woff2

Both are listed with Safari 10 and without Safari TP support.


Fyrd:
Thanks!

PR #2941, opened 2016-10-28
Add alternate stylesheet feature.

jornane:
I didn't find the **alternate stylesheet** feature on CanIUse so I decided to add it myself.

I filled out all the browsers I know about, plus some links to references. Browsers I don't know about I marked with `u`. Chrome does not support this feature but from what I've gathered the feature can be supported using a plugin. There doesn't seem to be a letter for that, so I just set it to `n` and mentioned the extension in the links.

I've set `shown` to `false`, because I don't know how many browsers we need to get a status on before it can be published on the site. However, may be a safe bet to assume only Firefox on desktop supports this out of the box.


Ugoku:
Old Opera (<14) used to support this, and according to [this page](https://www.w3.org/Style/Examples/007/alternatives.en.html) Internet Explorer supported this in 2001.

jornane:
Added a commit to show that Opera <14 and Internet Explorer >=8 support this as well. Edge, however, really does not support this.

Fyrd:
Thanks!

PR #2951, opened 2016-11-02
Added issue template

### LINKED ISSUE #2408: Can we use GitHub's reaction feature for issue voting?
[Recently](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments), GitHub implemented a "reactions" feature, allowing users to +1 issues without commenting.

I think it should be possible to have the issue "leaderboard" also read +1's left via this feature, in addition to the regular +1 comments.

This would help with, among other things, notification overload from issues that a given user has already commented on.

jimmywarting:
closes #2408

Given that we get all kinds of useless email spam with just +1 that adds no valuable information

I think it's much easier to just search and order ideas based on reaction then nr of comments or scrape github for +1 that are inconsistent with like +100 this way everyone only gets one vote which i think is fair 

This below is just what it will look like when you create a new issue:

----

<p align="center"><i>comments like +1 me too will be deleted</i><br><b>&#x25BC;Vote with reaction&#x25BC;</b></p>

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,8 @@
+[describe your issue here]
+
+
+
+
+----
+
+<p align="center"><i>comments like +1 me too will be deleted</i><br><b>&#x25BC;Vote with reaction&#x25BC;</b></p>


Fyrd:
I'll be happy to use this once GitHub finally has official support for their Reactions API and I've made the appropriate changes to the code that generates http://caniuse.com/issue-list/ . 

I believe an important thing you're missing that that list also includes all the votes that came through Google Moderator when issues were added there. There's no way to do a nice GitHub query that'll consider those numbers as well as the reactions. 

Also I can assure you that the script that reads the "+1" comments only considers one vote per user, regardless of how many times they comment or what number they actually put there.

PR #2963, opened 2016-11-05
Context menu support was removed from "Experimental Web Platform" flag

Christoph142:
https://bugs.chromium.org/p/chromium/issues/detail?id=412945#c48


Fyrd:
Thanks!

Issue #1871, opened 2015-08-31
Add ES7/ES2016 async functions


dstorey:
- One of the top trending features requests on Edge userVoice and one of the top JS features by votes too
- async/await is a very popular feature in other languages such as C#, and somewhat related to popular features such as Promises 
- An ES7 feature is already included here (object.observe)
- Implementation(s) have started

domenic:
As a general note, I'd prefer to remove ES versions, especially future ones. Version numbers are not very meaningful, and it's not guaranteed that something will land in a specific version.

tclzcja:
Just want to add that async function is already implemented in Microsoft Edge Windows Insider Preview Build 10532+. So this is a good time to add that entry in caniuse.

claar:
Searched for "await" today and was surprised not to find it in this awesome site.

NichUK:
+1

Ugoku:
+1

Landed in Chrome Canary recently

Schweinepriester:
+1

For the time being see <http://kangax.github.io/compat-table/es2016plus/#test-async_functions> as always, but I'd like to see this on caniuse, too.

mdsharpe:
+1

fundon:
https://developers.google.com/web/fundamentals/getting-started/primers/async-functions

vith:
Firefox:

> Support for `async/await` has been added in the Nightly build from the 31st of October (bug [#1185106](https://bugzilla.mozilla.org/show_bug.cgi?id=1185106)). Great stuff!

via https://blog.nightly.mozilla.org/2016/11/01/async-await-support-in-firefox/

michaellewiswilson:
+1

mikaturk:
+1

Fyrd:
Now available at http://caniuse.com/#feat=async-functions

PR #2965, opened 2016-11-05
#1871 - Add async functions

jarst:
Add feature file with async functions (issue #1871).


Fyrd:
Nice, thanks!

Fyrd:
@jarst Edge 14 has been set as "d", firstly it needs to include a "n" (I already did this), secondly there should be a note associated to it that describes how support can be enabled.

Issue #2925, opened 2016-10-24
Adjust CSS Hyphenation for Chrome 55


Schweinepriester:
<https://blog.chromium.org/2016/10/chrome-55-beta-input-handling.html>:

> Chrome now supports CSS automatic hyphenation, one of Chrome’s most frequently requested layout features, on Android and Mac. CSS hyphenation allows the browser to hyphenate words when line-wrapping, improving the visual consistency of text blocks. Hyphenation support will be extended to other platforms in future releases.

Revaluate/close #1993? Or close this one as duplicate?

PR #2967, opened 2016-11-06
CSS Hyphen Support for Chromium

PascalAOMS:
Further information: https://www.chromestatus.com/feature/5642121184804864


Schweinepriester:
This is not correct, sadly.

> Blink enables the following values on all platforms: "manual" - the initial value. "none" - disables soft hyphens. and the following value on Android and Mac: "auto" - enables the automatic hyphenation for languages where the underlying operating system has dictionaries.

Also see #2925.

Therefore it probably should be `y` for Android, but partial with comment for desktop.

PascalAOMS:
Oh shoot, how did I miss that?!
Thanks for the comment. I'll look into that later.

Issue #2940, opened 2016-10-28
Samsung Internet for Android 4 should support Webpush (Can someone verify it?)


PascalReintjens:
According to http://developer.samsung.com/technical-doc/view.do?v=T000000202 Samsung Internet 4 should support webpush.

Because I don't have a compatible device, I cannot test it, and therefore aren't creating a pull request yet, but if there is anyone who could verify that it works that would be great :)

PascalReintjens:
I visited a store today where I could successfully test this with an exhibition unit, and yes push notifications work fine under Samsung Internet 4. Pull request follows shortly :)

PR #2970, opened 2016-11-08
Update push-api.json - Samsung Internet 4 supports push

PascalReintjens:
As written in http://developer.samsung.com/technical-doc/view.do?v=T000000202 Samsung Internet 4 supports push. I tested that it works with an exhibition unit in a store. 

This fixes issue #2940


Fyrd:
Thanks!

PR #2971, opened 2016-11-08
Update push-api.json: Firefox for Android supports Push since v48

PascalReintjens:
As written in a footnote on https://developer.mozilla.org/de/docs/Web/API/Push_API: "Push has been enabled by default on Firefox for Android version 48." Firefox for Android supports push since version 48 by default.


Fyrd:
Thanks!

awebdeveloper:
also could u add a note about non-standard implementation in safari. Keep it red but add a note 

https://developer.apple.com/notifications/safari-push-notifications/

PR #2978, opened 2016-11-14
update android uc viewport-units support situation

SASUKE40:


Fyrd:
@SASUKE40 Only the latest version of UC browser (11.0) is being tracked on caniuse. It's not possible to change or add versions via pull request.

SASUKE40:
@Fyrd The latest version of UC browser is support of new features such as viewport units, you can update the latest situation so that developers will not be worried about using the those features.

Issue #2981, opened 2016-11-14
"cb() never called!" on NPM install


jonnyarnold:
I'm running into errors installing caniuse-db from NPM as of today:

```
npm install caniuse-db
npm ERR! cb() never called!
```

I've had this issue on my local Macbook Pro (MacOS Sierra) and on our CI Server (Codeship, using CentOS 6 on Docker).

Thanks for your help, and let me know if I can help you debug this issue.

catalin-miron-hs:
__node__: v5.9.1
__npm__: 3.7.3

```
catalin-miron ~/ubervu/thehole npm install -g caniuse-db@1.0.30000580
npm ERR! cb() never called!
```

__node__: v6.9.1
__npm__: 3.10.8

```
catalin-miron ~/ubervu/thehole npm install -g caniuse-db@1.0.30000580
/Users/catalin-miron/.nvm/versions/node/v6.9.1/lib
└── caniuse-db@1.0.3000058
```

jonnyarnold:
@catalin-miron-hs Thanks - good to know it's been fixed in Node 6! Do you know what the issue is here, and whether there's a Node 5.x version that doesn't run into this issue?

ovidiu-bute-hs:
@jonnyarnold the version `1.0.30000579` (one patch older) is compatible with Node 5.x

catalin-miron-hs:
@Fyrd what version of `npm` did you had when publishing this package? From what I can see, you published `1.0.3000058` with `npm > 3.x` and `1.0.3000057` with `npm < 3.x` is this correct?

Fyrd:
@catalin-miron-hs I'm not sure...I think I may have published it with another machine...didn't know that could have any side-effects. My understanding of npm publishing is pretty limited.

catalin-miron-hs:
@Fyrd that's not a problem, after talking with @ai it seems that _node v5.x_ it's not maintained/supported by nodejs https://github.com/nodejs/LTS. 
I could reproduce the bug only in `node v5.9.x` versions and seems legit to drop support for `v5`.

jonnyarnold:
@catalin-miron-hs If that is the case, can I suggest leaving a v5.x compatible version available and bumping the major version number? It is a breaking change, after all.

(Feel free to ignore me - nobody asked what I would do!)

PR #2982, opened 2016-11-14
Add multiple Node versions to Travis CI build

ovidiu-bute-hs:
This is related to issue #2981 , I think it's good practice to test multiple Node versions, not just `0.6`. The issue could have been avoided. I've added Node versions 4, 5, and 6 to the Travis build file.


Fyrd:
Thanks!

Issue #2734, opened 2016-08-17
MathML support shaky in Safari


Xyene:
caniuse data says MathML has been supported in Safari since 5.1, but support &mdash; even in 9.0 &mdash; seems to be closer to "partial support" than "full support" due to the large number of rendering issues making it hard to use for professional uses.

Some screenshots of the [MathML Torture Test](https://developer.mozilla.org/en-US/docs/Mozilla/MathML_Project/MathML_Torture_Test) from Safari 9.0 using Latin Modern Math fonts:

![image](https://cloud.githubusercontent.com/assets/1403503/17744792/3ec6e736-6477-11e6-9bc8-d91be9f49cbd.png)

![image](https://cloud.githubusercontent.com/assets/1403503/17744809/4c2159d4-6477-11e6-86a1-9260ca683e55.png)

![image](https://cloud.githubusercontent.com/assets/1403503/17744835/73fe4b60-6477-11e6-8ded-4cdf351305eb.png)

![image](https://cloud.githubusercontent.com/assets/1403503/17744938/df02f4c4-6477-11e6-9a48-fe353128a1ed.png)

The WebKit codebase recently received a bunch of improvements in its MathML support that resolved the above issues, among others (a very nice post on this can be found [here](http://frederic-wang.fr/mathml-improvements-in-webkit.html)), but as it stands I think at least the addition of a note mentioning rendering pitfalls would be appropriate.

atjn:
Seems like this should be closed

Xyene:
Yep :)

PR #2990, opened 2016-11-18
Clarify Safari's support of MathML

Xyene:
Refs #2734; Safari 10+ fixes the issues referenced in the latter ticket.

The [Mozilla MathML torture test](https://developer.mozilla.org/en-US/docs/Mozilla/MathML_Project/MathML_Torture_Test) doesn't have any issues in rendering common math symbols (e.g. summations, which were fairly buggy before). RTL support is pretty good (radicals render properly), and outstanding rendering issues with RTL math seem to be buggy rather than downright unsupported.


Fyrd:
I suspect you meant this to change to `a` for partial support? `p` refers to "supported with polyfill".

Additionally you should add a numbered note to explain the partial support. Thanks!

Xyene:
My bad, I made a mental note not to get tripped up by it, but still did.

I've added number notes to the Safari data, and also added a direct link to the MathML torture test it references.

Fyrd:
Looks great, thanks!

PR #3008, opened 2016-11-23
Update status of Grid in Firefox

0b10011:
Scheduled to ship in 52 (https://bugzilla.mozilla.org/show_bug.cgi?id=1217086#c5).


Fyrd:
Thanks!

PR #3010, opened 2016-11-24
Speech Synthesis: Updated info for Firefox

AurelioDeRosa:
As noted [here](https://developer.mozilla.org/en-US/Firefox/Releases/49), the Speech Synthesis is supported in Firefox by default since version 49.


Fyrd:
Ah, not sure how I missed that. Thanks!

PR #3020, opened 2016-11-29
Firefox for Android does not support internationalization yet.

bartolom:
Intl object. Because Firefox is one of the last major browsers without support, it
might be worth it to add the tracking bug as well.
https://bugzilla.mozilla.org/show_bug.cgi?id=1215247


Fyrd:
Thanks, added manually.

PR #3021, opened 2016-11-29
Fixed Safari flexbox bug #136041

mkurz:
https://bugs.webkit.org/show_bug.cgi?id=136041 has been marked as fixed minutes ago.

(I also changed the working for the ff bug so it's more clear on first sight)


Fyrd:
Since Safari 10 is still the current stable version, I don't think it makes sense to refer to it in the past tense. Can you adjust the wording? Thanks.

mkurz:
Updated.

Fyrd:
Awesome, thanks!

PR #3022, opened 2016-11-29
Update preload state for ucbrowser

zhaoz:
From testing ucbrowser, I'm not seeing this working. If it is, its with caveats, where did this change come from? It seemed like it was a blanket change for and_uc from 9.9 to 11 which may have accidentally flipped this from n to y.

https://github.com/Fyrd/caniuse/commit/37add34280280c99d350ce5dbcface404cecc0fb


Fyrd:
Hmm, you're right. Just tested again and can confirm it's not working.

PR #3026, opened 2016-12-01
Update pointer events and touch-action support in Firefox

lpd-au:
- touch-action is riding the trains for FF52 (https://bugzilla.mozilla.org/show_bug.cgi?id=1244402#c8)
- Pointer events remain nightly only by config and aren't planned to ship in any near future version (checked bugzilla, platform status, experimental features)
- Both `dom.w3c_touch_events.enabled` and `layout.css.touch_action.enabled` also exist in about:config on FF Android.


Fyrd:
Thanks!

PR #3028, opened 2016-12-03
PaymentRequest is supported since Chrome 53

dandv:
Per https://www.chromestatus.com/feature/5639348045217792


Fyrd:
The version number is not the first version with support, it's just the latest version, as only the latest version of most mobile browsers are displayed on caniuse.

PR #3033, opened 2016-12-06
"Only works on secure (https) servers" started with Chrome 50

bsvensson:
ref: https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only


Fyrd:
Thanks!

PR #3041, opened 2016-12-09
Remove confusing bug text from nav-timing.json

heavi5ide:
According to https://www.w3.org/TR/navigation-timing/#sec-navigation-timing the NavigationTiming values are all longs -- MDN says these are "number of milliseconds since the Unix epoch." So unlike the High Resolution Time API (`window.performance.now()`), these are _supposed_ to be integer numbers of milliseconds. This bug description implies that Chrome on Windows is not following the spec, but millisecond resolution would seem to be correct.

### DIFF
--- a/features-json/nav-timing.json
+++ b/features-json/nav-timing.json
@@ -18,9 +18,7 @@
     }
   ],
   "bugs":[
-    {
-      "description":"Chrome on Windows - window.performance timing is millisecond resolution."
-    }
+    
   ],
   "categories":[
     "DOM",


Fyrd:
Looks like this was changed in the Navigation Timing 2 spec: https://www.w3.org/TR/navigation-timing-2/

but also the related bug referred to in #175 was fixed:
https://bugs.chromium.org/p/chromium/issues/detail?id=158234

so yeah, can remove this entry. Thanks!

PR #3043, opened 2016-12-10
Update script-defer.json

Yaffle:
Information about a bug in Firefox, it is useful to know about it If you expect DOMContentLoaded event in your script.


Fyrd:
Thanks!

PR #3044, opened 2016-12-10
Create webkitcssmatrix.json

Yaffle:
https://github.com/Fyrd/caniuse/issues/1001


features-json/webkitcssmatrix.json
```diff
@@ -0,0 +1,282 @@
...
+    "DOM"
+  ],
+  "stats":{
+    "ie":{
+      "5.5":"n",
+      "6":"n",
+      "7":"n",
+      "8":"n",
+      "9":"n",
+      "10":"x #1",
```
Fyrd: These need an included `y` or `a` value

Fyrd:
Thanks! Few notes:
- The `x` values need a `y` or `a` included to specify support or partial support.
- Can you rename the file to "dommatrix.json" since DOMMatrix is the standard?

Yaffle:
@Fyrd, the situation is similar to "-webkit-text-stroke", when the prefix "webkit" is a de facto standard....
but OK, I will change the filename.

Fyrd:
Thanks! I'll merge this as it is right now, though I think we'll want additional notes to clarify which prefixed variation is supported. But I'll be happy to help out with that too if you like.

PR #3052, opened 2016-12-14
Added grab and grabbing cursors

Gerboa:
grab and grabbing are supported by the same browsers as zoom-in and zoom-out


Fyrd:
No, the support is actually different. There's another feature for that at `css3-cursors-grab` though it requires more research and doesn't appear on the site yet.

PR #3048, opened 2016-12-12
Safari TP enabled HTML interactive form validation

mkurz:
Tweet: https://twitter.com/chris_dumez/status/806601644855074818

Mentioned in: https://webkit.org/blog/7093/release-notes-for-safari-technology-preview-19/

Ticket: https://bugs.webkit.org/show_bug.cgi?id=165123

Commit: https://trac.webkit.org/changeset/209060/trunk/Source

PR #3058, opened 2016-12-19
Pattern attribute in Safari now works as well

mkurz:
I should have included this in #3048 
As a consequence of Safari TP enabling HTML interactive form validation this bug got fixed now as well.


Fyrd:
Thanks!

PR #3063, opened 2016-12-21
Add bug regarding the time control in Edge

ymor:
Source: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8810417/


Fyrd:
Thanks!

PR #3064, opened 2016-12-21
transforms2d: Add "skew" to search keywords

### LINKED ISSUE #3055: Please add skew as a keyword to CSS transforms
Due to <https://www.w3.org/TR/css-transforms-1/#identity-transform-function>, `skew` should be added.

cvrebert:
Fixes #3055


Fyrd:
Thanks!

PR #3076, opened 2016-12-28
Update css-sticky.json

matll42:
Support Edge 14


Fyrd:
Not according to my test. Do you have a test case where this works?

PR #3082, opened 2016-12-31
Update CSS Grid with Safari TP support

jensimmons:
Safari Technical Preview now ships with CSS Grid on by default. So the box should be green. I'm not sure whether you want to leave the note about how to toggle it on and off (since those folks who installed Safari TP in the past will need to toggle it on) or not. I've included two pull requests — commit both to keep the note. Or just the first to remove the note.


Fyrd:
Thanks!

PR #3086, opened 2017-01-04
Dialog element implemented in Fx 53 behind flag

mpopp75:
A basic implementation of the Dialog element is now available in Firefox Nightly:

https://bugzilla.mozilla.org/show_bug.cgi?id=1322938


Fyrd:
Thanks!

PR #3089, opened 2017-01-04
Firefox 53 has support for clip-path basic shapes

jensimmons:
This updates the page.


jensimmons:
Here's the Bugzilla ticket about Clip-path supporting basic shapes in Firefox, with a note about it already being committed to FF53: https://bugzilla.mozilla.org/show_bug.cgi?id=1247229#c11

Fyrd:
Thanks!

PR #3092, opened 2017-01-05
Opera 40+ supports FIDO U2F

tld:
Opera 40+ supports FIDO U2F:
https://www.yubico.com/support/knowledge-base/categories/articles/browsers-support-u2f/

Also tested this in Opera 42 on OS X, and it works just as well as in Chrome.
Maybe a small holdout for Opera 41, which had some issues:
http://forums.opera.com/discussion/1876096/u2f-stopped-working-in-opera-41/p1

At the moment, FIDO U2F-adoption in the browsers has been somewhat sluggish, so showing the support that's actually there would be a good thing.  Operas support is the first proper and official support outside of Chrome, making it somewhat of a milestone.


Fyrd:
Thanks!

PR #3102, opened 2017-01-10
Safari supports async functions

mkurz:
According to http://kangax.github.io/compat-table/es2016plus/ and https://github.com/kangax/compat-table/pull/962 Safari TP already has support.


Fyrd:
Thanks!

PR #3105, opened 2017-01-11
rrggbbaa and rgba - Chrome 53 on Android

jhpratt:
No support


Fyrd:
Did some of my own testing on this which already covers this one, thanks.

PR #3111, opened 2017-01-14
input type="color" correction

jhpratt:
No support in Safari, unknown in TP


Schweinepriester:
Testcase: <http://www.w3schools.com/html/tryit.asp?filename=tryhtml_input_color>

Seems to work in TP:

bildschirmfoto 2017-01-14 um 09 21 01
bildschirmfoto 2017-01-14 um 09 21 08
bildschirmfoto 2017-01-14 um 09 21 15

Not in 10 though.

Also couldn't find an entry in the changelog or the corresponding bug (short search).

Fyrd:
Yeah, not sure how this wound up being wrong for Safari 10, doesn't work for me either. Thanks! Will update TP manually.

PR #3112, opened 2017-01-14
Add new feature: onbeforeprint and onafterprint

lpd-au:
26th on the feature suggestion list.

Testcase: https://jsfiddle.net/obtx118a/6/

IE support: https://www.w3.org/wiki/List_of_events
Firefox support: https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeprint#Browser_compatibility
Chrome support: https://bugs.chromium.org/p/chromium/issues/detail?id=218205
Safari support: https://bugs.webkit.org/show_bug.cgi?id=19937
Opera 12/FF Android support: only tested manually


Fyrd:
Nice work, thanks!

PR #3113, opened 2017-01-14
IE/Edge: No CSS transforms on SVG elements

Sphinxxxx:
No version of IE/Edge supports CSS transforms on SVG elements. Moved notice from "Known issues" to "Notes" in the same way it's done on http://caniuse.com/#feat=svg-html5

Resources:
https://developer.microsoft.com/en-us/microsoft-edge/platform/status/supportcsstransformsonsvg/
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8713749/


Fyrd:
Thanks!

PR #3120, opened 2017-01-19
rel-noopener support starting from Firefox 52

Aljullu:
You can see the bug here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1222516


Fyrd:
Thanks!

PR #3127, opened 2017-01-23
Update passive-event-listener.json

maiis:
Support in Safari 10

screen shot 2017-01-23 at 14 23 29
screen shot 2017-01-23 at 14 23 34

http://webreflection.github.io/dom4/test/


Fyrd:
You're right, thanks!

PR #3129, opened 2017-01-23
Update text-decoration.json

mattbrundage:
Ships by default in Chrome 57/Opera 44. https://www.chromestatus.com/feature/5766823203045376


Fyrd:
Confirmed, thanks!

PR #3135, opened 2017-01-25
FF 52: Selection API, Touch API; 53: CSS3 tab-size

Schweinepriester:
Three missing tidbits I found while studying the changelogs:
- <https://developer.mozilla.org/en-US/Firefox/Releases/52>
- <https://developer.mozilla.org/en-US/Firefox/Releases/53>


Fyrd:
Very nice, thanks!

Issue #3143, opened 2017-01-30
String.prototype.padStart


C-Bouthoorn:
As per [this MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart), it already has support on multiple browsers. It'd be nice if you could include this function too.

jhpratt:
Done along with `padEnd()` in #3145

dotnetCarpenter:
@jhpratt Seems that it isn't pushed. https://caniuse.com/#search=padStart and https://caniuse.com/#search=padEnd **0 results found.

tomByrer:
Still not there @Fyrd 
In all browsers but IE11, but helpful to know how far back compat goes.

PamProg:
Still not there @Fyrd

Fyrd:
Now available at https://caniuse.com/#feat=pad-start-end

PR #3145, opened 2017-01-30
padStart()

jhpratt:
#3143


Fyrd:
Support in Firefox started at v48, not 51, just like for padEnd. Also since the properties are related and support appears to be the same for both, can you combine the two into one? Thanks.

jhpratt:
I was just going off the MSN page, which has it listed as 51. I can combine them if you'd like, but it really doesn't make a difference. I was just trying to keep it one file per PR.

Fyrd:
MDN has great info, but the browser support often has mistakes as it tends to go unverified. Please do your own testing when submitting features. 

Also multiple files does make a difference in various ways, I prefer not making the repo size any larger than it has to be when it can be avoided.

jhpratt:
Just merged the files into one.

What do you use for browser testing? I don't have anything.

Fyrd:
Thanks!

There's some suggested services mentioned under https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#adding-a-feature, or you can find older versions to download and run. Alternatively see if browsers themselves have the info, like chromestatus.com.

PR #3146, opened 2017-01-31
css-sticky.json: add examples & gotchas article

geddski:
adds a resource that shows examples of the kinds of things you can build with `position:sticky`, as well as gotchas you need to watch out for (like sticky sibling elements, `position: absolute` children, etc)


Fyrd:
Thanks!

geddski:
@Fyrd thank you for such a valuable resource! The entire community owes you big.

PR #3152, opened 2017-02-02
FF53 : update CSS masks support

iamvdo:
Most of the spec is now supported by Firefox 53, so I updated data.
Also, added a new link (personal one but I think this could help many of us), and remove a 404 one.


Fyrd:
Thanks!

PR #3160, opened 2017-02-04
Add keywords (rrggbbaa, rgba)

jhpratt:


Fyrd:
FYI Technically this isn't necessary as these words already appear in the title/description which are also used for searches.

jhpratt:
Ah, didn't know that. I just tried `rgba` and it didn't show up.

Fyrd:
Ah good point, that one was necessary then. :)

PR #3166, opened 2017-02-08
XHR2 - FormData corrupted IE10/11

usrbowe:
`FormData` are not working properly on IE10/11, when there is unchecked
checkbox at the end of form.

### DIFF
--- a/features-json/xhr2.json
+++ b/features-json/xhr2.json
@@ -32,6 +32,9 @@
     },
     {
       "description":"In Internet Explorer the `timeout` property may only be set after calling `open` and before calling `send`."
+    },
+    {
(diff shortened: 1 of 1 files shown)


mkurz:
Do you know if this is fixed in Edge?

usrbowe:
@mkurz not sure, I didn't tested, but I think that it will not be a problem in Edge. Here is article about that issue: https://blog.yorkxin.org/2014/02/06/ajax-with-formdata-is-broken-on-ie10-ie11

mkurz:
The article is from 2014, Edge was released in 2015. It would be nice if we can confirm this is fixed in Edge already.

Fyrd:
Thanks! If anyone's able to verify on Edge that would indeed be useful.

PR #3168, opened 2017-02-09
Change Edge support to partial

mooyah:
MS Edge doesn't support most forms of webm: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8802502/


Fyrd:
Partial support should be `a` not `p`.

PR #3171, opened 2017-02-13
create media-session-api features json file

entonbiba:
create media-session-api features json file

https://developer.microsoft.com/en-us/microsoft-edge/platform/status/mediasession/?q=media%20session
image


Fyrd:
Your JSON file has some syntax errors in it. Also, you can't add version numbers not already present in the sample file. Can you fix these? Thanks.

entonbiba:
@Fyrd fixed, how can we add version 57 for chrome?

Fyrd:
Thanks! Currently upcoming versions of Chrome for Android don't appear on the site, so there's no way to add it.

PR #3179, opened 2017-02-16
Correct Safari iOS support for passive listeners

RByers:
Tested with [this](http://jsbin.com/pujayoh) and verified performance impact with [this test](https://rbyers.github.io/scroll-latency.html)


Fyrd:
Thanks!

PR #3180, opened 2017-02-17
wasm is available behind a flag in Edge 15

chicoxyzzy:
https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/11423808-add-support-for-web-assembly-as-early-as-possible


Fyrd:
Thanks!

PR #3184, opened 2017-02-22
Fix Opera version for CSS Grid

zcorpan:
Opera's version number corresponds to that of Chrome minus 13. So current stable version is 43, which does not support grid.

---

I did not check if this problem is present elsewhere in caniuse, or why it says the current version is 42 when it should be 43.


Fyrd:
Hmm, I'd swear one of the recently released versions was different by 14 versions points instead of the previous 13. Ah well, my mistake. Has now been corrected and have updated CSS grid together with a few others. The current version has been updated to 43. Thanks!

zcorpan:
Thank you!

PR #3185, opened 2017-02-22
Add mediacapture-fromelement

miguelao:
A.k.a. capture from canvas, capture from media element, etc.

Stable in FF 43+, Cr 51+ (all platforms, only `<canvas>` capture), Opera 36+.


miguelao:
@uysalere, @martinthomson, FYI/comment if you want.

Fyrd:
Thanks!

PR #3186, opened 2017-02-22
Fix history API compatibility on Edge and IE

abramenal:
Hi,

I've spotted that the provided info is not valid,
Seems that IE and Edge still doesn't support scroll restoration API.

Proof info could be found here:
https://developer.mozilla.org/en-US/docs/Web/API/History

We may also close this issue: #1889

### DIFF
--- a/features-json/history.json
+++ b/features-json/history.json
@@ -47,14 +47,14 @@
       "7":"n",
       "8":"n",
       "9":"n",
-      "10":"y",
+      "10":"a #1",
       "11":"y"
     },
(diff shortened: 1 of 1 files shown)


Fyrd:
MDN browser support tables are usually correct, but I wouldn't call that "proof". Also based on that information many other browser versions don't support the new scrollRestoration API either so it would seem inappropriate to single out IE & Edge here. 

Since it's a new addition to the spec I think it may be better to make this its own new feature as is indicated by #1889.

PR #3189, opened 2017-02-23
Firefox supports logical props since v41

cvle:
Hi, 

I'm working on https://github.com/cssinjs/css-vendor, a vendor prefixer based on feature detection.
I noticed that firefox seems to fully support logical props and looking at the MDN it seems it does so since v41.

See e.g.:
https://developer.mozilla.org/en-US/docs/Web/CSS/border-block-start
https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-start

This is my first contribution to this project.
Can anyone double-check?


cvle:
See changelog as well here: https://developer.mozilla.org/en-US/Firefox/Releases/41

Fyrd:
Checked and confirmed. Thanks!

cvle:
How long does it normally take to have this in data.json?

PR #3191, opened 2017-02-23
flexbox.json: add width vs flex-basis resource

geddski:
answers the most frequently asked flexbox question


geddski:
you bet! Thank you!

PR #3200, opened 2017-02-27
Fixes #3164 - Downgrade Safari/Chrome for Custom Elements to Partial

### LINKED ISSUE #3164: WebKit support for CustomElements
I'd suggest this be downgraded to partial support. The spec defines two types of custom elements:

* Autonomous custom element
* Customized built-in element

https://html.spec.whatwg.org/multipage/scripting.html#custom-elements

WebKit supports the former only, and they are opposed to implementing the latter:

https://github.com/w3c/webcomponents/issues/509#issuecomment-222860736
https://github.com/w3c/webcomponents/issues/509#issuecomment-224212371
https://github.com/w3c/webcomponents/issues/509#issuecomment-230700060
https://github.com/w3c/webcomponents/issues/509#issuecomment-233222085

ctolkien:
As discussed in #3164, Safari only has partial support and is actively opposed to offering full support for the specification.

As per the contribution guidelines, I feel  that:

> "Can I use the feature by default?"

The answer is not "yes", but "partially".


Fyrd:
As mentioned in #3164 it seems support for Customized built-in elements doesn't exist in any browser. So either all support should be "partial" for now or we can split the feature into two (either is fine with me)

ctolkien:
@Fyrd  - I've changed Chrome to partial in a subsequent commit. My 2c, the spec defines it as 1 feature, so makes sense for caniuse to reflect it as 1 feature as well.

Fyrd:
Thanks!

PR #3202, opened 2017-02-28
UC Browser 11 doesn't need prefix for transition

cvle:
Working on https://github.com/cssinjs/css-vendor and with thorough testing on an Android device with the current UC Browser v11 and the first released version of UC Browser v11 revealed that it fully supports the `transition` property without prefixes.


cvle:
Ok I just saw this on PR https://github.com/Fyrd/caniuse/pull/2857.

> Feature detection, sadly, doesn't work for UC. e.g. it says it supports unprefixed transitions, but fires prefixed transition end event. This kind of stuff.

I need to take a closer look...

cvle:
Closing for now until I figured out something new.

PR #3203, opened 2017-02-28
Add note for browsers not supporting column-fill

cvle:
PR does 2 things:
- Moves general note to a numbered note.
- Mark browsers not supporting column-fill

column-fill is not support in:
- Chromium browsers < 50
- Opera < 37 (same as Chromium < 50)
- Opera Mobile < 37 (same as Chromium < 50)
- Safari & IOS Safari < 6.2 (verified using browserstack)
- All BlackBerry version (released before webkit patch was merged in master)
- Samsung 4 (based on a Chromium version < 50)
- UC Browser 11 (tested manually)

References:
- https://www.chromestatus.com/feature/6298909664083968
- https://bugs.webkit.org/show_bug.cgi?id=117693


Fyrd:
Nice job, thanks!

PR #3204, opened 2017-02-28
android needs prefix for border-image

cvle:
Similar to Chrome < 15 and Safari < 6, Android < 4.4 needs a prefix for `border-image`.
Tested on a VM using feature detection.


Fyrd:
Thanks!

PR #3208, opened 2017-02-28
Webkit #137730 was marked as fixed

mkurz:
See https://bugs.webkit.org/show_activity.cgi?id=137730

Related:
https://github.com/philipwalton/flexbugs/issues/197
https://github.com/philipwalton/flexbugs/issues/84

### DIFF
--- a/features-json/flexbox.json
+++ b/features-json/flexbox.json
@@ -53,7 +53,7 @@
       "description":"In IE10 the default value for `flex` is `0 0 auto` rather than `0 1 auto` as defined in the latest spec."
     },
     {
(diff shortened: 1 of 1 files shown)


gsnedders:
We don't actually know this will be fixed in 10.1; it's not implausible that it's too late for 10.1 at this point, but who knows what Apple will do given they never comment on future product release.

Fyrd:
Yeah, better for now to point out that it's fixed in the latest WebKit Nightly.

mkurz:
I think I will just wait until Safari 10.1 is released, check and then update this pull request if needed. We can keep it open until then.

Fyrd:
Yeah okay.

daytonlowell:
Safari 10.1 shipping in iOS 10.3 and this bug appears to still be present.

Fyrd:
Seems to finally be fixed in Safari 11. Have updated the data manually, so closing this PR.

PR #3213, opened 2017-03-01
Former firefox does not support align-content

cvle:
Old firefox versions do not support the align-content property.

See our CI log: https://travis-ci.org/cssinjs/css-vendor/builds/205736604#L462


Fyrd:
Cool, thanks!

PR #3215, opened 2017-03-03
Safari on iOS does support image/*

bofrede:
This must rely on a misunderstanding.
iOS presents a menu where you can choose to take a new photo, select a photo taken earlier or pick one from another app.
I have tested it myself on iOS 8.1.1, 9.3.5 and 10.2.
I am not sure if it works in iOS 6 and 7, since I don't have access to devices with those versions of iOS.


Fyrd:
Thanks!

PR #3218, opened 2017-03-04
Update Chrome support for Referrer-Policy

justin-morgan:
Chrome does not support `same-origin`, `strict-origin`, or `strict-origin-when-cross-origin` values ([issue 627968](https://bugs.chromium.org/p/chromium/issues/detail?id=627968))


Fyrd:
Can you change this to a numbered note instead of a bug? Else it won't be obvious to people that the bug refers to the partial support. Thanks!

justin-morgan:
All set, it's now a new numbered note.

Fyrd:
Excellent, thanks!

PR #3219, opened 2017-03-05
Android Chrome does support notifications

sheldon:
They have supported this since 2015, not sure where the idea came from that they didn't. http://www.theverge.com/2015/4/20/8454613/push-notifications-android-chrome-42-mobile-web


Fyrd:
That's a different API. The API that Android Chrome supports is http://caniuse.com/#feat=push-api

sheldon:
Thanks for the clarification, I'm amazed and saddened that the simpler api isn't supported in addition.

PR #3221, opened 2017-03-06
IE & Edge don't fully implement old snappoints spec

cvle:
Document IE & Edge versions not supporting the `scroll-snap-coordinate` and `scroll-snap-destination` from the old spec.


Fyrd:
Thanks!

PR #3222, opened 2017-03-06
IE & Edge do not support region-fragment

cvle:
Document the fact that IE & Edge do not support `region-fragment`.


Fyrd:
Thanks!

PR #3224, opened 2017-03-07
Adding bug/note for Chrome 55+ not fully working

jwmickey:
Chrome stops playback after about 15 seconds since version 55 on some platforms, including at least: Windows 7 & 10, Ubuntu 14.04.


Fyrd:
Thanks!

PR #3228, opened 2017-03-08
Opera 44 (Chromium 57) does not support `display: flow-root`.

Marat-Tanalin:
Opera 44 (currently Beta) is based on Chromium 57 and does NOT support
`display: flow-root`.
Opera 45 (currently Developer) (based on Chromium 58) does.


Fyrd:
Thanks!

PR #3229, opened 2017-03-08
Introduce auxclick feature

NavidZ:
@RByers @smaug----


RByers:
Cool, thanks for adding this!  I'm not deeply familiar with the format here, but still LGTM.  Maybe include in the CL description a link to the bugs (and/or browser status entries) that show when this shipped in Chromium and Gecko?

NavidZ:
Sure.
Here are a few links:
[MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Events/auxclick)
F[ireFox auxclick bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1304044) 
[Chrome auxclick feature](https://www.chromestatus.com/feature/5663174342737920)

There are some fields inside the json to add these I believe. For example I added the chrome one inside. But I wasn't sure about the others. So I wanted to see whether the maintainer has a particular guideline on how I should add them in the json file.

Fyrd:
The first two should be added under links, the Chrome one is unnecessary since indeed you added it as the chrome_id. Thanks!

NavidZ:
Done. Links added.

Fyrd:
Thanks!

features-json/aux-click.json
```diff
@@ -0,0 +1,289 @@
...
+  "spec":"https://wicg.github.io/auxclick/",
+  "status":"wd",
+  "links":[
+    {
+      "url":"https://developer.mozilla.org/en-US/docs/Web/Events/auxclick",
+      "title":"MDN documentation"
+    },
+    {
+      "url":"https://bugzilla.mozilla.org/show_bug.cgi?id=1304044",
+      "title":"FireFox implementation"
```
smaug----: s/FireFox/Firefox/

lpd-au:
Reviewing this data:

* The spec should be `unoff`, not `wd`, until https://github.com/w3c/uievents/issues/107 is closed and published as a working draft.
* Why are so many versions marked as unknown? Afaik there's never been an older/removed feature sharing this name so there's no reason for eg old IE to be (**U**)nknown instead of (**N**)o support.
* Why is Chrome 55+ (**P**)olyfill only instead of (**Y**)es supported?
* Is there a reason for Opera and Android Chrome not matching Desktop Chrome?
* Firefox 53+ should have a note/bug explaining that click handlers on `document` and `window` only are still run on secondary button clicks (for now) as a compatibility measure.
* Not sure if this is important but the keywords has a trailing comma.

Otherwise seems accurate.

NavidZ:
Do we need to do anything else here regarding the verification of the data or is it all done? I'm asking because the feature is not yet appeared on [caniuse.com](http://caniuse.com/#search=auxclick).

lpd-au:
@NavidZ 
> Do we need to do anything else here regarding the verification of the data or is it all done? I'm asking because the feature is not yet appeared on caniuse.com.

You can try submitting a test for it (http://tests.caniuse.com/submit.html) but honestly the fastest way would be to just get a bunch of people to leave +1 comments at https://github.com/Fyrd/caniuse/issues/2923 so it moves up the [feature suggestion list](http://caniuse.com/issue-list).

PR #3233, opened 2017-03-10
Firefox 54+ supports ES6 Modules behind a flag

dpogue:
Ref: https://blog.hospodarets.com/native-ecmascript-modules-the-first-overview#enabling-es-modules-in-firefox


Fyrd:
Thanks!

PR #3238, opened 2017-03-10
Credential Management API issue in Chrome 51 ~ 56

tjwudi:
In Chrome 51 ~ 56, you cannot use credential set at a.d.com in b.d.com. This was fixed in Chrome 57.

Using credentials across different subdomains were not enforced by the spec. The spec used the term MAY but not MUST. But Chrome chose to support this in Chrome 57, so I'd say it was a bug in versions prior to 57. Let's annotate Chrome 51 ~ 56 as partially supported then.


agektmr:
I'm hesitant to call this a bug as the spec change happened preceding our change IIRC.
What do you think @vabr-g ?

tjwudi:
I agree that this might not be a bug by definition, but definitely something worth mentioning so people can be aware of it. But it looks like other than "partially support" there is no better way to annotate Cr51~56's discrepancy with Cr57. I might be wrong though :)

agektmr:
I don't think marking this as a bug is good idea, but I agree that we lack documentation around this. Actually I'm planning one under Web Fundamentals as part of FAQ for CM API. I also think noting this in MDN makes sense.

@jpmedley can we work together to address this in MDN? We have a few notable updates around cross origin CM API calls.

vabr-g:
I agree with @agektmr that this was not a bug. The spec uses "MAY" intentionally to make this an optional feature. Chrome decided (https://crbug.com/666340) to change how it implements the spec, but it did not implement less of it before.
I am not sure how important it is to note this change, and I have no good ideas about how to note this. Given that 57 is now the stable Chrome, and that Chrome automatically updates, a negligible amount of users will be affected by the pre-57 behaviour soon.

agektmr:
Luckily I had a chance to chat with @jpmedley in person today about this and he's looking into documenting this in MDN. I will prepare one for Web Fundamentals as well.

tjwudi:
Sounds good! Thanks for following up.

I think I can change "partially support" back to support, and add this as a note_by_num (general note) in caniuse.com. Does this sound good to you guys?

vabr-g:
Sounds good to me. Thanks!

tjwudi:
@Fyrd Can you help reviewing this PR? =)

Fyrd:
Looks good, thanks!

PR #3241, opened 2017-03-15
APNG supported in new Chromium versions

therealmarv:
see
https://chromium.googlesource.com/chromium/src/+/7d2b8c45afc9c0230410011293cc2e1dbb8943a7


Fyrd:
Thanks!

PR #3247, opened 2017-03-16
Add note for Safari bug to multicolumn.json

Vestride:
Safari 5.1-10+ has a bug where setting `min-height` on an element with css columns will clip the contents to that min-height.

[test case](https://codepen.io/herrschuessler/pen/LNVWJE)

[bug report](https://bugs.webkit.org/show_bug.cgi?id=65691)


Fyrd:
Thanks!

PR #3249, opened 2017-03-17
Update hsts support by ios as of 8.4

kylerandolph:
See https://support.apple.com/en-us/HT205030


Fyrd:
You can't actually change the version number ranges. I suggest in this case you change "8.1-8.4" to partial support with a note explaining the support is in 8.4 specifically.

Fyrd:
Also this feature doesn't refer to HSTS, as mentioned in #3250.

Issue #3262, opened 2017-03-23
srcset: Edge 14 intermittently distorts images, scaling to a fixed height of 30px


OliverO2:
Edge 14 intermittently displays distorted images as soon as it encounters a srcset attribute. Only workaround seems to be filtering out srcset attributes before delivering HTML content to Edge.

Bug report: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7778808/
* First reported: 2016-06-03
* Confirmed: 2016-08-18
* Probably to be fixed in Edge 15 (does not seem to reproduce in pre-release builds, Microsoft has not confirmed the causality)

http://caniuse.com/#feat=srcset

OliverO2:
Reflected in commit fcfce29651880a7b3344450cc39ce44ac1db0114

PR #3263, opened 2017-03-23
Add srcset bug in Edge 14 (issue #3262)

OliverO2:
See #3262.

### DIFF
--- a/features-json/srcset.json
+++ b/features-json/srcset.json
@@ -14,7 +14,9 @@
     }
   ],
   "bugs":[
-    
+    {
+      "description":"Edge 14 intermittently displays distorted images as soon as it encounters a srcset attribute, seemingly dependent on caching and network timing. [see bug](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7778808/)"
+    }
   ],
   "categories":[
     "HTML5"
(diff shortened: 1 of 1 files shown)


Fyrd:
Support should be "partial" here at worst, which uses `a` instead of `n`. Also according to the bug report it occurs in Edge 13 as well.

OliverO2:
@Fyrd Thanks for the review. I've included Edge 13.

I would agree with "partial" support if the feature was at least useful under _any_ known circumstance or with some known workaround. Since Edge 13 and 14 cannot reliably display images _at all_ when `srcset` is present, what would be the meaning of "partial" in this case?

Fyrd:
I don't really have much experience with this issue, but I'll accept your changes for now since you're right that if it's not reliable at all it's effectively unusable. May revisit this in the future if others disagree.

OliverO2:
Agreed. We had a site which seemed to work reliably with Edge 14 and srcset and for some time. When we just changed the timing of asynchronous script and font loading, rendering of images broke on Edge 14 about 25% of the time, though the srcset stuff remained identical.

If someone came up with information to reproducibly correct the situation, it would be most welcome.

PR #3272, opened 2017-03-25
WebAssembly: Opera supports wasm since v44

chicoxyzzy:
WebAssembly is enabled by default in corresponding version of v8 (same as in Chrome)


Fyrd:
I believe this should be v44, just tried 43 and there's no support.

chicoxyzzy:
@Fyrd you are right. Updated my PR

Fyrd:
Thanks!

PR #3304, opened 2017-03-27
Custom Elements support added in iOS 10.3

daytonlowell:


Fyrd:
@daytonlowell FYI it's not possible to add new versions via PR. I'll include and test iOS 10.3 soon though.

daytonlowell:
Yeah, I figured you'd add 10.3 and then the tests would pass and you'd approve ¯\_(ツ)_/¯

PR #3305, opened 2017-03-28
Safari Custom Push notification support

awebdeveloper:
https://developer.apple.com/notifications/safari-push-notifications/
https://appadvice.com/appnn/2013/09/apple-to-enable-ios-inspired-safari-push-notifications-for-developers-in-os-x-mavericks


Fyrd:
Thanks!

PR #3315, opened 2017-04-04
iframe-seamless: various link fixes

waldyrious:
- restore link to spec that actually contained a reference to this attribute (changed in 719ee06c3)
- add link to the removal from the W3C version of the HTML5 spec
- include status for Firefox bug
- add Edge UserVoice official response (updated 2017-04-04)


Fyrd:
Thanks!

PR #3318, opened 2017-04-04
touch-action note for Firefox 52+

lpd-au:
Added a note about not all Firefox platforms supporting pointer events or touch events, rendering touch-action support moot (eg Windows without APZ, Mac OS).


Fyrd:
Thanks!

PR #3324, opened 2017-04-06
Update intrinsic-width.json

mattbrundage:
Updates to Safari TP per https://trac.webkit.org/changeset/213831/webkit/


Fyrd:
Already covered by previous PR, thanks.

yisibl:
As far as I know, this is only the removal of the prefix in CSS Grid.

Issue #1306, opened 2015-05-12
navigator.hardwareConcurrency    Chrome: Native support All other browsers that support Workers: few...


Fyrd:
navigator.hardwareConcurrency

http://wiki.whatwg.org/wiki/Navigator_HW_Concurrency

Chrome: Native support
All other browsers that support Workers: few seconds async timing attack w/polyfill
Native support in Safari coming soon

---
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)

Moderator votes: **+5**
Issue added by **Eli Grey** on **2014-05-15**
To vote this issue up or down, simply include `+1` or `-1` in your comment.

solendil:
This is a very useful feature for javascript intensive applications like [FractalJS](https://github.com/solendil/FractalJS).
+1

kenrussell:
+1

cyrusboadway:
+1

mudcube:
+1

eligrey:
+1

nmrugg:
+1

SergioCrisostomo:
+1

zackee12:
+1

lpd-au:
+1

Fyrd:
Now available at http://caniuse.com/#feat=hardwareconcurrency

PR #3332, opened 2017-04-07
Add incomplete data for navigator.hardwareConcurrency

lpd-au:
See issue #1306

I verified the Chrome, Firefox, IE and Opera Desktop data, couldn't figure out whether the support in Webkit was actually enabled in Safari and if so, when.


Fyrd:
Thanks!

PR #3333, opened 2017-04-07
service workers: y for chrome 45+, opera 32+, firefox 44+

dfabulich:
In #2023 Firefox 44 was marked as "almost supported" but https://jakearchibald.github.io/isserviceworkerready/ lists it as completely green.

Chrome is marked "a" since Chrome 40, but there's an [open PR on isserviceworkerready](https://github.com/jakearchibald/isserviceworkerready/pull/88) that Chrome 45+ and Opera 32+ should be marked with full support. 

Jake Archibald himself says that [he thinks it's good enough to be green now](https://twitter.com/jaffathecake/status/849064881424932868).


Fyrd:
Sounds good to me. Thanks!

PR #3335, opened 2017-04-10
Update intersectionobserver.json for Edge15 capability

ctolkien:
Updates Edge15 to full support (from partial). This was added in 14986 (Edge15 is 15063)

https://developer.microsoft.com/en-us/microsoft-edge/platform/status/intersectionobserver/


Fyrd:
Thanks!

PR #3344, opened 2017-04-11
Chromium bug #700383 is fixed

sailingdeveloper:
https://bugs.chromium.org/p/chromium/issues/detail?id=700383


Fyrd:
@sailingdeveloper I think you just wanted to remove the note there, not the entire entry. :)

PR #3347, opened 2017-04-12
Clarify the history of webgl2 in Firefox

jdashg:


features-json/webgl2.json
```diff
@@ -260,7 +260,8 @@
       "57":"n"
     },
     "and_ff":{
-      "52":"n"
```
Fyrd: @jdashg Sorry for the delay on this one! Can you replace the "and_ff" entry with just `"52":"y"` as it's not possible to add entries and only the latest version is tracked for this browser. Thanks!

Fyrd:
Updated manually.

PR #3350, opened 2017-04-12
Add safari support for navigator.hardwareConcurrency

lpd-au:
I tried in a couple of versions of desktop safari (9.1 and 10.0.3) and ios safari (10.2.1) and surprisingly couldn't find a pulse. That's all 5 major browsers covered now and this is a top 40 request.


Fyrd:
Thanks!

PR #3354, opened 2017-04-13
Update Chrome and Opera support of :focus-within

dwilsonmsft:
Per https://www.chromestatus.com/feature/5363834508279808, :focus-within is behind a flag in upcoming Chrome/Opera.


Fyrd:
Thanks!

PR #3355, opened 2017-04-13
Update Firefox support for text-justify

chenpighead:
Firefox supports text-justify property with values: auto | none | inter-word | inter-character | distribute (with the exact same meaning and behavior as "inter-character" for legacy reasons) from Firefox 54 (under layout.css.text-justify.enabled pref off) [1]. text-justify will be pref-on from  Firefox 55 [2]. 

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=276079
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1343512

### DIFF
--- a/features-json/css-text-justify.json
+++ b/features-json/css-text-justify.json
@@ -92,10 +92,10 @@
       "49":"n",
       "50":"n",
       "51":"n",
-      "52":"u",
-      "53":"u",
-      "54":"u",
-      "55":"u"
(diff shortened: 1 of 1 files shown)


chenpighead:
The MDN page has been created recently. See https://developer.mozilla.org/en-US/docs/Web/CSS/text-justify

Fyrd:
Thanks!

PR #3358, opened 2017-04-15
FLIF

jhpratt:
No support by any browser, current or planned. Polyfill is available.


Fyrd:
Sorry, I don't think having a polyfill warrants adding this as its own support table considering there's no browsers planning on implementing it. However I can add a search result as I did for BPG: http://caniuse.com/#search=bpg

Fyrd:
Okay, done: http://caniuse.com/#search=flif

jhpratt:
Yeah, that works. If there was a way to add search results, I would've done that instead.

PR #3359, opened 2017-04-15
Update Chrome support for SDCH

redburn:
"...we are on track for disabling SDCH in M59."
https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/nQl0ORHy7sw


Fyrd:
Thanks!

PR #3375, opened 2017-04-20
Update serviceworkers.json

KenjiBaheux:
Fix: support on Chrome is the same between Desktop and Android.
Fix: Android Browser was incorrectly marked as "partial support".

Issue: 3374


KenjiBaheux:
Hmm, I'm not quite sure how to fix this "[serviceworkers:android:56] Browser version missing: android 56" error. 

I believe it means that Android Browser 56 is expected to be present, which is odd given that there is no such a thing.

Fyrd:
@KenjiBaheux It's not possible to include extra or remove versions in the JSON files. Only the latest version of Chrome for Android is currently included in the data.

While there isn't an "Android Browser 56" exactly, this refers to the latest Chrome WebView which is used in some native Android browsers.

PR #3390, opened 2017-04-23
css-grid: mention Firefox bugs fixed in 54

Forkest:
I didn't add the note to all previous versions, but the bugs are probably there too.


Fyrd:
Thanks!

PR #3392, opened 2017-04-23
Remove unnecessary "guess"

OriginalEXE:
Support for hosting WebSockets in WebWorkers did indeed arrive in Firefox 38, so unneeded wording can be removed.


Fyrd:
Thanks!

PR #3402, opened 2017-04-26
Update opus.json

mattbrundage:
Fix typo; proper casing

### DIFF
--- a/features-json/aac.json
+++ b/features-json/aac.json
@@ -277,7 +277,7 @@
       "7.12":"y"
     }
   },
-  "notes":"Support refers to using the `audio` element, not other conditions.",
+  "notes":"Support refers to this format's use in the `audio` element, not other conditions.",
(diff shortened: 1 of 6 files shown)


Fyrd:
The intent in those notes is to say "Support refers to use **[of this format]** using the `audio` element, not other conditions" so removing the word "use" doesn't make sense. I'm open for other suggestions to rephrase that though.

mattbrundage:
The implied prepositional phrase "of this format" was not apparent to me. And the word "use" followed directly by "using" is awkward, especially when "use" can be construed as a verb.

A suggestion: "Support refers to this format's use in the ``audio`` element, not other conditions.

Fyrd:
@mattbrundage Good suggestion, let's go with that. Thanks!

Fyrd:
Thanks!

PR #3404, opened 2017-04-26
Update mpeg4.json

mattbrundage:
Attempting to escape the parenthesis characters in the MSDN URI, as the link was getting clipped and causing a 404


Fyrd:
Think you may need to add some extra slashes there to make it valid JSON...

mattbrundage:
Ok, try this revised commit.

Fyrd:
Thanks!

PR #3406, opened 2017-04-27
Add keywords for better findability of CSS3 selectors

laurens:
e.g. looking up "not" would not yield a result for the CSS3 `:not()` selector.


Fyrd:
This won't make a difference as the description is searched before keywords. Also "not" does yield a result for the `:not()` selector (among the CSS3 selectors), it's just at the very end of the results.

Note that a search for `:not` results in it appearing as the second result though.

Issue #2155, opened 2015-12-17
Hyphenation Options


animaux:
Support for these CSS4 Text-properties would be interesting:

+ hyphenate-limit-lines
+ hyphenate-limit-chars

And maybe further:

+ hyphenate-character
+ hyphenate-limit-last

Schweinepriester:
+1

<https://drafts.csswg.org/css-text-4/#hyphenation> currently lists:

- `hyphenate-character`
- `hyphenate-limit-zone` (new?)
- `hyphenate-limit-chars`
- `hyphenate-limit-lines`
- `hyphenate-limit-last`

KurtPreston:
I think this issue can be closed now with #3407?

stephane-MMDEV:
+1
Information on this support, with prefixes or not, is still difficult.

sznowicki:
+1 I agree on this. It looks like `-webkit-hyphenate-character` works since at least few years in Chrome but it's super hard to know exactly where it doesn't.

Now when I check it also works well on Safari.

robinrendle:
Just bumped into this again and thought the property didn’t exist because it wasn’t listed on caniuse.

Schweinepriester:
Arriving here again via https://groups.google.com/a/chromium.org/g/blink-dev/c/mVlDAHZge5g/m/J8-qrckhCAAJ (Chrome is unprefixing `-webkit-hyphenate-character` in 106):
At least `hyphenate-character` is available via MDN now: https://caniuse.com/mdn-css_properties_hyphenate-character

Kathara:
Still missing an entry on caniuse.com about hyphenate-limit-chars. Will this be added?

tobiasfabian:
### Resources

- [Chrome Platform Status](https://chromestatus.com/feature/5150761588097024)
- [Chromium Bug for Automatic hyphenations for Win/Linux/ChromeOS](https://bugs.chromium.org/p/chromium/issues/detail?id=924069)
- [Article CSS-Tricks](https://css-tricks.com/almanac/properties/h/hyphenate-limit-chars/)

Schweinepriester:
In the wake of Firefox preparing to ship `hyphenate-limit-chars`

* https://mastodon.social/@intenttoship.dev@bsky.brid.gy/113866707486680563
* https://groups.google.com/a/mozilla.org/g/dev-platform/c/rs3wccTx79s/m/cV723AYHDQAJ

[...]

PR #3407, opened 2017-04-28
Adding CSS4 hyphenate entry

KurtPreston:
For issue #2155.

Couldn't find much information on current support besides the table written up here:
http://webdesign.about.com/od/fonts/a/hyphenation-in-css3.htm

I can confirm that `-webkit-hyphenate-character` works in Chrome 55 and up.


Fyrd:
"CSS4" is not a valid category. I recommend just using "CSS".

KurtPreston:
@Fyrd Just pushed a change adding CSS4 as a valid category.  Seems like it should be, since CSS3 is used as a category and CSS4 has a working draft.  If you'd prefer though, I can just remove the tag instead.

Fyrd:
Ah, nice try, but [CSS4 is not a thing](http://www.xanthir.com/b4Ko0). Also categories are managed elsewhere, new categories can't just be added by PR.

If anything I'd like to remove the CSS2 & 3 categories at some point.

KurtPreston:
K, I've pushed an update that just changes the category "CSS".

But for the record, CSS4 is totally a thing.  The article you linked to was published in 2013.  W3C's CSS4 spec was posted in 2015:
https://www.w3.org/TR/css-text-4/

Fyrd:
Thanks!

Level 4 != CSS4 though. Here's a newer article if you don't believe me. :)
https://rachelandrew.co.uk/archives/2016/09/13/why-there-is-no-css4-explaining-css-levels/

Fyrd:
Oh, and more specifically:
https://www.w3.org/TR/CSS/#css-level-4

KurtPreston:
Hehe, should have known better than to dispute the maintainer of CanIUse.  Thanks for the clarification, and thanks for your excellent project!

PR #3408, opened 2017-04-28
Update auxclick spec location

lpd-au:
auxclick is now on a standards track, see https://github.com/w3c/uievents/issues/107


Fyrd:
Thanks!

PR #3416, opened 2017-05-05
Require support for table parts to be "supported"

0b10011:
Since one of the big features of `position:sticky` is to keep header rows in place, I think it's important to mark this as "almost supported" for browsers that don't support this feature.

Firefox just [finished the foundation for this change](https://bugzilla.mozilla.org/show_bug.cgi?id=929484), so it should hopefully be getting the fix in soon.

Chrome has shown that they want to [work on this](https://bugs.chromium.org/p/chromium/issues/detail?id=702927), but haven't started, as far as I can tell.

I'm not sure if Safari support is correct; I just believed the bug description that's currently on there.

---

If you'd rather not change the status to "almost supported" for these browsers, I think it's still important to separate out these issues per browser to make it easier to figure out which versions work and which ones don't.


Fyrd:
Sorry for the delay on this one, wanted to do a bit more research on it. Have updated the data manually based largely on your changes. I've also written my own test at https://codepen.io/Fyrd/pen/wegxqN to easily test support for upcoming browsers. Thanks!

Issue #3323, opened 2017-04-06
Safari 7.1/8 and Edge 13/14 only has partial support of HTML templates


manubb:
More precisely, in those browsers (and maybe others), cloneNode calls on nested templates produce incorrect results. Support can be checked by using the tests used in [webcomponents/template](https://github.com/webcomponents/template) polyfill.

manubb:
Support is also partial in iOS<9.

englishextra:
Edge 13 doesnt support template as MDN and caniuse say. It crashes as soon as you manipulate with it, whereas IE 11 survives, but it needs ccs display none for template tag. It needs polyfill loadded before body tag. The other thing is that it' about reusing markup in tepmplate tag instead of creating that markup and then appending new nodes.

manubb:
A more precise support table:

|                |Chrome       |Firefox        |IE/Edge    |Opera         | Safari    |
|----------------|-------------|---------------|-----------|--------------|-----------|
|Flawless native |&ge;35       |&ge;22         |-          |&ge;22        |&ge;9      |
|Buggy native    |26-34 (i)    |-              |13-15 (c,i)|15-21  (i)    |6.2-8 (c,i)|

On line "Buggy native", c denotes problems when cloning templates or nodes that contain templates; i denotes problems when importing templates or nodes that contain templates. 

Note that Safari 6.1 and 7 do not have native `<template>` whereas Safari 6.2 and 7.1 do have (buggy) implementations.

See [manubb/template](https://github.com/manubb/template) for polyfill details and test suite. (Do not load the polyfill in test/basic.html to test native support. Results are displayed in the console.)

PR #3420, opened 2017-05-07
Fixed <template> support table

manubb:
This is a fixed feature support table for `<template>` element as described in #3323.


Fyrd:
Thanks!

PR #3421, opened 2017-05-07
added new polyfill

mfranzke:


Fyrd:
Thanks! I'm curious though, why "polypill"? Is this a typo for "polyfill" or does it actually mean something different?

mfranzke:
@Fyrd, thanks for that. "polypill" might be an interesting new phrase - but actually it's just (german) autocorrection ... :-)

PR #3424, opened 2017-05-09
Extend srcset bug to Edge 15

OliverO2:
Unfortunately, the srcset bug (discussed in PR #3263) is also present in Edge 15. Confirmed by Microsoft, amending the original [bug report](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7778808/). (I have also personally verified the bug's existence.)

### DIFF
--- a/features-json/srcset.json
+++ b/features-json/srcset.json
@@ -15,7 +15,7 @@
   ],
   "bugs":[
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Unfortunate indeed, thanks.

PR #3427, opened 2017-05-11
ES6 modules behind a flag in chrome 60

mkurz:
https://www.chromestatus.com/feature/5365692190687232

Also in Opera 47, but there is no data yet in th json for this version.


Fyrd:
Thanks!

PR #3428, opened 2017-05-11
Remove Edge bug which has since been fixed.

ymor:
Fixed and released as per https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8810417/


Fyrd:
Thanks!

PR #3435, opened 2017-05-15
Update mutationobserver.json for UC 11.2

baribadamshin:


Fyrd:
It's not possible to add new browser versions via PR. Only the latest version of this browser is tracked.

baribadamshin:
But It latest version of this browser on Android 5.1, many people still use 5.1 most often on phones from chinese like Meizu or Xiaomi

PR #3443, opened 2017-05-17
Chrome 58 removed <link rel=prerender>

sergiolopes:
Chrome removed support for prerender on M58 and intends to (in the future) ship another solution to this feature.

https://groups.google.com/a/chromium.org/d/msg/blink-dev/0nSxuuv9bBw/PKu8Nv3yCAAJ


Fyrd:
Thanks!

PR #3448, opened 2017-05-18
Update WebVR information

HeinPauwelyn:
1. Added bug: 

    > WebVR is in development and is still changing. Because of this there could be bugs inside the standard of the W3C.

2. Chrome for Android supports WebVR but added note 5:

    > Supports only Google Daydream on [daydream-ready devices](https://vr.google.com/daydream/smartphonevr/phones/) and Google Cardboards

3. Samsung Internet supports WebVR but added note 4:

    > Supports only Samsung Galaxy devices with the Samgung Gear VR

4. Changed note 1:

    > Available only in Firefox nightly & Firefox Developer Edition, not yet on track to be enabled for stable & beta builds and supports only the Oculus Rift and the HTC vive on Windows VR-ready computers.

5. Changed note 2:
    
    > Enabled behind the WebVR & \"Gamepad Extensions\" flags under `chrome://flags`. Currently builds use an older version of the (still changing) specification and supports only the Oculus Rift and the HTC vive on Windows VR-ready computers.

6. Changed note 3:

    > [In development](https://blogs.windows.com/msedgedev/2016/09/09/webvr-in-development-edge/#3lMW05DTZXbXcK46.97) in the latest Edge builds and supports only [Windows Mixed Reality](https://developer.microsoft.com/en-us/windows/mixed-reality).


Fyrd:
If the support is behind a flag, you'll need to use `n` instead of `a` as `a` suggests partial support exists by default (not behind a flag)

HeinPauwelyn:
I've used `a` because browser like Chromium and Firefox Nightly supporting WebVR. Of course they aren't the same browsers but WebVR is almost in Chrome and Firefox.

Fyrd:
Thanks for the changes. I understand your reasoning, but I suspect leaving it as partial would result in people expecting to see that in an upcoming stable build by default.

PR #3465, opened 2017-05-26
Update notes on backface-visibility for Safari 10 (fixes #3464)

### LINKED ISSUE #3464: Backface-visibility still requires vendor prefix in Safari 10
The notes for [CSS 3D Transforms](http://caniuse.com/#feat=transforms3d) state that:
> Safari 9 is reported to still require a prefix for the related backface-visibility property.

I can confirm this issue is still present in Safari 10.1 as well, so the note should probably be updated.

CharlieHawker:


Fyrd:
Thanks!

PR #3471, opened 2017-05-30
IE and Edge don't follow the KeyboardEvent.key specification

### LINKED ISSUE #3468: KeyboardEvent.key not properly supported by IE and Edge
The current information is misleading as IE and Edge support for this is so different from the other browsers it is effectively useless.

The issues are:

 a) It has different names for symbols, probably based on an older version of the spec
 b) AltGraph + key generates incorrect values

Right now the information is very misleading and seems like it is safe to go ahead and use KeyboardEvent.key. At the very least there should be a note.

CendioOssman:
Fixed #3468.

I don't know how to test these modifications (except running `validate-jsons.js`), but hopefully this is the correct syntax. :)

### DIFF
--- a/features-json/keyboardevent-key.json
+++ b/features-json/keyboardevent-key.json
@@ -33,15 +33,15 @@
       "6":"n",
       "7":"n",
       "8":"n",
-      "9":"y",
-      "10":"y",
-      "11":"y"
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #3476, opened 2017-05-31
Background-clip: text support

nuxodin:
Check here for current browser support:
https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip


Fyrd:
Thanks!

PR #3478, opened 2017-06-01
Add an issue with macOS Safari

ruchern:


Fyrd:
Safari 5 is really old with only about 0.03% usage so I don't feel this is noteworthy.

ruchern:
@Fyrd Whoops. I added a wrong version. The actual version is 10. I am on macOS Sierra 10.12.5.

PR #3481, opened 2017-06-01
Update netinfo.json

Melatonin64:
Looks like this is available in current Chrome Canary (61).


Fyrd:
I'm not seeing support there... and https://bugs.chromium.org/p/chromium/issues/detail?id=368358 confirms the support is only in Chrome for Android right now.

Melatonin64:
I guess this is still in flux...
I see a `navigator.connection` object on windows Chrome Canary build, but not on OSX.

They're adding a few new properties (`downlink`, `rtt`, `effectiveType`), here's the bug:
https://crbug.com/719108
It says *OS: All*.

Melatonin64:
@Fyrd Chrome 61 stable is out.
Network Information is available also on desktops.
See https://www.chromestatus.com/features/6338383617982464

PR #3497, opened 2017-06-07
Removed ObjectRTC from Safari 11

daytonlowell:
I can't find any evidence that Safari 11 supports ObjectRTC, which as caniuse states: "is often dubbed WebRTC 1.1. It is possible to make ORTC communicate with WebRTC 1.0 endpoints.". I can't find anything in Apple's release notes that points to WebRTC 1.1 or ObjectRTC.

Resources: 
https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Safari_11_0/Safari_11_0.html
https://www.w3.org/community/ortc/
https://bugs.webkit.org/buglist.cgi?quicksearch=webrtc%201.1
https://bugs.webkit.org/buglist.cgi?quicksearch=object%20RTC


Fyrd:
Seems fair. Thanks!

PR #3498, opened 2017-06-08
it seems that IE 6/7/8 also support documentFragment!

xgqfrms-GitHub:
It seems that *IE 6/7/8* also support `documentFragment`!

ie5-test-ok


Fyrd:
Note that that's the DOM Range entry you're looking at there which mentions DocumentFragment in its description, but doesn't refer to support to it.

`documentFragment` itself is indeed already supported everywhere and doesn't have its own entry.

PR #3500, opened 2017-06-09
Add Resize Observer

ausi:
See #3006. It has currently 12 upvotes and two duplicates.

ResizeObserver is supported since Chrome 54 behind a flag: https://www.chromestatus.com/feature/5705346022637568


Fyrd:
Thanks!

PR #3513, opened 2017-06-14
Correcting Edge status for Payment Request

kypflug:
Payment Request is supported and on by default in Edge 15+. The "preliminary implementation" note here appears to come from an Edge changelog note which was specifically referring to an Edge 15 preview build. The stable release has a complete implementation. See https://aka.ms/devguide_edgehtml_15 for details.


Fyrd:
Thanks!

PR #3514, opened 2017-06-14
Geolocation API for FF55+ only in secure contexts

Schweinepriester:
<https://developer.mozilla.org/en-US/Firefox/Releases/55>:

> - The Geolocation API is now available only to secure contexts


Fyrd:
Thanks!

PR #3515, opened 2017-06-14
Update flag name and support status in Chrome

dandv:
image

The flag might have been called that way in M58 or older, but I don't have a copy handy.


Fyrd:
Has been updated manually due to conflict with previous PR. Thanks!

PR #3532, opened 2017-06-21
Srcset shipped on Edge

arjunyel:
Srcset is technically shipped on Edge but has a bug as mentioned in note 3


Fyrd:
See the discussion at https://github.com/Fyrd/caniuse/pull/3263: The reason support is not set to "partial" is because the bug is serious enough to make it inadvisable to use at all.

arjunyel:
Sorry about that, next time I'll look at the history of the file before making a pull request

PR #3533, opened 2017-06-21
Ambient Light Sensor API is available in Chrome behind a flag

NOtherDev:
See [Intent to Implement](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/TkfdVqYAYiE/xLGN2b1-AAAJ) and [Tracking Bug](https://bugs.chromium.org/p/chromium/issues/detail?id=606766).


Fyrd:
Thanks!

PR #3535, opened 2017-06-22
Add PaymentRequest is behind flag in FF

marcoscaceres:
We have a partial implementation that people can enable. More parts are currently landing, but there is enough there to play.


features-json/payment-request.json
```diff
@@ -108,9 +108,9 @@
       "52":"n",
       "53":"n",
       "54":"n",
-      "55":"n",
-      "56":"n",
-      "57":"n"
+      "55":"n #d 6",
+      "56":"n #d 6",
+      "57":"n #d 6"
```
Fyrd: These should be `n d #6`

marcoscaceres: D'oh! fixing

Fyrd:
Thanks!

PR #3542, opened 2017-06-23
Add know issue on edge support for blob url

sarod:
Reflect missing support for blob url in object.data and iframe.src tag

### DIFF
--- a/features-json/bloburls.json
+++ b/features-json/bloburls.json
@@ -31,11 +31,11 @@
       "11":"y"
     },
     "edge":{
-      "12":"y",
-      "13":"y",
-      "14":"y",
-      "15":"y",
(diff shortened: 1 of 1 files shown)


features-json/bloburls.json
```diff
@@ -291,7 +291,7 @@
   },
   "notes":"",
   "notes_by_num":{
-    
+    "1": "Created blob url can't be used in object or iframe (see https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/5411066/)",
```
Fyrd: @sarod Seems to be a trailing comma here, can you remove that? Thanks!

Fyrd:
Added manually.

PR #3549, opened 2017-06-27
Update referrer-policy.json

mattbrundage:
Chrome 60 gets full support per https://bugs.chromium.org/p/chromium/issues/detail?id=627968


Fyrd:
Thanks!

PR #3551, opened 2017-06-27
Noting Safari lack of support w/ Grid Layout

jpdevries:
Just discovered this. Pretty bummed. 

Webkit bug here 
https://bugs.webkit.org/show_bug.cgi?id=173873


Fyrd:
This should be added under "bugs" rather than the numbered notes, since it refers to a specific usage of the property.

Fyrd:
@jpdevries Can you make the change to this PR & #3552 ? Thanks.

Fyrd:
Added manually.

PR #3552, opened 2017-06-27
Noting lack of intrinsic sizing support in Safari

jpdevries:
See Also
https://github.com/Fyrd/caniuse/pull/3551


jpdevries:
Webkit bug here 
https://bugs.webkit.org/show_bug.cgi?id=173873

Fyrd:
Like #3551, this should be listed under bugs.

Fyrd:
Added manually.

PR #3555, opened 2017-06-28
Update history.json

zhouqicf:
Android 2.2+ support History API, test result in Android 4.1.1:


Fyrd:
See the Notes section. There appears to be support, but actually using it shows it doesn't work reliably and is effectively unusable.

PR #3557, opened 2017-06-29
Updating IE11 support availability

himanshuapril1:
It's now supportable in IE11 as well. Please review and update.


Fyrd:
Not according to my tests or Microsoft. Only in Edge.
https://developer.microsoft.com/en-us/microsoft-edge/platform/status/adownloadattribute/?q=download

PR #3561, opened 2017-06-30
Added STP and Safari 11 added support for HEVC

daytonlowell:


Schweinepriester:
While hurray, do you have a source for this?

I hoped and searched for this since Apple mentioned HEVC all over the place, but I don't think its in the changelog[1] for example, is it? 

[1] https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Safari_11_0/Safari_11_0.html

daytonlowell:
I never found any documentation but I can play the HEVC files from here https://x265.com/hevc-video-files/ in Safari on iOS 11.

daytonlowell:
I added a note to the desktop Safari 11 and TP entries that makes it clear it's only supporteed under macOS High Sierra.

Fyrd:
Thanks!

PR #3574, opened 2017-07-05
Adds resource to css-grid entry

the-simian:
Includes Rachel Andrew's Grid by example site that has patterns, tutorials and more resources to understand this spec


Fyrd:
Thanks!

PR #3583, opened 2017-07-08
CSS Grid is now enabled by default in Edge

Schweinepriester:
Per <https://developer.microsoft.com/en-us/microsoft-edge/platform/changelog/desktop/16237/>

> Unprefixed CSS Grid is now enabled by default

🎉


Fyrd:
Thanks!

PR #3589, opened 2017-07-13
Updated IntersectionObserver Firefox information

therealshark:
According to https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver
the flag is dropped with version 55.


Fyrd:
Thanks!

therealshark:
Thank you for providing this service. The least i can do is contributing if i find an error :)

PR #3593, opened 2017-07-13
Tweak some spec IDs for HTML

domenic:
These IDs also appear in the developer's edition of HTML (https://html.spec.whatwg.org/dev), and so including these allows caniuse integration with the dev edition (whatwg/html#2794).


Fyrd:
Thanks!

Issue #2300, opened 2016-02-13
Add <script type="module"> and module workers


domenic:
We'd love to add one of those nice caniuse boxes in the margin of the HTML spec for `<script type="module">` and module workers. Nobody implements yet, but everyone is actively implementing!

probins:
`<script type="module">` is available under http://caniuse.com/#feat=es6-module

domenic:
Thanks! Unfortunately that section is not accurate. ES6 modules (specified in ECMA262) are a different feature than `<script type="module">` (specified in HTML).

probins:
well, yes, but the two are interdependent. In practical terms, the answer to 'caniuse modules?' is 'yes, once module scripts are implemented'. @Fyrd, perhaps change the intro text to "Load ES modules as module scripts" or something similar, and add a link to the whatwg html spec as well? Also, Firefox needs adding to the list; its implementation recently landed in nightly behind a flag, and is scheduled for FF54 https://developer.mozilla.org/en-US/Firefox/Releases/54

Fyrd:
This was added with #3594 ( where https://caniuse.com/#feat=es6-module was updated ) so closing this issue.

PR #3594, opened 2017-07-14
Expand and add JavaScript modules entries

domenic:
* Turns the existing "ES6 module" entry into "JavaScript modules: <script type="module">", and cleans up its data a bit
* Adds new module-related entries for dynamic import() and the nomodule attribute

Test for nomodule: https://cdn.rawgit.com/jakearchibald/6110fb6df717ebca44c2e40814cc12af/raw/7fc79ed89199c2512a4579c9a3ba19f72c219bd8/

Test for dynamic import(): https://plnkr.co/edit/pgsxdd6hE7uzQY9dHCwz?p=preview

---

Partially solves https://github.com/Fyrd/caniuse/issues/2300, although I refrained from adding module workers as no browser seems to have started work on them yet.

### DIFF
--- a/features-json/es6-module-dynamic-import.json
+++ b/features-json/es6-module-dynamic-import.json
@@ -1,36 +1,16 @@
 {
-  "title":"JavaScript modules: <script type=\"module\">",
-  "description":"Loading JavaScript modules using `<script type=\"module\">`",
-  "spec":"https://html.spec.whatwg.org/multipage/scripting.html#attr-script-type",
-  "status":"ls",
+  "title":"JavaScript modules: dynamic import()",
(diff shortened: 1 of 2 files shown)


Fyrd:
Thanks! I'm not 100% certain these need to be separate though when support across browsers is largely the same for each. Differences can always be pointed out with notes. But I'll accept the PR and review further at another point.

domenic:
Support is very different though, as you can see from the data :)

Flimm:
+1 for a new entry for [dynamic import](https://developers.google.com/web/updates/2017/11/dynamic-import?utm_source=feed&utm_medium=feed&utm_campaign=updates_feed).

PR #3597, opened 2017-07-15
Add CSS prefix to flexbox

Psithief:
## Problem:
Currently, searching for `CSS` will not show this feature. 

## Proposal:
Prefix with CSS like (most?) other CSS features.

## Addendum:
According to the CR, the full name of this feature is `CSS Flexible Box Layout Module Level 1`.


Fyrd:
Fine by me. Thanks!

PR #3604, opened 2017-07-18
Use HTTPS links for ecma-international.org

xiaoyinl:
ecma-international.org now supports [HTTPS](https://www.ecma-international.org/ecma-262/6.0/), so let's use it.


Fyrd:
Thanks!

PR #3616, opened 2017-07-24
Corrected `background-attachment:fixed` Known Issue

ScottSmith95:
The issue referenced notes that it is `background-attachment: fixed` not `background-positon: fixed` that causes the issue. (`background-position: fixed` is an invalid property.)


Fyrd:
Thanks!

PR #3623, opened 2017-07-28
css-variables: Added Edge 15 bug

HHogg:
Related to a current and verified bug with Edge 15 that IMHO makes css-variables unusable in large projects. 

Microsoft bug link https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12143022/


Fyrd:
Thanks!

PR #3625, opened 2017-07-29
Update referrer-policy.json

mattbrundage:
Chrome: implementation bumped to v. 61 per this comment: https://bugs.chromium.org/p/chromium/issues/detail?id=627968#c15 (and confirmed with tests).


Fyrd:
Thanks!

PR #3630, opened 2017-07-31
Edge bug #7778808 is fixed

mkurz:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7778808/ was marked as fixed a couple of days ago.


Fyrd:
Thanks!

PR #3633, opened 2017-07-31
Patch 24

yukulele:
* remove outdated links
* add MDN
* Add Firefox bug link


Fyrd:
Hmm, shame those links stopped working. But thanks for noticing & fixing!

PR #3635, opened 2017-08-01
Firefox 56 will support preloading

danez:
https://developer.mozilla.org/en-US/Firefox/Releases/56

### DIFF
--- a/features-json/link-rel-preload.json
+++ b/features-json/link-rel-preload.json
@@ -97,8 +97,8 @@
       "53":"n",
       "54":"n",
       "55":"n",
-      "56":"y",
-      "57":"y"
+      "56":"a #1",
+      "57":"a #1"
     },
     "chrome":{
       "4":"n",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
While hooray, maybe its just partial?

<https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#Browser_compatibility>:

> [1] Firefox supports preloading of the following content (as values) —  script, style, image, video, audio, track, font, and fetch. The list of font types Firefox supports preloading of is defined in the fonts media type list (note font/collection is not supported).

<https://bugzilla.mozilla.org/show_bug.cgi?id=1222633>, e.g.:

> We do not support "per-document basis" for now. This is a lot of work that we cannot promise before 57. 

> (chrome supports no-cache as well)

danez:
Okay, so should I mark it as partial? and add a not saying only these certain types are supported?

Schweinepriester:
Maybe. I just wanted to start the discussion, as there apparently is a difference.

@Fyrd, what do you think?

danez:
It looks like `worker`, `object`, `document` and `embed` are not supported by looking at the examples in the spec. https://www.w3.org/TR/preload/#link-element-extensions

0gust1:
Thanks for the PR.
Android Firefox status could also be updated

Fyrd:
@danez Yeah, I think partial seems appropriate, with a note on what's still to do.

danez:
Ok changed.

Fyrd:
Great, thanks!

PR #3637, opened 2017-08-02
Updating Firefox status for preload links

0gust1:
Reference : https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#Browser_compatibility

### DIFF
--- a/features-json/link-rel-preload.json
+++ b/features-json/link-rel-preload.json
@@ -9,6 +9,10 @@
       "title":"Preload: What Is It Good For?"
     },
     {
+      "url":"https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content",
+      "title":"Preloading content with rel='preload'"
+    },
+    {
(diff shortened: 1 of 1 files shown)


0gust1:
ah sorry, didn't saw https://github.com/Fyrd/caniuse/pull/3635

features-json/link-rel-preload.json
```diff
@@ -270,7 +274,9 @@
       "59":"y"
     },
     "and_ff":{
-      "54":"n"
+      "54":"n",
+      "55":"n",
```
Schweinepriester: FYI:

<https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#unsupported-changes>

>
> - Add a new browser or browser version

0gust1: Thanks for pointing it out. 
Sorry, will do better next time.

PR #3647, opened 2017-08-08
Firefox 55 update

Utopiah:
source https://developer.mozilla.org/en-US/Firefox/Releases/55#New_APIs


Fyrd:
This should just be `y #1`, without the `d`.

Fyrd:
Thanks!

Issue #3483, opened 2017-06-02
Firefox, <input type="date">


tobireif:
Just in case it's news:

https://twitter.com/hsinyi71/status/870436875214782464

Calico90:
+1

tobireif:
Right now it's still only in Nightly, I think.

slokhorst:
Right now it's [marked](http://caniuse.com/#feat=input-datetime) as "on by default" (non-Nightly) for FF 56+, but in the [Mozilla wiki](https://wiki.mozilla.org/TPE_DOM/Date_time_input_types), it says 57+.

slokhorst:
This just landed in the stable [Firefox 57](https://developer.mozilla.org/nl/Firefox/Releases/57), so I think this issue can be closed.

PR #3648, opened 2017-08-08
Input datetime is not yet available in Firefox 56

slokhorst:
It still has to be enabled by a flag, but will be available by default in 57.

You can check that is not yet available in the current Firefox 56 beta.

Here it's mentioned that it will be available by default in Firefox 57:
https://wiki.mozilla.org/TPE_DOM/Date_time_input_types


slokhorst:
Related: #3483

Fyrd:
Thanks!

PR #3654, opened 2017-08-09
Update style-scoped.json

mattbrundage:
Firefox support clarification, per https://developer.mozilla.org/en-US/Firefox/Releases/55#HTML_2 (confirmed in v. 55)


Fyrd:
Thanks!

Issue #1278, opened 2015-05-12
Similar to MPEG-4 / WebM support page, you might want to add a page for MPEG-DASH support.  See:   A...


Fyrd:
Similar to MPEG-4 / WebM support page, you might want to add a page for MPEG-DASH support.

See:
http://stackoverflow.com/questions/22292224/which-web-browsers-support-the-mpeg-dash-streaming-protocol

Also, MSE support would be good to explain.

---
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)

Moderator votes: **+5**
Issue added by **ginger** on **2014-11-09**
To vote this issue up or down, simply include `+1` or `-1` in your comment.

sprice:
+1

jeremyshantz:
+1

jrstanley:
+1

ysbcc:
+1

Fyrd:
Now available at https://caniuse.com/#feat=mpeg-dash

PR #3658, opened 2017-08-11
Add MPEG-DASH

MekliCZ:


MekliCZ:
Fixes #1278 :)

Fyrd:
Thanks!

Fyrd:
@MekliCZ This information doesn't seem to be correct... for example I'm able to play videos using http://dash-mse-test.appspot.com/dash-player.html in multiple browsers you say don't have support. Also 
https://bugs.chromium.org/p/chromium/issues/detail?id=109652 states support has existed in Chrome since at least version 37.

MekliCZ:
The example you sent is using Media Source Extensions, it's not about native DASH support. You can use Shaka Player or Dash.js, but the browser doesn't support DASH natively. The same is Chromium bug, they are pointing to demo you sent, which is using MSE.

Fyrd:
@MekliCZ Ah, got it, thank you. Yeah the whole thing's kind of confusing and it's hard to find a demo out there that relies only on native DASH support.

MekliCZ:
@Fyrd Totally agree. It should be possible to check if browser supports DASH natively just by putting DASH stream to HTML5 video as a source. Or you could use JavaScript: ```video.canPlayType('application/dash+xml')```

Fyrd:
@MekliCZ Thanks! That helps a lot. Feature's available on the site now. :)

PR #3659, opened 2017-08-11
Added Chrome 61 for Android Support

dotjersh:
According to https://www.chromestatus.com/feature/5365692190687232, it's enabled by default in chrome 61 for android.


Schweinepriester:
Sadly per <https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#unsupported-changes>:

> ### Unsupported changes
> 
> * Add a new browser or browser version

As Chrome (for desktop) already has the correct state, I'm sure Fyrd will update Chrome for Android when its updated to 61.

mkurz:
I think this can be closed.

PR #3679, opened 2017-08-23
STP 38 supports parsing of the font-display property

daytonlowell:


Fyrd:
Haven't tried it yet, does this actually mean there's support for it? Seems like just "parsing" may just be a first step in supporting it.

daytonlowell:
Ah, yeah sorry, you’re right. Here is the umbrella bug for it https://bugs.webkit.org/show_bug.cgi?id=167332

PR #3684, opened 2017-08-25
Update css-marker-pseudo

myakura:
[WebKit recently landed support](https://trac.webkit.org/changeset/220207/webkit) for it and it's now available on Safari TP 38.


Fyrd:
Thanks!

PR #3688, opened 2017-08-26
getUserMedia/Stream API surely in Safari TP, too

Schweinepriester:
… best would be to prove this by testing 0:-)


Fyrd:
Thanks! Quick test shows indeed that it is supported.

PR #3689, opened 2017-08-27
Update datalist support for chrome 61, 62 and 63

Dexesttp:
The datalist scroll bug in chrome has been fixed in mid july, and has landed in Chrome 61.
https://bugs.chromium.org/p/chromium/issues/detail?id=375637#c77


Fyrd:
Thanks!

PR #3695, opened 2017-08-29
word-break: break-word exists in W3C text module

irekrog:
There is `word-break: break-word` so this is no longer unofficial. -> https://www.w3.org/TR/css-text-3/#word-break-property


Fyrd:
Thanks!

PR #3704, opened 2017-09-06
Add full support for scrollIntoView in Chrome 61+

styfle:
This was added in [CR Bug #648446](https://bugs.chromium.org/p/chromium/issues/detail?id=648446) in Chrome 61.

### Example

```js
document.querySelector('h1').scrollIntoView({'behavior': 'smooth'})
```


Fyrd:
Thanks!

PR #3706, opened 2017-09-06
css-fixed: Safari 9.1 doesn't draw element with `position: fixed` that has animated parent

malthejorgensen:
Safari 9.1 doesn't draw elements with `position: fixed` that have an
animated parent. The element will show up on redraws (e.g. resizing the
window).

Works in Safari 5.1, 6.2, 7.1, 8, and 10.1. I have not been able to test
Safari 9.0.

Screenshots:
https://www.browserstack.com/screenshots/05a0ae8ea84eab952eb13de45e1de4108fe8822e

jsbin:
https://output.jsbin.com/fuxipax

### DIFF
--- a/features-json/css-fixed.json
+++ b/features-json/css-fixed.json
@@ -181,7 +181,7 @@
       "7.1":"y",
       "8":"y",
       "9":"y",
-      "9.1":"y",
+      "9.1":"y #4",
       "10":"y",
       "10.1":"y",
       "11":"y",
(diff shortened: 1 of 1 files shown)


malthejorgensen:
Also haven't been able to find any issue on the WebKit/Safari bugtracker.

I believe this person has seen the problem as well: https://stackoverflow.com/a/37313620
And a similar issue with Safari lazy repaints (but not the same as that is in ): https://dzone.com/articles/safaris-lazy-repaint-bug-ios-7

PR #3709, opened 2017-09-07
web-share: Fix support data for Chrome and Chrome on Android

Turbo87:
The just released Chrome 61 enabled the feature by default but from https://www.chromestatus.com/feature/5668769141620736 it looks like the feature is only available on Android at this point. (My Chrome 61 on desktop doesn't know about `navigator.share` either)


Schweinepriester:
Sadly per <https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#unsupported-changes>:

> ### Unsupported changes
> 
> * Add a new browser or browser version

You are correct though, <https://www.chromestatus.com/feature/5668769141620736> only states:

> Enabled by default (launch bug) in:
Chrome for Android release 61

Turbo87:
@Schweinepriester good point, apparently I missed that. how can we move this forward?

Fyrd:
This has been taken care of, so closing.

Turbo87:
alright, thanks :)

PR #3710, opened 2017-09-07
css-mixblendmode: Add known issues for Chrome and Opera

lumio:
Chrome (including current Chromium) and Opera render mix-blend-mode differently in SVG when non-SVG elements are present.

- [Incorrect rendering](https://codepen.io/lumio/pen/YxoPWo)
- [Correct rendering](https://codepen.io/lumio/pen/vJqExB)


Fyrd:
I'd suggest combining the two into one and just refer to "Chrome/Opera".

lumio:
@Fyrd ... yep, did combine those two. They both use Blink - so it might be the same issue anyway

Fyrd:
Thanks!

PR #3724, opened 2017-09-14
Preload surely in iOS 11, too

Schweinepriester:
… best would be to prove this by testing 0:-)

At least <https://www.igeeksblog.com/how-to-enable-safari-experimental-features-in-ios-11/> mentions it (as "Link Preload").


Fyrd:
Yeah, seems safe to assume! Will test again when it's out.

PR #3729, opened 2017-09-15
Chrome 61 adds web share api

awebdeveloper:
https://developers.google.com/web/updates/2016/10/navigator-share

https://en.wikipedia.org/wiki/Google_Chrome_version_history


Fyrd:
Not on desktop, see https://bugs.chromium.org/p/chromium/issues/detail?id=764361: "Web Share is not anywhere near launching in Desktop".

PR #3731, opened 2017-09-16
Add Opus codec support for Safari 11, TP & iOS11

fixedmachine:
Opus Codec is now supported in Safari 11, TP & iOS11. It's support is mandatory for WebRTC and new versions of Apple browsers got WebRTC onboard.


Fyrd:
Thanks!

slokhorst:
Is it true that this only works with WebRTC?

My own test show that the following does *not* work in Safari 11:
* Opus in `<audio>` (with a `ogg/opus` or `webm/opus` source) does not work
* Opus with MSE does not work
  * `MediaSource.isTypeSupported('audio/webm; codecs="opus"')` returns `false`
  * `MediaSource.isTypeSupported('audio/ogg; codecs="opus"')` returns `false`
  * `MediaSource.isTypeSupported('video/webm; codecs="opus, vp9"')` returns `false`

If so, I think we should change it and show Safari having "partial support".

PR #3734, opened 2017-09-20
Fixes typo; adds various Safari experimental features that can be enabled

rgeraldporter:
Catches up featuresets to TP 39 and iOS 11 experimental features that are disabled by default but can be enabled.


Fyrd:
Thanks!

PR #3742, opened 2017-09-22
objectrtc (ORTC) not in iOS Safari 11, too (#3497 missed iOS)

Schweinepriester:
Title.

#3497 basically missed iOS. Also searched again for support, but found none.


Fyrd:
Just found out where this comes from...I believe I took the test from html5test.com, which checks for `window.RTCIceTransport`. This returns true only in Edge and Safari 11. I see it's in the WebRTC spec too now though, so that probably explains why the false positive.

PR #3744, opened 2017-09-22
Web Cryptography: Align Safari TP to 11

Schweinepriester:
I even checked using <https://diafygi.github.io/webcrypto-examples/> :)


Fyrd:
Thanks!

PR #3748, opened 2017-09-26
IE Edge and Mobile require prefix for appearance property

pkuczynski:


Fyrd:
As note <span>#</span>2 states, the support requires the `-webkit-` prefix. If an `x` was added, the site should show the `-ms-` prefix which doesn't work.

pkuczynski:
It's true, but in the table, it does not indicate it needs a prefix, which might be misleading...

PR #3760, opened 2017-10-02
Add :focus-ring pseudo class

robinwhittleton:
Currently this is only supported in Firefox with a prefix. There’s a spec in the draft Selectors 4 module: https://drafts.csswg.org/selectors-4/#the-focusring-pseudo, and bug IDs for the other browsers’ upcoming implementations in the JSON.

A testcase can be found at https://codepen.io/anon/pen/XeeJMo.

Fixes https://github.com/Fyrd/caniuse/issues/3267.


Fyrd:
Thanks!

PR #3765, opened 2017-10-04
STP 41 removes CSS Regions support

daytonlowell:
https://webkit.org/blog/7989/release-notes-for-safari-technology-preview-41/


Fyrd:
This makes me sad. But thanks for the PR!

PR #3770, opened 2017-10-10
Microsoft Edge does not support IndexedDB inside blob web workers

alexpusch:
Microsoft Edge does not support IndexedDB inside blob web workers. [See issue](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/5942817/).


Fyrd:
Thanks!

ameesme:
Just tested this on Microsoft Edge 41.16299.15.0, and can confirm it working using [this testing demo](https://bl.ocks.org/nolanlawson/863e464f3025d7199a1f).

screen shot 2017-12-30 at 18 01 37

Can anyone else confirm this? It's an old issue that finally seems to be fixed, even though this was not reported by Microsoft-staff.

PR #3771, opened 2017-10-12
Update transforms3d.json

Ricozor:
Safari 9, 10 & 11 are reported to still require a prefix for the related `backface-visibility` property.


Fyrd:
Thanks!

PR #3773, opened 2017-10-12
Update TLS 1.3 support for Chrome on Android

ewoerner:
Chrome on Android 61 supports TLS 1.3 behind a flag, as verified by visiting https://www.ssllabs.com/ssltest/viewMyClient.html with the flag enabled.
Removed unreferenced notes so that the first note is #1.


Fyrd:
Thanks!

PR #3779, opened 2017-10-17
Edge 16 supports es6-module

slokhorst:
JavaScript modules with `<script type=module>` were added to Edge 16, which was released today.
https://developer.microsoft.com/en-us/microsoft-edge/platform/status/moduleses6/


Fyrd:
Thanks!

PR #3785, opened 2017-10-18
Edge 16 fully supports Streams

dstorey:
Writable Streams was added in Edge 16.
I didn't change previous versions, as there is no note saying what basic Readable Streams means, but Edge 16 fully supports the Streams API


chicoxyzzy:
Great news! Does it support streaming fetch?

Fyrd:
@dstorey According to my tests it's still missing the pipeTo & pipeThrough methods, as well as BYOB (though no other browser has that yet either). Have updated the entry accordingly.

btw, Fetch support has been around since Edge 14... perhaps you got mixed up with something else? :)
https://developer.microsoft.com/en-us/microsoft-edge/platform/status/fetchapi/

PR #3810, opened 2017-10-25
FF 58+ enables font-display on all platforms

Schweinepriester:
<https://developer.mozilla.org/en-US/Firefox/Releases/58>:

> The `font-display` descriptor is now available by default on all platforms

🎉


Fyrd:
Awesome news. Thanks!

Issue #2812, opened 2016-09-14
Variable Open Type Fonts


matt7621:
Would love to follow the progress on this. https://medium.com/@tiro/https-medium-com-tiro-introducing-opentype-variable-fonts-12ba6cd2369#.4asds3ayo

Schweinepriester:
+1

https://twitter.com/bram_stein/status/782135926147411968

> Variation Fonts "In Development" in WebKit.

aaronbarker:
+1

jensimmons:
+100

jpamental:
Yes - would be hugely helpful!

Calico90:
+1

PeterConstable:
+1

thlinard:
+1

zachleat:
Looks like they’re in Safari 11, which is stable now I think?

thlinard:
Safari 11 on iOS 11 and macOS 10.13 (High Sierra), not Safari 11 on macOS 10.12 (Sierra).

jrrdnx:
+1

richardbruskowski:
+1

jpamental:
Variable Font support is now live in both Safari (desktop 10.13 and iOS 11) and Chrome, and in development in Edge - would be SUPER helpful to have this!

jensimmons:
Yes, we really need this one. @jpamental you should create a pull request. It's not too hard. Directions are here: https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md Grab a copy of the template and fill out the info. Having one of us do so will make this happen much sooner.

jpamental:
Thanks @jensimmons I'll give it a shot!

jensimmons:
@jpamental It’s incredibly satisfying. :) I was quite proud when a page I made was launched. Let me know if you get stuck, I’m happy to help.

jpamental:
PR opened @jensimmons ! Thanks for the nudge :)

Fyrd:
Now available at https://caniuse.com/#feat=variable-fonts

jpamental:
That’s awesome - thanks!

Jason

PR #3813, opened 2017-10-26
Adding ‘variable fonts’ feature

jpamental:
This is related to #2812 and #2893 

I've added information about support across all shipping browsers to the best of my knowledge.


features-json/variable-fonts.json
```diff
@@ -0,0 +1,318 @@
+{
+  "title":"Variable fonts",
+  "description":"Font format that supports",
```
Fyrd: Looks like the description wasn't quite completed?

features-json/variable-fonts.json
```diff
@@ -0,0 +1,318 @@
...
+      "52":"n",
+      "53":"n",
+      "54":"n",
+      "55":"n",
+      "56":"n",
+      "57":"n",
+      "58":"y x #1",
+      "59":"y x #1",
+      "60":"y x #1",
+      "61":"y x #1",
```
Fyrd: If the feature's only supported behind a flag, the value should be `n d #1`. Also the `x` refers to requiring a prefix, is that correct here?

features-json/variable-fonts.json
```diff
@@ -0,0 +1,318 @@
...
+      "url":"https://bugzilla.mozilla.org/show_bug.cgi?id=1302685",
+      "title":"Firefox status"
+    }
+  ],
+  "bugs":[
+
+  ],
+  "categories":[
+    "CSS",
+    "JS API"
```
Fyrd: Is there a JavaScript API that's part of this? If not, this can just be "CSS".

features-json/variable-fonts.json
jpamental: Thanks @Fyrd & @thlinard - I've addressed all of the above comments (Fyrd it is indeed a flag and not prefixing, so I've updated all the relevant entries). Let me know if there's anything more I need to do/fix!

Fyrd:
Thanks!

PR #3822, opened 2017-10-29
Datalist still buggy on Chrome 62

LorincJuraj:
Following bug is still present on Chrome 62:
"Partial support refers to a bug where long lists of items are unscrollable resulting in unselectable options."
Tested on Ubuntu 17.10/Chrome 62.0.3202.62 and using datalist element on https://html5tutorial.info/html5-datalist.php


Fyrd:
Thanks!

PR #3824, opened 2017-10-30
firefox 47+ supports u2f behind flag

keithamus:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1231681


Fyrd:
If it's behind a flag, it should be `n d #2`. Thanks!

keithamus:
like this?

Fyrd:
Yup!

PR #3829, opened 2017-11-02
Preconnect is partially supported in Edge 15+

nolanlawson:
Edge 15+ supports preconnect as an HTTP header, but not yet as a `<link rel>` tag.


Fyrd:
Thanks!

Issue #3808, opened 2017-10-24
IE 11 doesn't support assigning to classList


RusinovAnton:
```
// These two cases has same behavior
element.classList = 'my-classname';
element.className = 'my-classname';
```
So apparently major browsers are supporting assignment to classList.
Most popular sites say that `classList` is read-only. I haven't found much info on this being a setter in spec, just:
- test results from [w3c-test.org](http://w3c-test.org/dom/nodes/Element-classlist.html)
- and [mdn api reference](http://mdn.beonex.com/en/DOM/element.className.html#Summary) that say:
> className gets and sets the value of the class attribute of the specified element.  

The problem is that IE 11 ignores assignment to classList. So I suggest to add note about it in 
https://github.com/Fyrd/caniuse/blob/8c9d2d0e7f1fa9df091b5886e60862acfcf7d599/features-json/classlist.json#L319

![image](https://user-images.githubusercontent.com/13165540/31968080-9a4f509c-b918-11e7-8e37-1f542e522b03.png)

RusinovAnton:
Also [bryanbraun](https://stackoverflow.com/questions/15040297/clear-element-classlist#comment64208559_15040335) said on stackoverflow that he got an error `TypeError: Attempted to assign to readonly property.` in Safari (Version 9.1.1).

Ugoku:
> The problem is that IE 11 ignores assignment to classList

No, the problem is not using `classList` as intended :)
You should use [`classList.add('foo')`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) instead of doing `classList = 'foo'`

RusinovAnton:
@Ugoku but here is test passing in spec tests and as I said modern browsers are supporting it.
http://w3c-test.org/dom/nodes/Element-classlist.html
> Pass Assigning to classList (HTML node)

markadrake:
It's my understanding any browser allowing `classList` to be assigned to is not performing to spec. It's explicitly set to read-only, and `DomTokenList` adds it's own methods to interact with the object itself.

![image](https://user-images.githubusercontent.com/713355/34317554-82a38ebe-e77f-11e7-941d-9553b48286d2.png)

https://dom.spec.whatwg.org/#idl-index

![image](https://user-images.githubusercontent.com/713355/34317555-8b8e1cec-e77f-11e7-907c-a0ab10f17970.png)

https://dom.spec.whatwg.org/#interface-domtokenlist

cpplearner:
[...]

PR #3830, opened 2017-11-02
WIP Add note for IE11 about assign to classList

RusinovAnton:
refs: https://github.com/Fyrd/caniuse/issues/3808

TODO:
- [ ] - test other browsers

### DIFF
--- a/features-json/classlist.json
+++ b/features-json/classlist.json
@@ -46,7 +46,7 @@
       "8":"p",
       "9":"p",
       "10":"a #1 #2 #3",
-      "11":"a #1 #2 #3"
+      "11":"a #1 #2 #3 #4"
     },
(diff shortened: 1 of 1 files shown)


Fyrd:
Want me to merge this now or wait until other browsers have been tested too?

RusinovAnton:
I think it could be another PR with tested browsers so you can merge this one

dstorey:
classList is read only per spec, but now there is a putForward=value which puts forward the value to set value on DOMTokenList (not actually classList) instead. I implemented this in the upcoming version of Edge.

RusinovAnton:
@dstorey Nice. I would be really grateful if you could give me reference where I could any see info, discussion etc on why this implemented that way.

PR #3835, opened 2017-11-06
Update background-attachment.json

mypplication:
background-attachment:fixed supported from Android Chrome 62

https://pastebin.com/raw/YKwEcZuR


Fyrd:
Confirmed & updated manually.

PR #3836, opened 2017-11-06
add font-variant-east-asian

patrickkettner:


features-json/font-variant-east-asian.json
```diff
@@ -0,0 +1,307 @@
+{
+  "title":"CSS font-variant-east-asian ",
+  "description":"Controls the usage of alternate glyphs for East Asian scripts, like Japanese and Chinese.",
+  "spec":"https://www.digitalpreservation.gov/formats/fdd/fdd000114.shtml",
```
Fyrd: Spec URL doesn't seem to be correct...

Fyrd: Should be https://drafts.csswg.org/css-fonts-3/#propdef-font-variant-east-asian perhaps?

Fyrd:
Thanks!

PR #3841, opened 2017-11-08
Update Opera support for scrollintoview

anantoghosh:
From Opera 48 (based on Chromium 61), behavior: 'smooth' is supported.


Fyrd:
Thanks!

PR #3845, opened 2017-11-10
TLS 1.3 in Chrome for Android (56+), too

Schweinepriester:
Via https://www.chromestatus.com/feature/5712755738804224, this probably just wasnt updated here when the number for android was upped?


Fyrd:
Thanks!

PR #3849, opened 2017-11-12
URLSearchParams will land in Edge 17

mkurz:
The related Edge bug was just closed:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8993198/

> URLSearchParams has just been implemented. It will be included in a Windows Insider build in the near future.


mkurz:
Edge 17 needs to be added first, can you do that @Fyrd?

Fyrd:
Added and updated. Thanks!

PR #3864, opened 2017-11-19
Update dommatrix.json (Chrome support)

jdhoek:
According to MDN, Chrome from 61 onwards supports `DOMMatrix`.

I have tested this in Chromium 62:

```javascript
let m = new DOMMatrix([1.0432 ,0.24005, 0,0.00019034 ,-0.29276 ,0.37974, 0,-0.0023912 , 0, 0, 1, 0, 48 ,33, 0,1])
let p = new DOMPoint(522,172);
let p2 = m.transformPoint(p);

console.log(p2)
```

## Firefox

```
DOMPoint
  w: 0.688071072101593
  x: 542.1956787109375
  y: 223.62136840820312
  z: 0
```

## Chromium 62

```
DOMPoint {x: 542.19568, y: 223.62138, z: 0, w: 0.6880710800000001}
  w:0.6880710800000001
  x:542.19568
  y:223.62138
  z:0
```


Fyrd:
Ran it through my test and it seems to be missing support for the `scaleNonUniformSelf` method. Have updated appropriately. Thanks!

jdhoek:
Thanks!

By the way, I looked up `scaleNonUniformSelf` in the [standard](https://drafts.fxtf.org/geometry/#dom-dommatrix). The ChangeLog mentions:

> Changed the scale() and scaleSelf() methods to be more like the previous scaleNonUniform()/scaleNonUniformSelf() methods, and dropped the scaleNonUniform* methods. 

It looks `scaleNonUniformSelf` was dropped from the standard, so perhaps *partial support* on account of not having that method doesn't feel right here. What do you think?

Fyrd:
@jdhoek Ah, thanks for pointing that out! I thought I'd checked for that myself, but must have been looking at an older version. Have updated again now.

PR #3865, opened 2017-11-20
SVG Fragment identifiers - Safari Technology Preview now supports it.

qfel13:
It was fixed in release 40.
https://developer.apple.com/safari/technology-preview/release-notes/#r40


Fyrd:
Thanks!

PR #3868, opened 2017-11-22
Update safari support for rel="preload"

segmentationfaulter:
Safari support is there, check the links

https://webkit.org/status/#?search=preload
https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#Browser_compatibility


Fyrd:
It's still disabled by default in Safari 11 when I check the "Experimental features" menu option there, as pointed out on the current note there. So yes, support may be in WebKit but it's not visible to users by default.

PR #3872, opened 2017-11-23
Use HTTPS on more links

ngyikp:
There are still a lot of blogs/sites that I haven't checked yet, but let's do the easiest batch first.


Fyrd:
Thanks!

PR #3876, opened 2017-11-24
Correct Credential Management API support info for Samsung Internet

NOtherDev:
Samsung Internet 5 & 6 expose the interface but reject when trying to use it - it is not wired up to any actual implementation.
See https://twitter.com/poshaughnessy/status/933299590106927104.


Fyrd:
Thanks!

PR #3885, opened 2017-11-29
Add FF undraggable-unless-dataTransfer-is-set bug.

mhairston:
The bug text includes a link to a codepen containing a test case for this (as well as the FF can't-drag-a-button bug).

https://codepen.io/michai/pen/NwORqO


Fyrd:
Thanks!

PR #3890, opened 2017-11-30
Reduced status of font-loading to unoff

HolgerJeromin:
https://drafts.csswg.org/css-font-loading/

The spec is not a W3C Candidate Recommendation but an editors’ draft


Fyrd:
Thanks!

PR #3893, opened 2017-12-02
Chrome supports sticky on thead & tr with v64

mkurz:
See https://bugs.chromium.org/p/chromium/issues/detail?id=702927#c16

Also added a missing 5th note because some older Safari versions were refering to it but it didn't exist...


Fyrd:
Confirmed, thanks!

PR #3894, opened 2017-12-04
Correct Safari 11’s support for the Animations API

seanCodes:
Safari 11 actually supports the Web Animations API, but only as an experimental feature:

safari 11 web animations support

This change updates the support from none → partial.


Fyrd:
Thanks!

PR #3895, opened 2017-12-04
Update for SVG Backgrounds in Edge

PeterShaggyNoble:
Edge does not currently support data URIs for SVG backgrounds.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/6274479/


Fyrd:
Thanks!

PR #3903, opened 2017-12-07
Edge supports all the Form features

dstorey:
...since adding the form attribute


Schweinepriester:
What about <https://caniuse.com/#feat=datalist>?
At least its currently listed as subfeature.

@Fyrd:
If it is relevant, imho Chrome <63 should be partial in <https://caniuse.com/#feat=forms>, too.

dstorey:
We support datalist, just there are so usability issues (I think, the link doesn't exist any more), but I don't think they map to any spec bugs.

Fyrd:
Seems fair to me!

Fyrd:
@Schweinepriester I understand the logic there, but frankly `forms` isn't a terribly useful support table anyway, so making it partial just because one of the 13(!) sub-features has partial support seems a bit unnecessary.

PR #3910, opened 2017-12-10
Date input with min/max works fine

slokhorst:
This feature works as intended as of Chrome for Android 62. It was probably fixed in an earlier version.

I'm not sure in which version the bug was found, but the note was added in da91eb4b2fc5dad27bebd1e22e26ae20b211e5f3 (when Chrome 42 was stable).


Fyrd:
Confirmed, thanks!

PR #3913, opened 2017-12-12
Update css-transitions.json

Psithief:
IE11 did not start supporting background-size for transitions.


PeterShaggyNoble:
I suggest rewording it to read: "Internet Explorer does not support transitions of the `background-size` property."

Fyrd:
Thanks!

PR #3914, opened 2017-12-12
Web Share update

NOtherDev:
Assuming "Chrome" means Chrome on desktop, not "Chrome on any platform", Web Share was never available as an Origin Trial. It was only available on Android on Chrome 55-58 and is now turned on by default as of Chrome 61 (Android only, again). 

Origin Trials is Chrome-only thing, so I also removed erroneous #1 reference from Samsung Internet cells.

See [Origin Trial summary](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/rgIpGcOyDKI).


Fyrd:
Thanks for the corrections!

PR #3918, opened 2017-12-15
Update resource-timing.json

JohnRiv:
Discovered with the help of my coworkers, and confirmed on twitter (https://twitter.com/JohnRiv/status/930532827103399943) that Resource Timing in Safari is not supported on OSX prior to Sierra.

Technology Preview is only available for Sierra & up so Safari 11 should be the only partial support occurrence.


Fyrd:
Thanks!

PR #3923, opened 2017-12-19
CSS Backdrop Filter enabled by default in Edge

benaadams:
Windows 10 build 17063 Preview: https://developer.microsoft.com/en-us/microsoft-edge/platform/changelog/desktop/17063/


Fyrd:
Thanks!

PR #3924, opened 2017-12-19
Edge 17:  Service Workers, Push API, Ogg, SRI

Schweinepriester:
Per <https://developer.microsoft.com/en-us/microsoft-edge/platform/changelog/desktop/17063/?compareWith=16299>, section "New web platform features".

`backdrop-filter` is covered by #3923.

Push API seems indeed to be independent from Edge running per <https://blogs.windows.com/msedgedev/2017/12/19/service-workers-going-beyond-page/>, e.g.

> Compared to current web notifications, a push notification can be delivered to a user without needing the browser/app or page to be opened.

> As part of our ongoing commitment to expanding the possibilities of the web, Microsoft Edge and PWAs in Windows will handle these service worker push event handlers in the background—there’s no need for Microsoft Edge or your PWA to be running for the push to be handled. That’s because we’ve integrated with Windows to allow for a more holistic approach to push notifications.


Fyrd:
Nice, thanks!

PR #3928, opened 2017-12-21
Safari TP 46: Service Workers enabled by default

Schweinepriester:
<https://webkit.org/blog/8042/release-notes-for-safari-technology-preview-46/>

> Made Service Workers enabled by default 

Though

> **While work continues**, we’re excited to enable Service Workers by default in this release.

I was optimistic and chose `y`, but one could maybe(?) argue for partial, too.


Fyrd:
Thanks! I'm okay with going with `y` until it's obvious where it's partial.

PR #3932, opened 2017-12-22
Correct Chrome supported versions for promise rejection tracking

ewoerner:
Chrome supports unhandledrejection/rejectionhandled since version 49, not since version 36 (version 36 is the first supported version for Opera).
See also https://www.chromestatus.com/feature/4805872211460096


Fyrd:
Thanks!

PR #3933, opened 2017-12-22
Change demo page for battery status API

ewoerner:
Previous demo site only supports the outdated spec version.
New demo site supports both current and outdated spec version, and therefore works with Chrome.


Fyrd:
Thanks!

PR #3935, opened 2017-12-22
Update tabindex support for Android browsers

ewoerner:
Added Chrome on Android and Firefox on Android support for tabindex.
This can be verified using either a bluetooth keyboard or by using a virtual keyboard that exposes tab keys (e.g. Gboard exposes the forward tab on forms).


Fyrd:
Haven't tried it myself, but that does make sense. Thanks!

PR #3952, opened 2018-01-04
Clarify Firefox geolocation callback issue

ewoerner:
- Improve wording so that the issue doesn't come across as some random bug.
- The issue has also been fixed in Firefox 53 and above with the new permission prompt which cannot be dismissed without a permission decision.


Fyrd:
Thanks!

PR #3954, opened 2018-01-04
Added note that Opus support is High Sierra-only

daytonlowell:


Schweinepriester:
Would be nice if the note was more future-proof, for example „Supported on macOS 10.13 High Sierra and later“ 0:-)

features-json/opus.json
```diff
@@ -184,8 +184,8 @@
       "9.1":"n",
       "10":"n",
       "10.1":"n",
-      "11":"y",
-      "TP":"y"
+      "11":"y #1",
+      "TP":"y #1"
```
dstorey: This should probably be listed as partial. IE got downgraded due to not being included in all OS versions, e.g. https://caniuse.com/#feat=http2 and https://caniuse.com/#feat=mediasource

daytonlowell: Fair. If we do that, HEVC and FLAC will need the same treatment.

Schweinepriester: Yeah, how to handle OS-depended support is still inconsistent. I asked in #3572, but maybe this should be discussed on its own to find a solution for all occurrences?

Fyrd: I'd say as long as the browser is used on an OS version used by a significant amount of users it should be partial support. (as to what's a "significant" amount, ah...that gets tricky)

Fyrd: In this case MacOS Sierra still has lots of usage (even higher than High Sierra last month) I agree all these should be partial support, @daytonlowell.

Fyrd:
@daytonlowell Thanks!

slokhorst:
This is not true.. e.g. it doesn't work here: https://hpr.dogphilosophy.net/test/index.php

revert: #4146

PR #3993, opened 2018-01-11
HTTP/2: Full support for Safari 11+

Schweinepriester:
Because per <https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility> Safari 11+ is available starting with OS X 10.11, thus every version of Safari 11+ supports HTTP/2 per Note 3: <https://caniuse.com/#feat=http2>


Fyrd:
Thanks!

PR #3995, opened 2018-01-12
Edge 17 supports "Font Variation Properties with OpenType Variable Font Support"

mkurz:
See "New web platform features" here: https://developer.microsoft.com/en-us/microsoft-edge/platform/changelog/desktop/17074/

If I understand that right we can switch that to `y`...


Fyrd:
Yup, thanks!

PR #3996, opened 2018-01-12
Firefox 59 supports modules

pointlessone:
The feature was enabled by default in Firefox 59 nightlies on 2018-01-10.

* [Upstream ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1428002)
* [Announcement tweet](https://twitter.com/FirefoxNightly/status/951382754125545473)

### DIFF
--- a/features-json/es6-module.json
+++ b/features-json/es6-module.json
@@ -113,8 +113,8 @@
       "56":"n d #2",
       "57":"n d #2",
       "58":"n d #2",
-      "59":"n d #2",
-      "60":"n d #2"
+      "59":"y #4",
+      "60":"y"
     },
     "chrome":{
       "4":"n",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
I don't think Note 4 is necessary.

In general: I've always wondered how "Enable … in nightly builds" is meant - is it either the version which currently happens to be nightly will get this feature and carry it with itself to stable or from this point on every nightly version will have this feature enabled, but only the versions which are nightly until the feature is promoted to stable with a future version?

<https://bugzilla.mozilla.org/show_bug.cgi?id=1428002#c4>

> oh, this is nightly only... well, even then it would be good to notify dev.platform, but this isn't really intent-to-ship yet.

<https://bugzilla.mozilla.org/show_bug.cgi?id=1428002#c9>

> I think I can make it just skip this test on nightly.  I'll do that.

Lets me think modules won't be enabled by default in 59 when it goes to beta or even stable. But I'm not sure.

Fyrd:
Agreed, I don't think it's going to matter to anyone that it wasn't enabled in that version before a certain date.

PR #3998, opened 2018-01-12
tooShort and reportValidity are in Edge 17

dstorey:
(the latter hasn't reached insider preview yet, but I'm updating this while I have time. It is in our internal equivalent of nightlies. I'm not sure if you want to wait until it is public to merge or not, but it will be in Edge 17)


Fyrd:
I'm happy to take your word for it for now. :)

PR #4000, opened 2018-01-12
Update Edge info for Clipboard API

dstorey:
Edge 17 supports the full synchronous Clipboard API:
* DocumentAndElementEventHandlers mixin was created with oncopy/paste/cut
* clipboardData already existed on ClipboardEvent 
* Constructor was added to ClipboardEvent

#2 doesn't have any tests linked so haven't been able to test yet if that issue is resolved.


Yaffle:
>#2
yes, this issue is resolved, seems

PR #4002, opened 2018-01-12
Edge 17 now supports <link rel="preload">

dstorey:


Schweinepriester:
Surely this will soon popup in the changelog as well? 0:-)

<https://developer.microsoft.com/en-us/microsoft-edge/platform/status/preload/?q=preload> Is still under consideration?

Anyway: Hurray :)

Fyrd:
I see it's "in development" now...will again take your word that it'll be in 17. :)

PR #4005, opened 2018-01-14
Preload now a Candidate Recommendation

tomByrer:
as of 26 October 2017
https://www.w3.org/TR/preload/

(Also Edge 'in progress' but I guess you pull automatically from https://developer.microsoft.com/en-us/microsoft-edge/platform/status/preload/ )


Fyrd:
Thanks!

PR #4017, opened 2018-01-17
Update css-placeholder.json

mattbrundage:
Microsoft article was erroneously redirecting to the MDN :placeholder-shown page, rather than the MDN ::placeholder page.


Fyrd:
Thanks!

PR #4029, opened 2018-01-26
Safari 11.1+: Web App Manifest

Schweinepriester:
<https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html>:

> Safari 11.1 ships with iOS 11.3 and macOS 10.13.4. It is also available in macOS 10.12.6 and 10.11.6.

<https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html#//apple_ref/doc/uid/TP40014305-CH14-SW6>:

> - New in Safari 11.1—Web App Manifest
>   - Added support for websites declaring how web apps behave and appear when using “Add to Home Screen”.

<https://twitter.com/rmondello/status/956256845311590400>:

> iOS 11.3 and macOS 10.13.4 include Service Workers — a powerful specification that allows background scripts to power offline web applications. iOS 11.3 also consults Web App Manifest when adding web apps to the home screen.

The tweet suggests its iOS only, but the changelog suggests otherwise?

I was optimistic with this pull request 0\:-)


Fyrd:
Yeah sure, we'll see. :)

PR #4042, opened 2018-01-30
Removing spam link from css-masks links

hay:
The article on `thenittygritty.co` seems to be a spam link, i just get banners for some betting site instead of a CSS tutorial.

### DIFF
--- a/features-json/css-masks.json
+++ b/features-json/css-masks.json
@@ -13,10 +13,6 @@
       "title":"HTML5 Rocks article"
     },
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks! FWIW here's what used to be there:
https://web.archive.org/web/20160304015555/http://thenittygritty.co/css-masking

hay:
@Fyrd thanks! Shame that blog went out to the spammers, it actually looked quite informative :/

PR #4043, opened 2018-01-30
Update Safari data for image-set

csnardi:
image-set was unprefixed in Safari Technical Preview 9 (https://webkit.org/blog/6800/release-notes-for-safari-technology-preview-release-9), which was shipped in Safari 10 (I believe). Add a note that this is not completely spec-compliant.


Fyrd:
Thanks!

PR #4044, opened 2018-01-30
image-rendering: pixelated on FF

Janiczek:
Caniuse.com says Chrome supports `image-rendering: pixelated;` and Firefox supports `image-rendering: crisp-edges;`.

Trying out [this example](https://css-tricks.com/almanac/properties/i/image-rendering/#article-header-id-2) in Firefox, **Firefox actually supports `pixelated`** and not `crisp-edges` (similar to Chrome). And without the `-moz-` prefix.

Using Browserstack I checked that it is supported this way till Firefox 6.


Fyrd:
Ah, not actually. That pen sample is using autoprefixer I believe, which includes all properties needed to make it work cross-browser. If you inspect the element you can see what's going on:
screen shot 2018-01-30 at 8 57 42 pm

Janiczek:
Ah, sorry. I looked into the SCSS tab but didn't think to look into Inspector. Sorry for the false alarm :)

PR #4045, opened 2018-01-30
Updating dead link in Intersection Observer

MKRazz:
Not sure if it was better to link to GitHub or NPM. Here's the NPM link just in case:

https://www.npmjs.com/package/intersection-observer


Fyrd:
Thanks!

PR #4048, opened 2018-02-01
Update UC Browser stats(It now supports grid)

pantharshit00:
UC Browser 12.0 rolled out today and it supports css grid. Cheers!


pantharshit00:
Oops i cant add a browser version

PR #4049, opened 2018-02-01
chrome still supports prerender

paulirish:
To test:
open net-internals and open https://lodash.com/

image

while chrome team doesn't really LOVE the feature, it's not planned for removal just yet.

cc @patrickhulce


Fyrd:
@paulirish So the change "M58 is rolling to Stable with Prerendering disabled for 99% of users" from https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/0nSxuuv9bBw/PKu8Nv3yCAAJ  was undone?

Fyrd:
Will go ahead and merge since tests indeed indicate you're correct. :)

paulirish:
fair. I'll check in with the author of that thread to see what happened to that proposal.

Malvoz:
@paulirish 

https://github.com/Fyrd/caniuse/pull/4544 might be of your interest.

✌️

PR #4061, opened 2018-02-06
Chrome 65+ and Opera 52+ enable display:contents by default

SelenIT:
According to https://www.chromestatus.com/features/5663606012116992


Fyrd:
Thanks!

PR #4064, opened 2018-02-06
update csp2 link to spec

bergernetch:
The https://www.w3.org/TR/CSP link without version now redirects to the CSP3 spec, therefore for CSP2 the version must now be specified in the link.


Fyrd:
Thanks!

PR #4066, opened 2018-02-07
Update dead MS connect link

mkurz:
MS Connect is dead.

The bug was moved from https://connect.microsoft.com/IE/feedback/details/811408/ie10-11-a-textareas-placeholder-text-becomes-its-actual-value-when-using-innertext-innerhtml-or-outerhtml to https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/101525/

I also mention the bug at the innertext features because it is effected as well.

### DIFF
--- a/features-json/innertext.json
+++ b/features-json/innertext.json
@@ -26,7 +26,9 @@
     }
   ],
   "bugs":[
-    
+    {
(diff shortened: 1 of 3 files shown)


mkurz:
See #4067

Fyrd:
Thanks!

PR #4068, opened 2018-02-07
URL is implemented in IE 11

noelebrun:
Tested myself


Fyrd:
IE11 supports a URL object to access `createObjectURL()`, etc, which has its support tracked at https://caniuse.com/#feat=bloburls Perhaps that's what you were testing?

It does not support the URL API defined at https://url.spec.whatwg.org/#api

moos:
Can this be added as a note, that it supports the static method `URL.createObjectURL()`?  I believe even IE10 supports it.

PR #4073, opened 2018-02-08
Warn others of Safari 10's let bug

tmcw:
Hello!

- Here's the WebKit thread: https://bugs.webkit.org/show_bug.cgi?id=171041
- And the uglify-es thread: https://github.com/mishoo/UglifyJS2/issues/1753

I tried to word this as best as I could but it's tricky to explain: basically `let` bindings in for loops are broken in Safari 10: they act like `var` bindings in terms of scoping, but `let` bindings in terms of allowing a single initialization. This causes crashing bugs in typical code.

### DIFF
--- a/features-json/let.json
+++ b/features-json/let.json
@@ -180,8 +180,8 @@
       "8":"n",
       "9":"n",
       "9.1":"n",
-      "10":"y",
-      "10.1":"y",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #4075, opened 2018-02-08
rellist on <a> too for Chrome 65+ & Opera 52+

Schweinepriester:
Per <https://blog.chromium.org/2018/02/chrome-65-beta-css-paint-api-and.html>

> Chrome now supports the `HTMLAnchorElement.relList` property to indicate the relationship between the resource represented by the `<a>` element and the current document. Thanks to Samsung for this contribution!

Also <https://www.chromestatus.com/feature/5631885325631488>.

Interestingly they didn't mark Android Webview or Chrome for Android, but Opera for Android… oO


Fyrd:
Thanks!

PR #4080, opened 2018-02-12
Update css-transitions.json

fdalvi:
Safari 11 (Webkit specifically) has a bug where CSS transitions and
animations on the flex-basis property are ignored.


Fyrd:
Thanks!

PR #4091, opened 2018-02-19
Better description for `display:flow-root`

SelenIT:
The statement "The result [of display:flow-root] is the same as the "clearfix" hack" is misleading. The classic "clearfix" hack doesn't establish the new BFC. These two approaches result in the fundamentally different behavior regarding floats going _before_ the "clearfixed" container in the flow and margin collapsing of the descendants of that container ([demo of this difference](https://codepen.io/SelenIT/details/qrORXm)). I suggest different wording saying that it is the better (and standard) solution to the same problem, but not the full equivalent of "clearfix".


Fyrd:
Sounds good. Thanks!

PR #4092, opened 2018-02-21
Switch Edge back to y for variable font support

gregwhitworth:
I'm not entirely sure why this was listed as having issue #4 as we have support for `truetype-variations`, `woff-variations`, `woff-variations` and `opentype-variations`. Additionally, we support the `font-weight` and `font-stretch` properties and their aliasing to their respective axes within the font.


gregwhitworth:
That said, Chrome does have the issues in #4, as well as not supporting local() variable font references. At any rate my thinking is this must have been a mistake or I'm missing something. Thanks for taking the time to review @Fyrd

Fyrd:
@gregwhitworth Thanks for confirming! That note was based on [this article](https://medium.com/clear-left-thinking/how-to-use-variable-fonts-in-the-real-world-e6d73065a604#2093) which suggests Edge 17 has the same shortcomings as the other browsers.

PR #4096, opened 2018-02-23
Bug has been fixed in Edge

mkurz:
I am 100% sure that has been fixed in Edge - because I was the one who added that notice but on an other feature, someone copied over wrong.
Also see [here](https://github.com/Fyrd/caniuse/pull/4066/files) - it's mentioned three times but once wrong.


Fyrd:
Thanks!

PR #4103, opened 2018-02-28
multicolumn.json - Add helpful workaround details to note about Firefox support

dhoulb:
Add to note that `page-break-*` properties in Firefox can accomplish the same result as `break-` properties, as noted in https://bugzilla.mozilla.org/show_bug.cgi?id=549114 and tested in latest Firefox. Not sure how long this support has been available.


Fyrd:
Thanks!

PR #4108, opened 2018-03-03
css shapes fixes

dartess:
1) add firefox support (partially and flagged)
2) fix description and keywords in accordance with spec level 1


Fyrd:
Thanks!

PR #4111, opened 2018-03-04
Mark ES6 modules as supported by FireFox 60 and onwards

jonco3:
ES6 modules are enabled in version 60, see:

https://bugzilla.mozilla.org/show_bug.cgi?id=1438139


Fyrd:
Thanks!

PR #4112, opened 2018-03-05
CSP2 not supported in Safari for iOS 9.3

mkurz:
We just ran into that.
I tested with an iPad with iOS 9.3.5 and we are 100% sure that CSP2 doesn't work on it (we tested hash-source, didnt work).

To back up my claim look at the Safari 10 release notes: https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_0.html
> Content Security Policy (CSP) support has been enhanced by including version 2.0 of the standard.

So it starts with Safari 10, not 9.3.


Fyrd:
Confirmed, thanks!

PR #4118, opened 2018-03-12
Clarify Chrome Variable Fonts Support

drott:
format("*-variations") supported from M66, font-stretch and font-weight
in @font-face working, compare https://codepen.io/anon/pen/bvdOgV

### DIFF
--- a/features-json/variable-fonts.json
+++ b/features-json/variable-fonts.json
@@ -97,15 +97,15 @@
       "50":"n",
       "51":"n",
       "52":"n",
-      "53":"n d #2 #4",
-      "54":"n d #2 #4",
-      "55":"n d #2 #4",
-      "56":"n d #2 #4",
-      "57":"n d #2 #4",
-      "58":"n d #2 #4",
-      "59":"n d #2 #4",
-      "60":"n d #2 #4",
-      "61":"n d #2 #4"
+      "53":"n d #2 #4 #5",
+      "54":"n d #2 #4 #5",
+      "55":"n d #2 #4 #5",
+      "56":"n d #2 #4 #5",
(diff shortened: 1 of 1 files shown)


drott:
CC @brawer

features-json/variable-fonts.json
```diff
@@ -170,8 +170,8 @@
       "63":"a #4",
       "64":"a #4",
       "65":"a #4",
-      "66":"a #4",
```
Fyrd: This can change to `y` if there are no further issues. Else there should be a note describing why support is still partial.

features-json/variable-fonts.json
```diff
@@ -317,7 +317,7 @@
     "1":"Works with Experimental Web Platform features enabled",
     "2":"Requires MacOS 10.12+ and the following about:config flags to be enabled:\r\n`layout.css.font-variations.enabled`,\r\n`gfx.downloadable_fonts.keep_variation_tables`",
     "3":"Requires MacOS 10.13+",
-    "4":"Does not support the `font-weight` and `font-stretch` properties, nor `format('truetype-variations')`"
```
Fyrd: I believe this note also applies to Firefox, so you should probably change the one that applies to Chrome (& Opera) if it is incorrect.

drott: Done, I separated the notes, and tested additional format strings in FF, then updated FF accordingly.

features-json/variable-fonts.json
drott: Done.

drott:
Thanks for the review, comments (hopefully) addressed.

Fyrd:
@drott Looks great, thanks!

PR #4120, opened 2018-03-14
Opera since 42 don't support h264

bes-internal:


Fyrd:
I looked into this and it seems to be true only for Opera on Linux (though there appear to be workarounds). Instead of changing the support I'd suggest adding that to known issues instead?

bes-internal:
Test at [html5test.com](html5test.com) and this mp4 [file](http://techslides.com/demos/sample-videos/small.mp4) from [this page](http://techslides.com/sample-webm-ogg-and-mp4-video-files-for-html5) for playing. 

Opera 52 linux; Opera 51, 52, 53 Windows 10; Firefox 57, 58, 59 Windows 10
```
Video codecs
MPEG-4 ASP support No ✘
H.264 support No ✘
H.265 support No ✘
Ogg Theora support Yes ✔
WebM with VP8 support Yes ✔
WebM with VP9 support Yes ✔
```

Such a feeling that this is a problem on windows 10...

Fyrd:
That's not what I see...just tried it both on Windows 10 and MacOS. In both cases html5test.com and the file from that page play just fine in Opera.

bes-internal:
I sampled on 8 machines in our office and got random support results regardless of the operating system for Opera and Firefox browsers. Stable result shows only the Chrome that always support h264
I close this PR because browsers provide software support

PR #4126, opened 2018-03-15
Chrome/Opera both support unprefixed WebRTC APIs for a while

nazar-pc:
https://www.chromestatus.com/feature/5312661013135360


Fyrd:
Confirmed, thanks!

PR #4133, opened 2018-03-16
Change title of html5semantic.json; add relevant keywords

jake-low:
This is just a bit of housekeeping work. I noticed that `html5semantic.json` didn't have any keywords associated with it, so I added some. I also changed the title since "new" is a relative term that will grow less and less correct over time.

Comments welcome.

### DIFF
--- a/features-json/html5semantic.json
+++ b/features-json/html5semantic.json
@@ -1,5 +1,5 @@
 {
-  "title":"New semantic elements",
+  "title":"HTML5 semantic elements",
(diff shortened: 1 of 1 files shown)


Fyrd:
The search already looks inside the "description" field so the keywords aren't actually necessary. The change to the title is a good one though.

jake-low:
Cool. Should I remove the keyword change from the PR?

Fyrd:
Yeah, if you don't mind. :)

jake-low:
Not at all, just updated the commit. Thanks for the feedback.

Fyrd:
Thanks!

PR #4134, opened 2018-03-16
Update for Chrome Android missing in #4126

nazar-pc:
I forgot to update Chrome Android in #4126, it also supports unprefixed WebRTC APIs


Fyrd:
Thanks!

PR #4141, opened 2018-03-20
Correct Firefox flag information for WebComponents

ExE-Boss:
Firefox 59 removed the `dom.webcomponents.enabled` preference and split it into `dom.webcomponents.shadowdom.enabled` and `dom.webcomponents.customelements.enabled`.

Also, custom elements are not yet enabled in Firefox 59 stable.

# See also:
- mdn/browser-compat-data#1294
- [https://bugzil.la/1400762](https://bugzil.la/1400762 "1400762 - Deprecate dom.webcomponents.enabled and only use dom.webcomponents.customelements.enabled for Custom Elements")
- [https://bugzil.la/1428685](https://bugzil.la/1428685 "1428685 - Consider renaming pref dom.webcomponents.enabled to dom.webcomponents.shadowdom.enabled")

### DIFF
--- a/features-json/custom-elements.json
+++ b/features-json/custom-elements.json
@@ -126,9 +126,9 @@
       "56":"p d #1",
       "57":"p d #1",
       "58":"p d #1",
-      "59":"p d #1",
-      "60":"p d #1",
-      "61":"p d #1"
+      "59":"p d #2",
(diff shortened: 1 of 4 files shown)


Fyrd:
Thanks!

PR #4146, opened 2018-03-22
Opus in the <audio> element is NOT supported in Safari

slokhorst:
Revert #3731

As noted in https://caniuse.com/#feat=opus : "Support refers to this format's use in the audio element, not other conditions."

Safari simply does not support Opus in the audio element. Test for instance with the file under "Example of use in a web page" on http://opus-codec.org/examples/ or with the Opus Audio file from https://hpr.dogphilosophy.net/test/ .

More details in https://github.com/Fyrd/caniuse/pull/3731#issuecomment-354119681

### DIFF
--- a/features-json/opus.json
+++ b/features-json/opus.json
@@ -187,9 +187,9 @@
       "9.1":"n",
       "10":"n",
       "10.1":"n",
-      "11":"a #1",
-      "11.1":"a #1",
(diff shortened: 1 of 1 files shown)


Fyrd:
Yup, confirmed. Thanks!

PR #4155, opened 2018-03-27
Add feature: Navigator.doNotTrack JS API

jake-low:
Addresses part of #2642.

I made a pen to test support for this: https://codepen.io/jakelow/pen/xWPXzJ

Desktop browser support info is based on research (browser bug trackers and StackOverflow mostly; I can provide citations if you're interested).

Mobile browser support info is limited; I was only able to test on two devices (one iOS 10.3 and one iOS 11.2).

If someone with a BrowserStack account could run through a few versions of different Android browsers (Chrome/FF/Samsung/UC ideally) and fill in some gaps, that would be awesome.


Fyrd:
Thanks!

dstorey:
This is not correct for Edge 17. We moved from Window for that release. https://developer.microsoft.com/en-us/microsoft-edge/platform/catalog/?page=1&q=donottrack

PR #4160, opened 2018-03-29
Update css-focus-visible.json

robdodson:
Hey @Fyrd !

Here's a bit of an update for :focus-visible (sorry for the noise on my earlier PR attempt). This updates a few of the bugs, and adds a note that the feature works in chrome behind a flag.


features-json/css-focus-visible.json
```diff
@@ -177,8 +185,8 @@
       "64":"n",
       "65":"n",
       "66":"n",
-      "67":"n",
-      "68":"n"
+      "67":"p d #2",
+      "68":"p d #2"
```
jake-low: Forgive my drive-by reviewing: Do you mean `"a d #2"` here? `p` means "No support, but has polyfill", whereas `a` means "almost supported" and corresponds to "partial support" in the caniuse.com UI.

robdodson: oh, yes I think that's what I want then :) I'll update. Thanks!

robdodson:
Should be fixed now :)

Fyrd:
@robdodson `a d` actually means there's currently partial support, with "something else" available behind a flag. If there's no support at all yet by default it should be `n d`.

robdodson:
OK, just to double check...

The feature is currently only available behind a flag in Chrome, and there is also a polyfill available. So that would be `n d`?

Fyrd:
Ha, in that case technically you do want your original `p d` though
currently polyfill'd support isn't reflected in the UI so it'll appear the
same as `n d`.

robdodson:
ok, switched back to `p d` :D

jake-low:
Sorry for causing this confusion 😬

Fyrd:
@jake-low It was a reasonable question, that one often stumps people!

PR #4161, opened 2018-03-30
Safari 11.1 does NOT support input type color

brian-wilson:
Tested with release version of Safari 11.1 against:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color


slokhorst:
Confirmed in 11.1.

What about 11.0? Was it working there?

Fyrd:
Confirmed, thanks! Also indeed it doesn't seem to be supported in 11.0, will fix that.

BenjaminHoegh:
As far I know it's only work in the Technology Preview version of Safari. Else look up in the [Safari release notes](https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html#//apple_ref/doc/uid/TP40014305-CH14-SW1)

daytonlowell:
It doesn't work for me in STP 63 (Tracking Safari 12.1) or Safari 12 beta(build 13606.1.36.1.2). Did it get removed?

chicoxyzzy:
yep it doesn't work in SF12/SFTP as well

PR #4165, opened 2018-04-05
Fix incorrect data on let in IE11

JLRishe:
The `let` data indicates that `let` is fully supported in IE11, but `let` variables are not correctly bound in `for` loops. This seems similar to Note #4, but the variables are in fact scoped to the `for` block and not the function so it seems like a separate issue.

This behavior can be observed with the following:

```js
for (let i = 0; i < 10; i += 1) {
    setTimeout(function () { console.log(i); }, 200);
}
```

In Chrome, this correctly logs the numbers 0-9. In IE, this logs the number 10, 10 times.

### DIFF
--- a/features-json/let.json
+++ b/features-json/let.json
@@ -27,7 +27,7 @@
       "8":"n",
       "9":"n",
       "10":"n",
-      "11":"y"
+      "11":"a #5"
     },
     "edge":{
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #4166, opened 2018-04-05
Position Sticky not supported in Edge 16 on <thead> or <tr>

styfle:
Edge 16 suffers from the same problem as Chrome.
Sticky position works on `th` but not `thead`.

I was able to reproduce by using the [HTML/CSS](https://jsfiddle.net/hr5dktah/128/) from the Chromium Bug.

```css
div {
  height: 200px;
  width: 200px;
  overflow: auto;
}

table {
  width: 100%;
  border-collapse: collapse;
}

thead {
  position: sticky;
  top: 0;
  background: #EEE;
}
```

The workaround is the same as Chrome, change `thead` to `th` in your css.


styfle:
The workaround is the same as Chrome, change `thead` to `th` in your css.

Fyrd:
Thanks!

mkurz:
@styfle Can you confirm in Edge 17 this is fixed and working?

styfle:
@mkurz I cannot. Edge 17 requires running a beta version of Windows 🤢

However, I did report [the bug](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/16765952/) to Microsoft so we'll see if they find it in newer versions and I can send another PR if Edge 17 has the same bug.

/cc @MicrosoftEdge @jdalton @aarongustafson @kbabbitt @melanierichards

jdalton:
The current Edge 17 preview still repros the issue.

mkurz:
That sucks!

styfle:
I created a follow up PR in #4169 so we note that Edge 17 has the same bug.

PR #4168, opened 2018-04-06
Payment Request API to Candidate Recommendation

ctolkien:
As of 20 March 2018, this is now CR.
Ref: www.w3.org/TR/payment-request/


Fyrd:
Thanks!

PR #4169, opened 2018-04-06
Position Sticky fails in Edge on <thead> or <tr>

styfle:
This is a follow up to #4166

We have had Microsoft confirm its a bug on Edge 16 and Edge 17.

I also added linked to the bug.


Fyrd:
Thanks!

PR #4001, opened 2018-01-12
Edge 17 supports relList on links…

dstorey:
…on other interfaces to come.

PR #4171, opened 2018-04-06
Edge 17 does not support Ogg Vorbis/Theora

Herst:
According to #4001 Microsoft Edge 17 would natively support the formats Ogg Vorbis/Theora but this is not the case (tested using <http://openspim.org/File:10_Sample_holder.ogv> and <https://hpr.dogphilosophy.net/test/>). The following extension (which already worked before EdgeHTML 17) would be required for support: https://www.microsoft.com/store/productId/9N5TDP8VCMHS


Herst:
The Web Media Extensions Package package got preinstalled with preview builds, this could have caused this confusion: https://developer.microsoft.com/en-us/microsoft-edge/platform/changelog/desktop/17063/

dstorey:
My understanding is the media pack will be distributed by default and enabled in Edge 17 like it was with preview builds. I can check if this has changed but it my Edge 17 it is enabled

Fyrd:
Will assume what @dstorey says is the case, and close this PR. Feel free to re-open if anything changes in that regard.

Herst:
I stand corrected, I now checked with a vanilla install and indeed the extension pack is preinstalled. TBH, I find this approach a bit confusing but I guess that this approach makes sense in some other regards (legal, easy to disable should patent problems appear, …).
Would this be worth adding a note for like in the Firefox with H.264 decoding on Linux case?

dstorey:
In reality it makes no difference. It is just an implementation detail.

Issue #3801, opened 2017-10-24
navigator.language and navigator.languages


smalllong:
https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

smalllong:
sorry, it's duplicated.

OndrejSpanel:
Is the duplicated request still somewhere? I have searched for it but I wound only low quality #3423 

I would also like to know about `navigator.languages` and `navigator.language` browser support.

smalllong:
@OndrejSpanel  Duplicate of #3214

PR #4181, opened 2018-04-09
Add feature: NavigatorLanguage JS APIs

jake-low:
HTML 5.1 (rec) specifies three language detection related features supported in many (but not all) modern browsers:

- `navigator.languages` which provides an array of [BCP47](https://tools.ietf.org/html/bcp47)-style language codes representing the user's (explicit or inferred) preferred languages in most-preferred-first order.
- `navigator.language` which is identical to `navigator.languages[0]`, mimicking nonstandard behavior found in many older browsers.
- a `languagechange` event which fires on `window` when the user makes changes to their preferred language settings.

This PR adds data for this feature. Addresses #3801 and #4144.

Support notes:
- support for only `navigator.language` (but not `languages` or the `languagechange` event) is treated as "no support" rather than "partial support", since many old browsers have non-standard behavior for this feature (e.g. returning language codes in other formats), which makes it impossible to rely on when the result of `navigator.language` is going to be used as an argument to an i18n function.
- Samsung Internet support is inferred from Chromium support, based on [this document](https://gist.github.com/poshaughnessy/5718717a04db20a02e9fdb3fc16e2258) by @poshaughnessy
- I was unable to test UC 11.8 (so I left it as `"u"`), but UC 11.3 has no support.
- I was also unable to test or find info for BlackBerry, Opera Mini, QQ, or Baidu mobile browsers.

Here's a pen I used to test support: https://codepen.io/jakelow/pen/geqpzE/


Fyrd:
Thanks!

PR #4182, opened 2018-04-10
overflow-wrap is in Edge 18

dstorey:


xehpuk:
Source?

dstorey:
Edge is closed source

xehpuk:
From [Wiktionary](https://en.wiktionary.org/wiki/source):
> 1. The person, place, or thing from which something (information, goods, etc.) comes or is acquired.

Issue #4013, opened 2018-01-16
SameSite data is not accurate for Firefox


alex:
https://caniuse.com/#feat=same-site-cookie-attribute

indicates that Firefox supports SameSite cookies starting with Firefox 58. This is not correct. Work on this is tracked here: https://bugzilla.mozilla.org/show_bug.cgi?id=795346 and as you can see the bug is still open.

Fyrd:
Fixed!

silverlyra:
caniuse is currently showing that this landed in Firefox 59, but this is also not correct. [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1286861) indicates that it's landing in 60.

PR #4195, opened 2018-04-13
Fix SameSite cookie support for Firefox 59

silverlyra:
Firefox 59 does not support this yet. The relevant changes to Firefox's core [just landed][bug1286821] and won't ship until Firefox 60.

re: #4013

[bug1286821]: https://bugzilla.mozilla.org/show_bug.cgi?id=1286861


Fyrd:
Thanks!

PR #4201, opened 2018-04-18
downlinkMax instead of downloadMax

gmetais:
Hi there!

It looks like this is a typo.

https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlinkMax


Fyrd:
Thanks!

PR #4208, opened 2018-04-19
update opera mini support for css variables

kaycebasques:
Howdy,

Just tested out my personal site (which uses CSS Variables) on Opera Mini for iOS. Seems to be working!


Fyrd:
It works in the default version on iOS and Android, but not the proxy version which is used by the majority of Opera Mini users on feature phones, which is what is tracked on caniuse.

kaycebasques:
Got it, thanks. Consider explaining that in the Caniuse UI?

kaycebasques:
In retrospect it's obvious that Opera Mini must be using a similar arrangement as Chrome for iOS (just a wrapper on top of the Apple-sanctioned browser stack), but it didn't cross my mind at the time!

PR #4210, opened 2018-04-19
Only partial support for IE and "Edge"

jabcreations:
Only partial support for IE and "Edge" due to no support for .x and .y properties.

### DIFF
--- a/features-json/getboundingclientrect.json
+++ b/features-json/getboundingclientrect.json
@@ -32,19 +32,19 @@
       "5.5":"a #1 #3",
       "6":"a #1 #3",
       "7":"a #1 #3",
-      "8":"a #1 #3",
-      "9":"y",
-      "10":"y",
-      "11":"y"
+      "8":"a #1 #3 #5",
+      "9":"a #5",
+      "10":"a #5",
+      "11":"a #5"
     },
     "edge":{
-      "12":"y",
-      "13":"y",
-      "14":"y",
-      "15":"y",
-      "16":"y",
-      "17":"y",
-      "18":"y"
+      "12":"a #5",
+      "13":"a #5",
+      "14":"a #5",
+      "15":"a #5",
+      "16":"a #5",
+      "17":"a #5",
+      "18":"a #5"
     },
     "firefox":{
       "2":"n",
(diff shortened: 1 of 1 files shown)


dstorey:
This is more partial support for DOMRect than getBoundingClientRect() (which is basically the same issue as https://caniuse.com/#feat=dommatrix ... all the CSS/SVG interfaces were changed to shared DOM version)

jabcreations:
@dstorey If you delete your comment I'll delete mine. You work on the IE team and clearly what you proposed is not only biased though an outright lie. If Gecko doesn't support something on something else than mark it as so in a separate bug and/or commit.

dstorey:
(actually Edge team, not IE). I'm not sure I follow? Edge doesn't support x and y IDL attributes of the returned object. Firefox doesn't support toJSON and fromRect methods of the same. Either both are partial or neither, but I think this difference stems from the renaming of clientRect/SVGRect to DOMRect along with SVGMatrix and WebkitCSSMatrix to DOMMatrix, SVGPoint to DOMPoint, et al. The partial implementation is in renaming and merging those interfaces (SVGRect has x and y already)

jabcreations:
1. Edge is simply a rebranding of IE therefore it's still IE. There is nothing wrong with IE being IE. When Hitachi had an 18% failure rate of hard drives out the door what did they do? They fixed the problem and now have the highest reliability in the industry according to one business's reports. The rebranding actually hurt IE. So while the GUI was destroyed in IE7 the rendering engine has only ever continued to improve.

2. I'm far from a Mozilla supporter. In fact I'm about as anti-Mozilla as one can get after their daily cocaine parties ended with the destruction of customization and the cloning of Chrome's crappy GUI. The only customizable / usable browser left is Waterfox. In fact while I also generally do not favor Microsoft I am always willing to admit when anyone does a great job and how they handled the rendering of border-style in combination with border-radius years ago is one example of when Microsoft commits to doing something right that they come through.

3. If Gecko (or any engine) has impartial support for something then mark it as so. That is the whole point of caniuse: can I use it? Yes/Partially/No? Both are partial cases. I've never even heard of that method though it's clearly unrelated. I also hardly use SVG (just inline CSS for icons for when there is no appropriate Unicode glyph available for an immediate need).

So please post something independently about the issue with Gecko having partial support. Caniuse is for developers including myself a tool to determine which API is the right tool for the job; it is not to showcase which browser is more or less capable.

jabcreations:
...and Microsoft astroturfing my replies!

Fyrd:
I've spent some time thinking about this one, since the API itself works and the lack of x & y values is easily fixed using left & top I don't think it needs to qualify as partial support, so I've added a note for it instead. If enough people feel strongly enough about it that may change.

PR #4233, opened 2018-05-02
Preload is partially supported in Edge 17+

nolanlawson:
Similar to preconnect (https://github.com/Fyrd/caniuse/pull/3829), Edge 17 only has partial support for this feature.


Fyrd:
Thanks!

PR #4235, opened 2018-05-05
CCS Masks on Edge 17 are partially supported behind a flag

HM100:
I found a flag in Edge 17 that enables CCS masks, Proof: https://aka.ms/devguide_edgehtml_17


Fyrd:
Thanks!

PR #4236, opened 2018-05-05
Edge 17+ spports CSS transsforms on SVG elements

HM100:
Proof: https://docs.microsoft.com/en-us/microsoft-edge/dev-guide


Fyrd:
Thanks!

PR #4245, opened 2018-05-10
Edge Does not support @apply

HM100:
You have to use Edge 15+ and go to http://codepen.io/zeke/pen/XdVopX a and you'll observe that styles made from @apply from a variable that has a CSS code in it won't get parsed. Tried today with Edge 17 but also affects Edge 15-16 and 18 as well


Fyrd:
Thanks!

PR #4250, opened 2018-05-13
Chrome 65+ Supports :matches behind a flag

HM100:
https://www.chromestatus.com/feature/5445716612743168 is the proof


Fyrd:
This should be `d` rather than `f`. It should also have a separate note number explaining what it refers to.

HM100:
@Fyrd Fixed

Fyrd:
Nice, thanks!

PR #4255, opened 2018-05-15
Edge 17 supports it fully, Fixing note 2 to avoid people thinking Edg…

HM100:
…e 17 also supports it differently


Fyrd:
Thanks!

PR #4256, opened 2018-05-16
update(es6-module): update Opera data

CITguy:
According to https://www.chromestatus.com/feature/5365692190687232 Opera has had ES6 module support since v48 (which parallels Chrome support).


Fyrd:
Thanks!

PR #4257, opened 2018-05-16
Removed incorrect platform status mapping

phistuck:
The mapped platform status entries talk about v1, not v0.
Hopefully, it will also fix the lack of WebKit status on the v1 entry...
image

### DIFF
--- a/features-json/shadowdom.json
+++ b/features-json/shadowdom.json
@@ -332,9 +332,6 @@
   "ucprefix":false,
   "parent":"",
   "keywords":"web components",
-  "ie_id":"shadowdom",
   "chrome_id":"4507242028072960",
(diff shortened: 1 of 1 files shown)


Fyrd:
These should just be empty strings, not removed entirely.

phistuck:
You got it.

Fyrd:
Thanks!

Fyrd:
@phistuck FYI the reason Safari's status doesn't show up is because it's listed as "Supported". When that's the value it's not shown since in theory the support table should already reflect that. You can see it in the Resources tab though.

phistuck:
The entry above it (v0) did show "Supported" for WebKit. Is that because the table did not show it as supported?

PR #4266, opened 2018-05-21
Correct note reference for Chrome 

mattbrundage:
From v. 65 onward, the note "_Only_ supports the `:-webkit-any()` pseudo-class ..." should be "_Also_ supports the `:-webkit-any()` pseudo-class ..." , as that's when support for ':matches()' became available behind the "Experimental Web Platform features" flag.


Fyrd:
Thanks!

PR #4270, opened 2018-05-22
xhr2.json: update stale w3c link

bmalehorn:
If you follow thie original link, you get a warning:

    This version is outdated!
    For the latest version, please look at
    https://www.w3.org/TR/XMLHttpRequest/.

Replace the old link with the new one now that this is a standard.


Fyrd:
The link there is the intended one, as it's part of the "Previously known as XMLHttpRequest Level 2" sentence, so it should show the old specification. The (whatwg) link to the current specification is already included. (see the "spec" entry)

bmalehorn:
Ah, makes sense. Thanks for the quick reply.

PR #4277, opened 2018-05-26
Update es6-module-dynamic-import.json

yukulele:
chrome 63+ support import()


Fyrd:
You'll wanna leave that "66" alone, it's the latest version, not "version in which support started"

yukulele:
so we can't have more precision on chrome mobile support?

Fyrd:
Thanks! Not yet, but I'm working on something that'll allow it.

PR #4295, opened 2018-06-08
Add css-block-overflow.json

jserrao:
Looks like a promising new feature on the horizon.


Fyrd:
Caniuse doesn't generally include features that are still in early phases of specification, unless browsers developers have indicated they're working on supporting it. Do you know if any browser's working on this?

jserrao:
Yeah I hear you, looks pretty early yet. I'll keep an eye if it pops up.

PR #4298, opened 2018-06-11
Fix data for IE & Edge

morille:
Blob URL as iframe "src" does work in Edge since v15 and did never work in IE.
Can try this pen: https://codepen.io/grimen/pen/lBuiG
The issue for Edge concerns PDF blob urls only (https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/5411066/)


Fyrd:
Thanks!

PR #4303, opened 2018-06-13
Update spec for box-sizing

frivoal:
The most up to date spec for box-sizing is no longer css-ui, but css-sizing.


Fyrd:
Thanks!

PR #4307, opened 2018-06-14
Add Edge feature request to AV1

Schweinepriester:
Title :)

https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/34544281-support-aomedia-video-1-av1

Meta: That feature request is probably unnecessary, as Edge is already listed at https://aomedia.org and Microsoft is a member, but since they don't have it listed at https://developer.microsoft.com/en-us/microsoft-edge/platform/status/ maybe at least the request will lead to that…

### DIFF
--- a/features-json/av1.json
+++ b/features-json/av1.json
@@ -11,6 +11,10 @@
     {
       "url":"https://bitmovin.com/demos/av1",
       "title":"Sample video"
+    },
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #4313, opened 2018-06-16
Webviews don't support getUserMedia

tomasdev:
Facebook, Instagram, Twitter, Snapchat they all have the same problem.


Fyrd:
Confirmed, thanks!

PR #4317, opened 2018-06-19
safari partial gradient support (fixes #4243)

### LINKED ISSUE #4243: transparent keyword in gradient badly supported in iOS/Safari
According to the w3c, gradients should use premultiplied colors...
https://www.w3.org/TR/2012/CR-css3-images-20120417/#color-stop-syntax
Basically this: https://jsfiddle.net/hqcgq1dm/ should be a nice transition from white to turquoise. In iOS and Safari, it isn't because they do a transition from `rgba(255,255,255,1)` to `rgba(0,0,0,0)`.
This postcss plugin which addresses this issue explains it a little better:
https://github.com/gilmoreorless/postcss-gradient-transparency-fix

Anyways.... shouldn't this make the support of gradients in safari and iOS "partial"?

Nemo64:
I tested this using: https://jsfiddle.net/hqcgq1dm/
Here is a potential polyfill https://github.com/gilmoreorless/postcss-gradient-transparency-fix

This is my first time adding something to this repo.
Is this here really that much manual labor? I would have expected some kind of automatic testing environment.


Nemo64:
I did change the text. However, I'm not sure if i can add links to the text.

And I'm pretty sure that opera's old rendering engine had the same problem but i'm not sure how to easily test that.

Nemo64:
done

HM100:
Thanks

Fyrd:
Thanks!

PR #4322, opened 2018-06-20
Adjust notes about Edge <template> support for version 15+

tuespetre:
I have tested version 15 through Browserstack, and versions 16/17 on my own machine and did not find the issues to be valid for these versions.

Here is the document I used to test:

```
<!DOCTYPE html>

<title>&lt;template&gt; testing</title>

<div id="wrapper">
    <template id="outer">lol<template id="inner">wut</template></template>
</div>

<script>
    var wrapper = document.getElementById('wrapper').cloneNode(true);
    var outer = wrapper.querySelector('#outer');
    var clone = outer.cloneNode(true);
    var inner = document.importNode(clone.content.lastChild, true);
    var content = document.importNode(inner.content, true);
    document.body.appendChild(content);
</script>
```


Fyrd:
Thanks!

PR #4324, opened 2018-06-20
Fix some Edge values

FremyCompany:
I also planned to change `-webkit-line-clamp` too but apparently someone already reported that change before.


features-json/webp.json
```diff
@@ -48,7 +48,7 @@
       "15":"n",
       "16":"n",
       "17":"n",
-      "18":"n"
+      "18":"d"
```
HM100: Change **d** to **n d** so as to be correct

HM100: @FremyCompany Any way of fixing that ones I gave? It is needed to be done by now.

FremyCompany:
@HM100 Change done!

HM100:
Thanks!

Fyrd:
Thanks! One change required is that the support for `text-size-adjust` should be prefixed, but I'll take care of that.

PR #4326, opened 2018-06-21
Web Animation is on by default in Safari TP 59

chicoxyzzy:
https://webkit.org/blog/8343/web-animations-in-webkit/


Fyrd:
Thanks!

PR #4328, opened 2018-06-22
Chrome 69 will add Conic Gradients without any flag according to Chro…

HM100:
…mestatus.com

https://www.chromestatus.com/feature/5706155347148800


Fyrd:
Thanks!

PR #4338, opened 2018-06-27
FF <19 actually support the initial value

HM100:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/initial), initial in FF1-18 require -moz- prefix for initial value while -moz-prefixed version removed in FF24. V2 to V3 will be maked as partial because it lacks supports on `quotes` property as what [this article](https://developer.mozilla.org/en-US/docs/Web/CSS/quotes) says.


Fyrd:
Thanks!

PR #4346, opened 2018-06-28
AV1 - Added demo video from facebook which works in chrome

pwed:
Edited bitmovin video link to show that it only works on Firefox Nightly currently


Fyrd:
Thanks!

PR #4347, opened 2018-06-29
IE and Edge should be partial due to a bug found

HM100:
Just found a [bug](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10792541) that Edge and IE still allow alphabetic characters on number input types where it shouldn't.

### DIFF
--- a/features-json/input-number.json
+++ b/features-json/input-number.json
@@ -55,17 +55,17 @@
       "7":"n",
       "8":"n",
       "9":"n",
-      "10":"y #1",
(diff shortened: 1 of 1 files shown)


Fyrd:
I don't think this is a serious enough problem to downgrade support to partial, an entry under "bugs" should be sufficient. I also just checked and Firefox & Safari also allow alphabetic characters to be entered, and Chrome allows any amounts of the letter "e"

PR #4351, opened 2018-07-02
Added notes on high-resolution timestamps

lightswitch05:
* High-resolution timestamps can be rounded
* Firefox has multiple settings that have a major affect on High-resolution timestamps


Fyrd:
Thanks!

PR #4356, opened 2018-07-03
Update web-app-manifest status for firefox

armujahid:
Web Application Manifest support was landed in firefox 58
as per https://platform-status.mozilla.org/#app-manifest


mgiuca:
See #4357 --- does Firefox support appinstalled and beforeinstallprompt events?

I'd like to make it so not supporting these events earns you an "a"; you need to implement them to get a "y" (currently this is inconsistent).

armujahid:
Ok I will update it to `a #2` `The installation events, appinstalled and beforeinstallprompt, are not supported on all platforms.`

Fyrd:
I don't believe Web Application Manifest is supported in desktop versions of Firefox, is it? For Android it's already covered by the`and_ff` entry.

As for the events, based on [this comment](https://github.com/Fyrd/caniuse/pull/4357#issuecomment-402533604) it seems it may not be appropriate to consider the support partial.

mgiuca:
I'd rather mark all these as partial (except Chrome Android and the others that support BIP). Either that, or create a separate caniuse entry for BeforeInstallPrompt event, since it's something that developers are likely to want to know if they can use on various browsers.

Fyrd:
Closing this issue until there's evidence of it being supported in Firefox outside of Android.

PR #4357, opened 2018-07-04
web-app-manifest: Updated Chrome and Edge support.

mgiuca:
Chrome was listed as partial support; clarified in the footnote that this is due to missing installation events (starting with 67, fully supported on Chrome OS but not other platforms).

Edge was listed as full support, but it is also missing the installation events, so downgraded to partial support to match Chrome.

Test case: https://killer-marmot.appspot.com/web/ (logs installation events in the page).

### DIFF
--- a/features-json/web-app-manifest.json
+++ b/features-json/web-app-manifest.json
@@ -31,8 +31,8 @@
       "14":"n",
       "15":"n",
       "16":"n",
-      "17":"y",
-      "18":"y"
+      "17":"a #1",
+      "18":"a #1"
     },
     "firefox":{
       "2":"n",
@@ -163,10 +163,10 @@
       "64":"a #1",
       "65":"a #1",
       "66":"a #1",
-      "67":"a #1",
-      "68":"a #1",
-      "69":"a #1",
-      "70":"a #1"
+      "67":"a #2",
+      "68":"a #2",
+      "69":"a #2",
+      "70":"a #2"
     },
     "safari":{
       "3.1":"n",
@@ -312,7 +312,8 @@
   },
   "notes":"",
   "notes_by_num":{
-    "1":"On Windows, Chrome will read the manifest to generate a desktop icon when selecting the \"Add to desktop...\" option."
+    "1":"The installation events, appinstalled and beforeinstallprompt, are not supported.",
+    "2":"The installation events, appinstalled and beforeinstallprompt, are not supported on all platforms."
   },
   "usage_perc_y":47.8,
   "usage_perc_a":28.37,


mgiuca:
@marcoscaceres @boyofgreen @dominickn FYI

marcoscaceres:
Thanks for the update. Given that neither Safari, Firefox, or Edge are supporting BIP, we should probably make BIP optional in the spec (if we haven't already).

mgiuca:
Hmm, wouldn't that discourage them from implementing? We should try and get commitment from another vendor to implement BIP or at least appinstalled.

I'm happy either way (for caniuse) to make browsers that don't support BIP/appinstalled either "y" or "a", but we should be consistent.

marcoscaceres:
Let’s pick this up over email or in our repo :)

mgiuca:
Done: See w3c/manifest#699.

Fyrd:
Have updated this one manually.

mgiuca:
@Fyrd The updated version is still inconsistent as to whether you get full or partial support based on the implementation of these APIs.

None of Edge, Safari and Firefox support appinstalled or beforeinstallprompt (hence they have the footnote 2), but they are still given "full support". Chrome supports these APIs on all platforms other than macOS, yet it is classified as "partial support". Either all the browsers should be listed as "partial support", or more realistically, Chrome should be upgraded to full support.

(In fact Edge technically doesn't support Web App Manifest at all, since there's no way to install a site from Edge, you must do it from the Windows Store which happens to use the Edge rendering engine as a backend.)

Also, you should be able to remove Footnote 1. I don't really know what "Chrome will read the manifest to generate a desktop icon when selecting the "Add to desktop..." option." means. Isn't this just "it's supported"?

marcoscaceres:
@mgiuca we (Editors/implementers) are probably in the best positions to answer these questions based on testing and our own investigations. You and I should probably just sit down and put together a quick PR to make sure the current state of things is adequately reflected.

mgiuca:
@marcoscaceres I did put together a PR... this one. It was closed and a different set of changes were applied instead.

Granted, my PR resolved the inconsistency by setting all browsers to "partial". Perhaps it's more straightforward to change them all to "full" and keep the footnotes about the event support. Happy to grab some time with you tomorrow.

marcoscaceres:
> Granted, my PR resolved the inconsistency by setting all browsers to "partial". Perhaps it's more straightforward to change them all to "full" and keep the footnotes about the event support. Happy to grab some time with you tomorrow.

I don't think that's correct tho. For Payment Request, for instance, we it clear that we only have partial support because people don't implement a particular feature (e.g., `retry()`):
https://w3c.github.io/payment-request/

mgiuca:
Updated and re-opened as #4617.

PR #4360, opened 2018-07-04
Chrome intends to ship Scroll-Snap points fully in M69

HM100:
https://www.chromestatus.com/feature/5721832506261504


Fyrd:
Thanks!

PR #4364, opened 2018-07-05
removed polyfill

ColinEberhardt:
The linked polyfill hasn't been updated for 3 years and the repo has been 'locked'


Fyrd:
Is there a better polyfill out there to include instead? If not, and this one still works, I don't see why it needs to be removed.

Fyrd:
Closing for now.

PR #4365, opened 2018-07-06
Updated BackgroundSync and ES6 Modules for Android webview

bjornarvh:
BackgroundSync is not available in Android webview based on [this comment](https://groups.google.com/a/chromium.org/d/msg/chromium-dev/zVO57FQFydg/SzYh5fmkBQAJ) and my own tests.

ES6 modules seems to be working (and dynamic import is already marked as supported).

Also noticed dynamic import disappeared from the site shortly after it was added. The link in news doesn't work. Seems shown was never set to true.


Fyrd:
Thanks!

PR #4367, opened 2018-07-09
Edge 13 has a bug to a[download]

HM100:
See https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7160092/


Fyrd:
Thanks!

PR #4368, opened 2018-07-09
Chrome M69 implements the new edition of Logical Properties

HM100:
https://www.chromestatus.com/feature/6041507983261696


Fyrd:
Thanks!

PR #4369, opened 2018-07-10
[CSS text-rendering] Update unoff spec link to CR

Malvoz:
There's also a SVG2 draft spec on a new location (https://svgwg.org/svg2-draft/), I don't know if the CR will move to that location in the future... I would assume it is not preferred over the CR though.


Malvoz:
Hey.. I'm new to this 😉

Fyrd:
It's a bit confusing, but this feature is actually about using the property in HTML for the purpose of enabling high-quality kerning and ligatures, not the behavior in SVG. As the behavior in HTML is unspecified (see the note at https://developer.mozilla.org/en-US/docs/Web/CSS/text-rendering), the status is considered "unofficial".

Malvoz:
You're right, how silly of me not to realize that. I should pay more attention, I roam alot of specs. 

Anyhow, the updated link in this PR points to https://www.w3.org/TR/SVG2/ which is the successor of  https://www.w3.org/TR/SVG11/

Fyrd:
I've just updated the link as well as the description, it should be a bit more clear now what this feature is about and the status of the property. Thanks!

PR #4373, opened 2018-07-10
WebAuthn not enabled in Chrome for Android 67

grzuy:
https://www.chromestatus.com/features/5669923372138496 states it is only
enabled for desktop


Fyrd:
Thanks!

PR #4374, opened 2018-07-11
WOFF 2.0 - Remove note for Safari 12+

Schweinepriester:
Removing the note #​1

> Supported only on Safari for Mac OS Sierra, not El Capitan & older.

for Safari 12 and TP, since neither is available for OS X 10.11 El Capitan & older:

- <https://developer.apple.com/safari/whats-new/>

> Safari 12 ships with iOS 12 and macOS 10.14. It is also available for macOS 10.13.6 and 10.12.6.

- <https://developer.apple.com/safari/download/>


Fyrd:
Thanks!

PR #4375, opened 2018-07-12
Update web-animation.json

Yaffle:
Firefox has only "Element#animate" enabled (with Animation#play, ...);

According to the bug https://bugzilla.mozilla.org/show_bug.cgi?id=1471814 - they are also shipping `composite: 'add'` on the keyframes by mistake.

I dont' know if they plan to enable other things at some point.


Fyrd:
Thanks!

PR #4377, opened 2018-07-12
Old Firefox Version also set as partial the Gradient Support (Matches…

HM100:
… note 2)

See [this page](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient) which says until FF35, it had the same bug Safari has


Fyrd:
Thanks!

PR #4390, opened 2018-07-22
https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/dev-guides/dn323733%28v%3dvs.85%29

HM100:
See this that IE11 on Win7 do not support this


Fyrd:
Thanks!

PR #4397, opened 2018-07-25
Edge has no support for RTCDataChannel

heyheyjc:
Edge's support for WebRTC is only partial: it does not support data channels, and there is no similar functionality provided by ORTC. Data-only transfer p2p with Edge is still NOT possible using WebRTC. Reference: https://developer.microsoft.com/en-us/microsoft-edge/platform/catalog/?q=specName%3Awebrtc


Fyrd:
Thanks!

PR #4406, opened 2018-07-30
Media Source Extensions still aren't supported by Samsung Internet

niklaskorz:
Media Source Extensions are marked as supported for Samsung Internet 5, 6.2 and 7.2.
A test on both Samsung and non-Samsung devices running the Samsung Internet Browser shows that unfortunately this still isn't the case.

screenshot_20180728-180947
screenshot_20180728-181042


niklaskorz:
Sorry for the size of the screenshots, will replace them with smaller versions if desired.

HM100:
Can you also do that on MDN? On pages like on [this one](https://developer.mozilla.org/en-US/docs/Web/API/MediaSource), it shows MSE data of SIB as supported browser
`SBI = Samsung Internet Browsers`

PR #4409, opened 2018-07-31
Add passwordrules attribute

LarsKumbier:
Safari 12 includes a new attribute `passwordrules` to describe the requirements for the passwords. The first usage is in iOS 8's Password Manager Autofill, which will now generate harder passwords according to the specified rules.

I've searched all other browser's feature lists, but could not find any references to a planned implementation. Safari 12 is [confirmed](https://webkit.org/blog/8327/safari-technology-preview-58-with-safari-12-features-is-now-available/), though. Therefore, I've switched all versions beside the newest to "no" due the standard appearing first in March.

A demo page can be found here: https://developer.apple.com/password-rules/

The standard is described at [this WhatWG issue](https://github.com/whatwg/html/issues/3518), and although it is not accepted yet, one major browser implementing it should be enough to include it in caniuse.


Fyrd:
Thanks!

PR #4421, opened 2018-08-06
Correct TLS support for Opera

bazzadp:
Opera does not yet support TLS1.3 by default.
41.0.2323.0 possible with command line switch https://forums.opera.com/topic/16519/new-opera-developer-41/13
54.0.2920.0 has added a flag: https://blogs.opera.com/desktop/changelog-for-54/


Fyrd:
Thanks!

PR #4422, opened 2018-08-07
Added note about firefox bug

Qwertie-:
CORS is broken for resources which redirect. I don't know every version affected so I added it to my version and all newer versions as the bug is still open.


Fyrd:
Thanks!

PR #4423, opened 2018-08-07
iOS 12 supports indeterminate checkboxes

mkurz:
https://bugs.webkit.org/show_bug.cgi?id=160484 was marked as fixed a couple of hours ago.

### DIFF
--- a/features-json/indeterminate-checkbox.json
+++ b/features-json/indeterminate-checkbox.json
@@ -10,7 +10,7 @@
     },
     {
       "url":"https://bugs.webkit.org/show_bug.cgi?id=160484",
-      "title":"WebKit Bug 160484 - iOS doesn't support indeterminate checkboxes"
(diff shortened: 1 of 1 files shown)


HM100:
Note that you can make custom indeterminate checkboxes too. Tested and works on iOS 11.3-11.4 too but native ones is missing in 11.3-11.4- versions of iOS safari

Schweinepriester:
Is this already in iOS 12 (beta) or are we sure that this is going to be in iOS 12?

In any case: good news :)

mkurz:
@Schweinepriester I don't actually know, it was just a guess that this will make it in iOS 12. We can wait for the next beta and see...

Fyrd:
Thanks!

HM100:
It does not appear to work for me on iOS12. Has it got removed too.

PR #4425, opened 2018-08-08
Update css-text-orientation.json

HM100:
According to [FX41 Chancelog](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/41), text orientation is supported since 38 via flag and since 41 without flag, not since 48.


Fyrd:
Confirmed, thanks!

HM100:
@Fyrd found that data about tetxt-orientstion has been messed up Fx support again. Could you backport my changes done in that PR I did again? Because support of Fx has been changed from v41 (v38 under flag) into v51. Thanks

PR #4428, opened 2018-08-10
Update font-loading.json

harking:
Supported in Chrome Android since 35


HM100:
Hey. You can't add new browser versions with PRs and for android browser beyond 4.4.4, only the latest chromium version will be shown so this PR is useless and @Fyrd please close the PR.

PR #4429, opened 2018-08-10
Fixes for Text-Orientation

HM100:
My Past PR edited chrome data, not fx ones


Fyrd:
Ah yes, should have caught that. Thanks!

PR #4431, opened 2018-08-11
Update known issues of HTML5 audio

marto55555:
* Changing playbackRate is now fixed on Chrome for Android: https://bugs.chromium.org/p/chromium/issues/detail?id=263654
I manually tested it on my phone (Android 8.1, Chrome 68) and it works. Here is a quick demo: https://html5-audio-playbackrate.glitch.me/

* Include that iOS also does not support autoplay as advised by the spec
* Include that iOS does not fire the `ended` event when the screen is off or the browser is in background: https://bugs.webkit.org/show_bug.cgi?id=173332

### DIFF
--- a/features-json/audio.json
+++ b/features-json/audio.json
@@ -42,10 +42,10 @@
       "description":"Volume is read-only on iOS."
     },
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #4434, opened 2018-08-14
Web Notifications work just fine on CFA

rayfoss:
Chrome for Android 67 does support Web Notifications, it works great! Just have to use service workers as usual. https://sagiegurari.github.io/simple-web-notification/

screenshot_1534265211


Fyrd:
Confirmed. However since the feature works without service workers in Chrome for desktop and Firefox for Android, there should be a note pointing out that support in CFA requires service workers.

Fyrd:
Thanks!

PR #3855, opened 2017-11-14
refactor: Update browser support

LuisCaceres:
Chrome 61 for both desktop and Android support this feature without a flag. This file has been modified in order to reflect this update. For more information visit Chrome Platform Status.

PR #4444, opened 2018-08-17
CSSOM Scroll-behavior supported by Android WebView, too.

Schweinepriester:
#3855 didnt change this for Android WebView, probably because 61 wasnt in the data yet, but according to https://www.chromestatus.com/feature/5812155903377408 its supported. Didnt test though.

Also not sure about note 1, but since #3855 didnt add it to Chrome for Android…


Fyrd:
Thanks! Not sure about the note either, will need to investigate that one.

PR #4449, opened 2018-08-21
Update input-color.json

daytonlowell:
It doesn't work for me in STP 63 (Tracking Safari 12.1) or Safari 12 beta(build 13606.1.36.1.2) on High Sierra. Is it maybe Mojave-only?


HM100:
It should be partial then, not no support IDK what to say. Wait for official answer from @Fyrd so as to answer you since he may use STP. I don't have a mac computer, sorry

daytonlowell:
Mojave isn't released to the public yet, so it should be no.

Fyrd:
Confirmed, looks like it was in an older build of TP but not seeing it now.

HM100:
Why safari lags too much on HTML5 forms? No Datalist and No color input type now. Too bad. Edge and others have them too

PR #4456, opened 2018-08-23
Changes to WebXR

Utopiah:
Might be good to rename webvr.json to webxr.json for search but unsure of the consequences.


Fyrd:
Thanks!

PR #4457, opened 2018-08-23
Update note that Firefox on Linux will support prefers-reduced-motion

Aljullu:
Previously, the note said Firefox 63 would support `prefers-reduced-motion` but only on Windows.

According to this bug, it's implemented on Linux too:

https://bugzilla.mozilla.org/show_bug.cgi?id=1478519

I tested it and it seems to work as expected.


Fyrd:
Thanks!

PR #4469, opened 2018-08-29
CSS Logical Properties - Safari TP 64 changed something

Schweinepriester:
Maybe this pull request should have been an issue, as I'm not quite sure if `y` is truly correct. So this might be more a reminder to investigate. Feel free to close or change this.

---

<https://webkit.org/blog/8406/release-notes-for-safari-technology-preview-64/>

> Implemented flow-relative margin, padding, border and sizing properties (r234798)

⇒ https://bugs.webkit.org/show_bug.cgi?id=188386

<http://tests.caniuse.com/> seems to show the same result as Chrome Canary (Version 70.0.3535.5), but for some reason the column on the right is empty…

bildschirmfoto 2018-08-30 um 01 23 03

Also in the ticket they mention some caveats, but seem to have achieved parity with Chrome.

Thus not sure about partial or not and the note(s). But at least its unprefixed!
I've gone the optimistic route with the commit 0:-)


Fyrd:
Confirmed, thanks!
FYI the reason the column on the right is empty is because the browser-detecting code isn't currently working for TP. You'll see it if you manually select it (or any other webkit browser) from the dropdown though.

PR #4472, opened 2018-09-02
Update datalist.json

mfranzke:
According to Google Chrome Canary testing results, a provided bug fix should make this feature working correctly within Chrome 69


Fyrd:
Great, thanks!

Issue #4473, opened 2018-09-03
Various erroneous notes on innerHTML page


vlakoff:
**edit:** Because of the confusion between parseFromString for XML and for HTML, the message below is wrong. Please ignore it!

----

It seems there are various mistakes on [innerHTML](https://caniuse.com/#search=innerhtml), about the notes (partial support).

For example, concerning IE:
* IE 6-9 do support `insertAdjacentHTML` (though it's broken for tables, which is already noted in the known issues) – Refs: [insertAdjacentHTML](https://caniuse.com/#feat=insertadjacenthtml)
* IE 9 do support `parseFromString`, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser#Browser_compatibility).

So, IE 6-8 should not have note 2, but note 3, and IE 9 should have no note.

<br> I also found issues concerning Firefox:
* Firefox supports `parseFromString` since first version, but `insertAdjacentHTML` since version 8.

So, all current notes for Firefox should be removed, and a 4th note should be created, for Firefox 1-7: "Partial support refers to lacking support for insertAdjacentHTML on the DOMParser."

<br> Concerning Chrome:
* Chrome supports all features since first version.

So, notes that are currently on Chrome 1-30 should be removed.

vlakoff:
Just for the record, the Firefox ticket: [Bug 613662 - Implement insertAdjacentHTML](https://bugzilla.mozilla.org/show_bug.cgi?id=613662).

vlakoff:
OK, I have all checked again, and the data seems to be entirely correct, good work!

There is only one remaining thing:
* IE 6-9 do support insertAdjacentHTML, though it's broken for tables. Currently, you note IE 6-9 don't support insertAdjacentHTML at all.

vlakoff:
The current presence of notes emphasizes there are issues with insertAdjacentHTML (also innerHTML?) in IE < 10, even if they do exist.

So I guess it's fine as it is, instead of making it fully green thus making the issues much less noticeable.

PR #4475, opened 2018-09-04
insertAdjacentHTML is broken in IE 9 as in preceding versions

vlakoff:
Follow-up to #4473.

IE 10 did both: it fixed insertAdjacentHTML and it implemented parseFromString for HTML.


Fyrd:
Thanks!

PR #4477, opened 2018-09-05
added iOS notes about PWA and related

odahcam:
This is VERY important for those who wants to create a PWA with camera access.

Sources: 
- https://stackoverflow.com/questions/46228218/how-to-access-camera-on-ios11-home-screen-web-app
- https://github.com/webrtc/samples/issues/933

Implementation: https://minhahavan.com


odahcam:
This data was added to the wrong table. Please ignore.

PR #4478, opened 2018-09-05
Added iOS notes about PWA and related

odahcam:
This is VERY important for those who wants to create a PWA with camera access.

Sources: 
- https://stackoverflow.com/questions/46228218/how-to-access-camera-on-ios11-home-screen-web-app
- https://github.com/webrtc/samples/issues/933

Implementation: https://minhahavan.com

### DIFF
--- a/features-json/stream.json
+++ b/features-json/stream.json
@@ -277,8 +277,9 @@
       "9.3":"n",
       "10.0-10.2":"n",
       "10.3":"n",
-      "11.0-11.2":"y",
-      "11.3-11.4":"y",
+      "11.0-11.2":"y #3",
+      "11.3":"y",
+      "11.4":"y #3",
       "12":"y"
     },
     "op_mini":{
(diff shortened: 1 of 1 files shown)


features-json/stream.json
```diff
@@ -277,8 +277,9 @@
       "9.3":"n",
       "10.0-10.2":"n",
       "10.3":"n",
-      "11.0-11.2":"y",
-      "11.3-11.4":"y",
+      "11.0-11.2":"y #3",
+      "11.3":"y",
```
HM100: This one should be here. Please remove it

HM100: Please do it

features-json/stream.json
```diff
@@ -277,8 +277,9 @@
...
+      "11.4":"y #3",
```
HM100: And change 11.4 to 11.3 -11.4
Per Repository rules, no new versions can be added or be splitted into other versions. Please fix that ASAP. Thank you for the time being!!!!

HM100: Even if only affects 11.4 and not both 11.3 to 11.4, you mustn't break browser versions.

odahcam: Damn, that can get very confusing, but ok.

Fyrd:
Thanks!

odahcam:
Yw!

slokhorst:
Also doesn't seem to work in iOS 12. Can you confirm?

odahcam:
Confirmed, at least `getUserMedia` does not work inside a standalone-running PWA. 😞

PR #4482, opened 2018-09-08
Update custom-elements.json

HM100:
According to chromestatus.com, Chrome intends to deprecate CustomElements V0 at M70, following by an intension to remove them at M73. See [here](https://www.chromestatus.com/feature/4642138092470272) for more info about this.


Fyrd:
Thanks!

Issue #4095, opened 2018-02-22
::placeholder information incorrect for IE11


yourboyblue:
Notes state:

> Partial support refers to using alternate names: 
:-ms-input-placeholder for IE. 

But, as stated here: (https://github.com/necolas/normalize.css/issues/550#issuecomment-200511864), at least in IE11, :-ms-input-placeholder implements :placeholder-shown, and therefore attempting to style it actually styles the whole input, and not the ::placeholder itself.

ai:
Is it fixed already?

PR #4483, opened 2018-09-08
Fixing #4095

HM100:
According to [this issue](https://github.com/Fyrd/caniuse/issues/4095), IE implements `:-ms-input-placeholder` as `:placeholder-shown`, not as `::placeholder` so this PR fixes it.


HM100:
Everything is taken from https://developer.mozilla.org/en-US/docs/Web/CSS/:placeholder-shown

HM100:
More changes were done than the Issue requested because Mozilla Firefox 4-18 do not support `:-moz-placeholer` (Which was until V51) as `::placeholder` but as `:placeholder-shown`.

Fyrd:
Thanks!

HM100:
@mattbrundage Because pic says it all 
image
See that in IE whole input is with red bg
image
And in Edge and other browsers, only its placeholder is styled (Input still white bg)
That's why I removed reference

PR #4488, opened 2018-09-13
Update Edge CSP2 support

lweichselbaum:
Unfortunately Edge has a broken implementation of CSP nonces that can lead to severe breakages for developers using a CSP based on nonces instead of whitelists.
Since Edge ignores nonces on sourced scripts they'll get blocked and likely break the site. Developers have to resort to useragent sniffing.
Bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13246371/
Demo: https://arturjanc.com/cgi-bin/edge-nonce.py

### DIFF
--- a/features-json/contentsecuritypolicy2.json
+++ b/features-json/contentsecuritypolicy2.json
@@ -15,7 +15,7 @@
   ],
   "bugs":[
     {
-      "description":"Partial support in Edge refers to [broken nonce suport](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13246371/) which will lead to breakages since soucred script tags with a valid nonce will get blocked."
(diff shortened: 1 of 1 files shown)


features-json/contentsecuritypolicy2.json
```diff
@@ -14,7 +14,9 @@
     }
   ],
   "bugs":[
-    
+    {
+      "description":"Partial support in Edge refers to [broken nonce suport](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13246371/) which will lead to breakages since soucred script tags with a valid nonce will get blocked."
```
mikispag: Typo. "sourced".

Fyrd:
Thanks!

PR #4491, opened 2018-09-14
AudioContext.createMediaElementSource(): Remove mentioned Safari bug

vinyldarkscratch:
Safari fully supports `AudioContext.createMediaElementSource()` [since version 6](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/createMediaElementSource#Browser_compatibility).


Fyrd:
Thanks!

PR #4493, opened 2018-09-16
Update css-case-insensitive.json

AmeliaBR:
Selectors 4 was re-published as Working Draft with the case-insensitive attribute selector included in it.


Fyrd:
Thanks!

PR #4500, opened 2018-09-21
Firefox Bug 1269276: Enable unprefixed Fullscreen API

ExE-Boss:
See [bug&nbsp;1269276](https://bugzil.la/1269276) for details.


Fyrd:
Thanks!

PR #4516, opened 2018-09-28
Update (Audio|Video) Tracks Opera+Chrome+FF data

sideshowbarker:
Audio and video tracks are supported under a flag in Chrome 45+ and Opera 32+. See https://www.chromestatus.com/feature/5748496434987008

Also, bring the Firefox data for Video Tracks in sync with what we already have for Audio Tracks (= supported under a flag in FF 33+).


Fyrd:
Thanks!

PR #4521, opened 2018-10-01
Update web-share.json

MarkSmit:
Safari Technology Preview does support the Web Share API


HM100:
Can you do the same with Conic Gradients as well. Thank you!!!

Fyrd:
Thanks! I'll update the conic gradients entry.

Issue #1143, opened 2015-05-12
SVG 2 features (currently drafting)


Fyrd:
SVG 2 features (currently drafting): https://svgwg.org/svg2-draft/single-page.html

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+9__
Issue added by __Bell__ on __2014-08-20__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

Llbe:
I can add to this that some of the features in SVG 2 are available already, for example the paint-order property exists in Chrome and Firefox.

Llbe:
+1

Ugoku:
+1

rotsee:
+1

dcorb:
+1

Schweinepriester:
+1

violacase:
+1

Kay2dan:
+1 pleeasseee!!! We need more SVG more than ever!

codyn329:
+1

pixeltherapy:
+1

artemist:
+1

AmeliaBR:
To clarify: this would not be one table, but many. Browsers are implementing individual features as the spec stabilizes and interest exists, they are not waiting to implement "SVG 2" all at once.

Some examples of features for which compatibility data would be useful:

- `paint-order` (as mentioned, already implemented in Firefox, WebKit & Blink)
- [`z-index` for SVG elements](https://svgwg.org/svg2-draft/render.html#ZIndexProperty)
- [keyboard focus control](https://svgwg.org/svg2-draft/interact.html#Focus) ([`tabindex`](https://svgwg.org/svg2-draft/struct.html#tabindexattribute) and the `focus`/`blur` functions, already supported in Blink & WebKit)
- [geometry properties](https://svgwg.org/svg2-draft/geometry.html) (`x`, `y`, `width`, `height`, `cx`, `cy`, `r`, `rx`, `ry`) specified via CSS (already supported in Blink & WebKit)
- [multi-line SVG text](https://svgwg.org/svg2-draft/text.html) (partial support already in Firefox, for hard line breaks)
- [`white-space` control via CSS](https://svgwg.org/svg2-draft/text.html#WhiteSpace) (some support already, haven't tested recently)
- [SVG-specific text-decoration fill and stroke styling properties](https://svgwg.org/svg2-draft/text.html#TextDecorationProperties) 
[...]

PR #4527, opened 2018-10-04
SVG 2

browner12:
SVG 2 is currently in Candidate Recommendation, and expected to go to Proposed Recommendation sometime after December 4th 2018.

https://www.w3.org/TR/2018/CR-SVG2-20181004/

This is one of those weird features because it started soooo long ago, and then got derailed at some point.  It looks like browser started working on support awhile back, but development has stalled a little while W3C worked out the kinks.

https://www.chromestatus.com/feature/5760616295825408

First time contributing to this repo, tried to follow the contribution guide as best I could, but let me know if there's anything else that would help for this PR.

Thanks!


Fyrd:
Considering SVG 2 is a broad specification with multiple features that can each have varying degrees of support it would be more appropriate to be split out into each single feature. See Issue #1143 that's intended to list the multiple features.

browner12:
thanks for pointing that one out. I'll see what I can do to help on it

PR #4528, opened 2018-10-05
Add Webkit id for Pointer Events

sandstrom:
https://webkit.org/status/#?search=pointer%20events

Couldn't figure out how to link to it via the `webkit_id` attribute. Happy to change if you want me to.


Fyrd:
Looks like the ID for that is `specification-pointer-events-level-2` (if you hover over the feature title there you can see it in the link)

sandstrom:
@Fyrd Thanks, didn't know!

I've updated this PR. Let me know what you think!

(you can use 'squash and merge' to turn the two commits into one)

PR #4531, opened 2018-10-06
update(tabindex): update IE and Edge compatibility with notes

CITguy:
Edge Bug: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/4365703/

I cannot find a documented bug for IE, but I have verified similar behavior through manual testing.


Fyrd:
I don't feel this bug is serious enough to warrant partial support, just listing it under "bugs" seems sufficient

CITguy:
Thanks for the feedback. I was struggling on where I should put it. 

I've updated the PR.

PR #4533, opened 2018-10-06
update(css-nth-child-of): update spec status

CITguy:
* add link to MS Edge platform status
* update spec to reflect published standard

Relates to #1058


Fyrd:
If you set "cssselectorslevel4" as "ie_id" value that should fetch the status automatically.

CITguy:
@Fyrd: Is that all I need to change? What about `spec` or the new link?

PR #4539, opened 2018-10-10
Remove npmignore

alik0211:
npmignore is only needed to publish npm packages


alik0211:
I found the npm package. Excuse me

PR #4549, opened 2018-10-12
viewport-units: Added specifics on vw/vh resize issue

kiding:
The note related to the vw/vh issue has been generalized for somewhat better understanding.

References
- [Benjamin Poulain](https://bugs.webkit.org/show_bug.cgi?id=141832#c5)
- [David Bokan](https://developers.google.com/web/updates/2016/12/url-bar-resizing)

### DIFF
--- a/features-json/viewport-units.json
+++ b/features-json/viewport-units.json
@@ -57,7 +57,7 @@
       "description":"In IE9 inside an iframe, viewport units will be calculated in the context of the parent window and not the iframe."
     },
     {
-      "description":"`vh` on iOS is reported to include the height of the bottom toolbar in the height calculation, and the width of the sidebar (bookmarks) in the `vw` width calculation."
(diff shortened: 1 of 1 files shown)


kiding:
It still applies to “third party” browsers if they change viewport on scroll. DuckDuckGo on iOS behaves the same way.

Fyrd:
Thanks!

kiding:
@Fyrd It seems my changes were overwritten by the bulk "data update" commits: a7eda70420ca9e6b2a53b06fd2f84752b7f42feb and 78c2cefdecf520661fb3947768448c6f50ce33fd. I believe this wasn't your intention, so could you check and revert it back?

I've got a strong feeling several contributions near the commit date were lost like this. We might need to investigate if that's the case.

Fyrd:
@kiding Yeah, I was able to revert the changes made then some time after already. If you look at the current https://github.com/Fyrd/caniuse/blob/master/features-json/viewport-units.json#L60 you'll see your change is there. I did just double-check other PRs merged that day and all the changes seem to be there.

Fyrd:
Not a problem! :)

PR #4550, opened 2018-10-12
Move note about nomodule polyfill for Safari to the right place

chicoxyzzy:
It's not dynamic import polyfill but static import polyfill


Fyrd:
Thanks!

Issue #4305, opened 2018-06-13
CSS Values and Units Function Notations


matthew-dean:
Some browsers have already added support for `min()` and `max()` even though it's barely documented anywhere.  (Added to Webkit here - https://github.com/WebKit/webkit/commit/b6e6b5db9259fffeb38061a5b4748152336765b9#diff-518f9771ccd63924645e1713a5875265)

IMO it would make sense to include the function notations from that spec as a specific group of support - https://drafts.csswg.org/css-values-4/#functional-notations, since it appears that browsers are not implementing the entirety of CSS Values 4 all at once. Then there can be specific notes of which functions (`min()`, `max()`, `clamp()`, `toggle()`) have support in which browser.

Note that the Webkit implementation which, by its documentation, seems to suggest `min()` and `max()` are only allowable within `calc()` is undoubtedly incorrect.

Fyrd:
Math functions now available at https://caniuse.com/#feat=css-math-functions

PR #4552, opened 2018-10-16
Add CSS math functions min(), max() and clamp()

ausi:
See #4305.

`min()` and `max()` is supported since Safari 11.1.


Fyrd:
Thanks!

Issue #3787, opened 2017-10-19
CSS env() support


kanongil:
See https://github.com/w3c/csswg-drafts/issues/1693.

Safari TP currently includes support for this. It is also supported on iOS 11.0 using the previous `constant()` syntax.

toothbrush7777777:
iOS 11.1 uses the current `env()` syntax.

kiding:
> . . . Beginning with Safari Technology Preview 41 and the iOS 11.2 beta, constant() has been removed and replaced with env().

See [WebKit Blog](https://webkit.org/blog/7929/designing-websites-for-iphone-x/).

Malvoz:
+1

chrisblakley:
+1

Malvoz:
https://drafts.csswg.org/css-env-1/

Schweinepriester:
+1 for keeping track of this.

<https://www.chromestatus.com/feature/5710044637167616>: Enabled by default in Chrome for Android 69

ausi:
I created pull request #4555 for this feature.

Malvoz:
@Fyrd 
I'm not sure what happened, but https://caniuse.com/#feat=css-env-function doesn't seem to exist anymore.

Fyrd:
@Malvoz Yeah, not sure what went wrong there. Fixed now!

PR #4555, opened 2018-10-16
Add CSS Environment Variables env()

ausi:
See #3787.

`env()` is supported since Safari 11, Chrome 69, Opera 56.


Malvoz:
Want to add MS Edge feature suggestion to that?
https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/34589707-css-environment-variables

Should probably extend keywords to "CSS variables" here aswell.

ausi:
@Malvoz done.

Malvoz:
Thanks, I have further requests on that though - if a user searches explicitly for e.g. "css env()" or "css variables" this would not show with the current keywords:

`env,safe-area-inset,constant,variables`.

Can I bother you with an updated list on that?

`css env(),constant(),css variables,safe-area-inset-top,safe-area-inset-right,safe-area-inset-bottom,safe-area-left`

ausi:
@Malvoz done.

Fyrd:
Thanks!

kiding:
https://github.com/Fyrd/caniuse/blob/12e8cedf52b3caf5adcfa7ccac31357484c5634d/features-json/css-env-function.json#L348 
I believe `shown` should be true unless there is any particular reason.

PR #4561, opened 2018-10-17
TLS 1.0 and 1.1 deprecated in all modern browsers

daytonlowell:
Links to all sources are included in the links section of the JSON.

### DIFF
--- a/features-json/tls1-1.json
+++ b/features-json/tls1-1.json
@@ -7,6 +7,22 @@
     {
       "url":"https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.1",
       "title":"Wikipedia article on TLS"
+    },
+    {
+      "url":"https://security.googleblog.com/2018/10/modernizing-transport-security.html",
+      "title":"Modernizing Transport Security - Google Security Blog"
+    },
+    {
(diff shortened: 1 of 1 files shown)


features-json/tls1-1.json
```diff
@@ -317,7 +333,7 @@
       "7.12":"y"
     }
   },
-  "notes":"",
+  "notes":"TLS 1.0 & 1.2 are deprecated in Chrome, Edge, Firefox, Internet Explorer 11, & Safari. Support will be removed in March 2020.",
```
Schweinepriester: It should be "TLS 1.0 & 1.1", shouldn't it? 0:-)

```suggestion
  "notes":"TLS 1.0 & 1.1 are deprecated in Chrome, Edge, Firefox, Internet Explorer 11, & Safari. Support will be removed in March 2020.",
```

daytonlowell:
Yup, thanks. Fat fingered it, I guess.

Fyrd:
Thanks!

daytonlowell:
Oh also, I noticed that TLS 1.0 on the site says “This network security protocol is supported in effectively all browsers (since IE6+, Firefox 2+, Chrome 1+ etc)”.

That should probably be changed to say deprecated instead of supported, but I don’t think it’s in the JSON. I couldn’t find it, anyway.

PR #4568, opened 2018-10-19
Add Picture-in-Picture feature

beaufortfrancois:
This PR is about adding brand new Web API for Picture-in-Picture.


features-json/picture-in-picture.json
```diff
@@ -0,0 +1,341 @@
...
+  "usage_perc_y":0,
+  "usage_perc_a":0,
+  "ucprefix":false,
+  "parent":"",
+  "keywords":"Picture-in-Picture,pip",
+  "ie_id":"pictureinpicture",
+  "chrome_id":"5729206566649856",
+  "firefox_id":"",
+  "webkit_id":"",
+  "shown":true
```
HM100: Please change it to false and check if Safari supports it

beaufortfrancois: What is "shown"?
Safari supports a webkit API for Picture-in-Picture which is not the same.

HM100: “shown” is used if the feature is shown on the site. Set it to ‘false’ please and you can include a note that Safari  supports a custom API like who was done on Payment Request API and guessn which versions do have it. Thanks!

beaufortfrancois:
I'm sorry. I don't understand why this feature should not be shown on the
site.

kiding:
>New additions will always start out with "shown": false (regardless of the initial value set in the PR). This is so the data can undergo a certain level of verification to guarantee the correctness of information shown on the site. This verification happens after the pull request has already been accepted because it allows the data to automatically be updated with newly released browser versions when necessary so the pull request won't need to require manual updates during this period.

See `CONTRIBUTING.md` for more information.

beaufortfrancois:
Thank you both!
I've edited PR.

beaufortfrancois:
Thank you both!
I've edited PR.

Fyrd:
Thanks!

beaufortfrancois:
Quick question: When may I expect to see this in https://caniuse.com/#search=picture-in-picture?

PR #4570, opened 2018-10-21
Firefox 64 will support interaction media queries

HM100:
https://bugzilla.mozilla.org/show_bug.cgi?id=1035774 marked as fixed for M64


Fyrd:
Thanks!

PR #4573, opened 2018-10-23
Some removals that will happen in Chrome 73

HM100:
# Links
https://www.chromestatus.com/feature/4642138092470272

https://www.chromestatus.com/feature/5144752345317376

https://www.chromestatus.com/feature/4507242028072960
## Conclusion
That's all the things that Chrome 73 plans to remove in accordance with these status pages


mgol:
It seems Chrome has moved the deadline and Can I Use data no longer reflects reality - those features are still in Chrome.

I submitted #4974 to fix the data.

PR #4578, opened 2018-10-23
Add support for font-variant-ligatures

jpamental:
I'm hoping to add more tests for OpenType features. Test case support page is here:

https://jpamental.github.io/css-examples/font-features/font-variant-ligatures.html

If the feature is supported correctly, the left example should match the right (which is done with the lower-level syntax font-feature-settings)


features-json/font-variant-ligatures.json
```diff
@@ -0,0 +1,340 @@
...
+  "usage_perc_y":0,
+  "usage_perc_a":0,
+  "ucprefix":false,
+  "parent":"",
+  "keywords":"font-feature,font-feature-settings,font-variant",
+  "ie_id":"",
+  "chrome_id":"",
+  "firefox_id":"",
+  "webkit_id":"",
+  "shown":true
```
HM100: Too many unknown support. Please change this line to “shown”:false

jpamental: @HM100 Is it the previous version support, or number of different browsers that's an issue? Just wondering how I can help this along as I'd really like to get more of the font features included.

HM100: The number of previous versions support is almost of unknown support excluding IE and Edge

jpamental: I've updated as requested

jpamental: I found a bunch more detail on when the feature shipped in almost all of the browsers, so updated with that info and set back to shown: true (please let me know if that's a problem and I'll update again)

HM100: and please fix those `x` to `y` or `y x` if those versions with `x` are prefixed supported

HM100: @jpamental Any ETA on fixing the issues?

Fyrd:
Closing this PR as it's already covered by https://caniuse.com/#feat=mdn-css_properties_font-variant-ligatures and doesn't add a significant amount of new info.

PR #4583, opened 2018-10-25
Update css-sel3.json

tomtomau:
Updated to include information about IE11 failures with `last-of-type` CSS3 selector with custom elements

See [StackOverflow Discussion](https://stackoverflow.com/questions/38666233/last-of-type-doesnt-work-with-custom-elements-in-ie11-and-edge#38669965)

Fixed in [Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/5249985/)


Fyrd:
Thanks!

PR #4595, opened 2018-11-01
Firefox 65 supports WebP image format

mkurz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1294490


Schweinepriester:
Faster than me, damn :P

It's also in the changelog: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/65#APIs

PR #4596, opened 2018-11-01
Update pointer-events.json

KIRuhan2:
Feature that should not be ignored.

The SVG 2.0 specification also defines a bounding-box value. When the value of pointer-events is bounding-box, the rectangular area around the element can also receive pointer events. Only Chrome 65+ supports the bounding-box value.

### DIFF
--- a/features-json/pointer-events.json
+++ b/features-json/pointer-events.json
@@ -25,6 +25,9 @@
       "description":"Moving the scrollbar on an object with `pointer-events: none;` works in Firefox, but doesn't work in either Chrome or IE."
     },
     {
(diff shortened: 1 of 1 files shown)


features-json/pointer-events.json
```diff
@@ -24,6 +24,9 @@
     {
       "description":"Moving the scrollbar on an object with `pointer-events: none;` works in Firefox, but doesn't work in either Chrome or IE."
     },
+    {
+      "description":"The SVG 2.0 specification also defines a bounding-box value. When the value of pointer-events is bounding-box, the rectangular area around the element can also receive pointer events. Only Chrome 65+ supports the bounding-box value."
+    }
```
Fyrd:
```suggestion
    },
```
Missing a comma here

Fyrd:
Thanks!

PR #4598, opened 2018-11-03
update(css-sel3): add IE bug for fieldset:disabled

CITguy:


CITguy:
Given the number of bugs/quirks for IE, should its support be updated to "partial"?

Fyrd:
Thanks! I feel the issues are subtle enough that that's probably not necessary.

PR #4612, opened 2018-11-10
iOS supports caret-color

HM100:
According to what I tested on https://awikia.github.io/SourcePro/Source%20Pro.html found that caret-color is recognized. Since iOS 11.3 is based on Safari 11.1, I also marked that version as supported


Fyrd:
Confirmed, thanks!

PR #4617, opened 2018-11-14
web-app-manifest: Update browser support for consistency

mgiuca:
Continued discussion from #4357.

- Removed footnote 1 (this is just stating how to use the feature and doesn't add any information about how it is supported).
- Downgraded Edge support from "yes" to "no". (Technically, Edge does not implement app installation at all. _Bing_ interprets the manifest for store listing, but that is not the Edge user agent. Added a footnote to explain this.)
- Downgraded other browsers from "yes" to "partial" to match Chrome. Chrome on desktop is considered to have "partial" support due to not supporting the `appinstalled` and `beforeinstallprompt` events on macOS. These events are, to my knowledge, not implemented on any other browser except Chrome on Android. So downgrading all of them to "partial" with a footnote.


mgiuca:
@marcoscaceres (editor). Does this match your understanding?

Also I'll note that `beforeinstallprompt` technically never needs to be fired by a user agent, so arguably a user agent that doesn't fire it is still compliant. But you do need to fire `appinstalled` when an app is installed. So I think we could just remove "and `beforeinstallprompt`" from the footnote. Are there any browsers that implement `appinstalled` but not `beforeinstallprompt` that would thus become fully supported?

features-json/web-app-manifest.json
```diff
@@ -296,39 +296,39 @@
...
+      "7.12":"a #1"
     }
   },
   "notes":"",
   "notes_by_num":{
-    "1":"On Windows, Chrome will read the manifest to generate a desktop icon when selecting the \"Add to desktop...\" option.",
-    "2":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported.",
-    "3":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported on all platforms"
+    "1":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported.",
+    "2":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported on all platforms."
```
marcoscaceres:
```suggestion
    "2":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported on all platforms.",
```

features-json/web-app-manifest.json
```diff
@@ -296,39 +296,39 @@
...
+    "3":"The manifest may be used to list apps in the Microsoft Store, which will use Edge as a back-end. However, Edge does not directly implement the Manifest spec."
```
marcoscaceres: @patrickkettner, as this is Microsoft specific, could you kindly check this wording to make sure it's accurate?

marcoscaceres:
> Also I'll note that beforeinstallprompt technically never needs to be fired by a user agent, so arguably a user agent that doesn't fire it is still compliant.

True, tho `BeforeInstallPromptEvent` interface should be on the global object and whatever attribute is supposed to be the event handler for it (even if the AU doesn't fire it).   

>  But you do need to fire appinstalled when an app is installed. So I think we could just remove "and beforeinstallprompt" from the footnote. Are there any browsers that implement appinstalled but not beforeinstallprompt that would thus become fully supported?

Not sure.

Fyrd:
@patrickkettner Any comments?

patrickkettner:
as usual, @mgiuca is correct

marcoscaceres:
Thanks @patrickkettner!  Ok, bar the merge conflict, then we should be good to go on this!

mgiuca:
Thanks folks. Resolved the merge conflict.

Fyrd:
Thanks!

PR #4619, opened 2018-11-14
Edge support of OverscrollBehavior in V18 can't be full support and according to MDN, it supports something missing

HM100:
Because it lacks support of the none value (Actually supported but incorrectly treaten as contain), please see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior#Browser_compatibility) and [Edge 18 Changelog](https://aka.ms/devguide_edgehtml_18)


Fyrd:
Thanks!

PR #4621, opened 2018-11-15
Added new payments pref in Firefox

marcoscaceres:
Firefox introduced a new pref "dom.payments.request.supportedRegions", which takes a list of country codes. Only CA and US are supported right now.

### DIFF
--- a/features-json/payment-request.json
+++ b/features-json/payment-request.json
@@ -350,7 +350,7 @@
     "6":"Can be enabled via the `dom.payments.request.enabled` flag in \"about:config\" flag since 55.",
     "7":"Missing support for PaymentResponse.prototype.retry() method",
     "8":"Enabled by default in Nightly, but \"browser.search.region\" pref in \"about:config\" must be set to 'US'",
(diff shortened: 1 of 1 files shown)


features-json/payment-request.json
```diff
@@ -349,7 +349,8 @@
     "5":"Unlike Desktop Chrome, support has been in Chrome for Android since version 53.",
     "6":"Can be enabled via the `dom.payments.request.enabled` flag in \"about:config\" flag since 55.",
     "7":"Missing support for PaymentResponse.prototype.retry() method",
-    "8":"Enabled by default in Nightly, but \"browser.search.region\" flag in \"about:config\" must be set to 'US'"
+    "8":"Enabled by default in Nightly, but \"browser.search.region\" pref in \"about:config\" must be set to 'US'",
```
mnoorenberghe: I would suggest removing (8) because it's no longer relevant due to being Nightly-only: you can't change that pref in 64 Beta/Release to get the feature. Users should also never be changing that pref so I'd rather not call it out.

marcoscaceres: Good point.

marcoscaceres: So, let's just replace 8 with 9, yeah?

Fyrd:
Thanks!

PR #4623, opened 2018-11-15
 Downgrade SVG fragments support for IE

CyberAP:
IE can't load SVG fragments from external resources, which means it only provides partial support for them.


Fyrd:
Thanks!

PR #4629, opened 2018-11-19
[web-app-manifest.json] Add partial support for FireFox

Malvoz:
Add partial support for FireFox, implemented in Android only (from v58):
https://www.mozilla.org/en-US/firefox/android/58.0/releasenotes/

Also added a few keywords to make it easier to find in search.


Fyrd:
Firefox for Android is already covered by `and_ff` so I think adding it to just "Firefox" is redundant. Feel free to add the keywords in another PR however.

PR #4630, opened 2018-11-19
Safari 12 requires prefix for backface-visibility

caaker:
In Safari, version 12.0 (14606.1.36.1.9), this line of code has no effect, and you can still see the "backface".

`backface-visibility: hidden;`

However, when I use the prefix for Safari, as seen below, the "backface" is now hidden.

`-webkit-backface-visibility: hidden;`

Safari, version 12, still requires the prefix.


HM100:
Why Safari still hasn't properly unprefixed properly the CSS 3d transforms. Ah Apple you suck on unprefixing CSS properties. 😒

caaker:
Here is a fiddle with the fix for Safari.   You can view the code working correctly by un-commenting the "Safari fix" in the CSS.  Are you asking me to create a bug for this?

http://jsfiddle.net/hcuaw7p3/11/

Fyrd:
Confirmed, thanks!
Found a bug filed here: https://bugs.webkit.org/show_bug.cgi?id=170983

PR #4634, opened 2018-11-21
Default media playback controls blocked in Safari with strict CSP

da2x:
Affects Safari 11 and 12 for iOS 11 and 12 and macOS.


Fyrd:
Thanks!

PR #4635, opened 2018-11-22
Update webgl2.json

MaxBittker:
Webgl2 can be enables on ios safari 12 (but not on ios chrome) via the experimental features settings


Fyrd:
Thanks!

PR #4636, opened 2018-11-24
Update IE in insert-adjacent.json

Fuzzyma:
IE does not support that until version 12 (https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement#Browser_compatibility)


HM100:
> Forget what I just said, its there in IE 11. However, I cannot believe that this is true for everything under IE 10. Cant test it though...

You can inspect and change document modes as well to see it

PR #4640, opened 2018-11-27
Safari doesn't support list of origins in the iframe allow attribute

marian-r:


marian-r:
@Fyrd done, thanks for the suggestion.

Fyrd:
Thanks!

PR #4642, opened 2018-11-28
Issues with Firefox in Private Browsingmode.

kyro38:
It's a known bug for years. 
https://bugzilla.mozilla.org/show_bug.cgi?id=781982


Fyrd:
Thanks!

PR #4644, opened 2018-11-29
Fix bugzilla link for :focus-visible

darthmaim:
The old link always forced you to login, the fixed link works anonymously.


Fyrd:
Thanks!

PR #4654, opened 2018-12-05
Fix Public Key Pinning data

FranklinYu:
I tested with Chrome Stable (70) and Beta (71). Maybe some one with Chrome Dev (72) or Chrome Canary (73) can fix those two versions.

Test URL: https://projects.dm.id.lv/Public-Key-Pins_test

Edit: I tested in Chrome Dev (72) and it has been removed. Not sure whether that would be promoted to Chrome Beta in next release cycle.


Fyrd:
Confirmed, thanks!

PR #4655, opened 2018-12-05
Intersection Observer support enabled by default in Safari TP 71

CyberAP:
[Source](https://webkit.org/blog/8517/release-notes-for-safari-technology-preview-71/)


CyberAP:
Ok, this is already submitted, by bad.

PR #4658, opened 2018-12-06
CSS Paint support in STP

daytonlowell:


features-json/css-paint-api.json
```diff
@@ -197,7 +197,7 @@
       "11":"n",
       "11.1":"n",
       "12":"n",
-      "TP":"n"
+      "TP":"a d #1"
```
HM100: Change **a** to **n** since **a d** means without any flag enabled, it has partial support

Fyrd:
Have updated manually.

PR #4663, opened 2018-12-10
es6-module: document issues with Edge

### LINKED ISSUE #3195: http://caniuse.com/#feat=url wrongly claims Edge has full support although it doesn't support the .searchParams property
Part of the [URL living standard](https://url.spec.whatwg.org/#api) is the `searchParams` property:

    [SameObject] readonly attribute URLSearchParams searchParams;

Currently, some browsers (e.g. Edge) implement `URL` objects that do not have this property. They should thus be marked as having partial support at http://caniuse.com/#feat=url, but are instead marked as having full support.

I would assume (but have not tested) that the browser versions missing the `searchParams` property are exactly those that do not implement the `URLSearchParams` class, already noted at http://caniuse.com/#feat=urlsearchparams.

Simple test script to determine support that could be added to http://tests.caniuse.com:

    setResult(new URL('http://example.com').searchParams !== undefined);

Suggested "note" to add to `url.json` for these cases:

> The `.searchParams` property is not supported.

### LINKED ISSUE #4662: `nomodule` is not correctly supported in Edge
https://caniuse.com/#feat=es6-module

Edge does support `nomodule` (i.e. the script is not executed), but the `nomodule` script is still fetched in Edge 16.
Additionally, in Edge 17+ there's a bug where `<script type=module>`  is fetched twice (hence overall there are 3 fetches: 2 module and 1 nomodule) when using module/nomodule in HTML.

See https://gist.github.com/jakub-g/5fc11af85a061ca29cc84892f1059fec

jakub-g:
Test page https://jg-testpage.github.io/es-modules/module-nomodule/simple.html

Result in Edge 17: 
image

Close #4662

### DIFF
--- a/features-json/es6-module.json
+++ b/features-json/es6-module.json
@@ -35,10 +35,19 @@
     {
       "url":"https://hospodarets.com/native-ecmascript-modules-nomodule",
       "title":"Blog post on using nomodule"
+    },
+    {
+      "url":"https://gist.github.com/jakub-g/5fc11af85a061ca29cc84892f1059fec",
+      "title":"Will it double-fetch? Browser behavior with `module` / `nomodule` scripts"
     }
   ],
   "bugs":[
-    
+    {
+      "description": "`nomodule` scripts are not executed in Edge, but they're still fetched"
+    },
+    {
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #4667, opened 2018-12-11
Update status of support IO in Safari 12

Efimenko:
image


daytonlowell:
I believe it's only in Safari Tech Preview, not in Safari 12. source: https://webkit.org/blog/8517/release-notes-for-safari-technology-preview-71/

Fyrd:
Agreed, does not appear to be working in regular Safari 12.

Issue #3450, opened 2017-05-19
Add Clear Site Data header


augbog:
Show browser support for Clear Site Data header as specified in the W3C spec:
https://w3c.github.io/webappsec-clear-site-data/

It looks like a few places have already started implementing or supporting this but it isn't very clear what the status of browser support is? Related links:

https://bugs.chromium.org/p/chromium/issues/detail?id=607897
https://bugzilla.mozilla.org/show_bug.cgi?id=1268889

mkurz:
+1

Schweinepriester:
+1

- https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/63
    > The Clear-Site-Data header is implemented and no longer behind a preference (bug 1470111).
- https://www.chromestatus.com/feature/4713262029471744:
    Chrome 61+, Opera 48+
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data

conanite:
+1

redeye86:
+1

augbog:
PR merged behind flag I think only outstanding work is maybe tests :) gonna go ahead and close.

PR #4672, opened 2018-12-12
Issue #3450 add clear-site-data header

augbog:
Feature request to address #3450 and add support for Clear Site Data header. Related links:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data#Browser_compatibility
https://bugs.chromium.org/p/chromium/issues/detail?id=607897
https://bugzilla.mozilla.org/show_bug.cgi?id=1268889
https://w3c.github.io/webappsec-clear-site-data/
https://www.chromestatus.com/feature/4713262029471744:

I tried my best to follow the Contributing guidelines so please let me know if there's anything I can do. I was hoping to get usage metrics but not sure where to look? Totally willing to investigate if people can point me in the right direct though :)

Please let me know if anything can be improved 🙏


Fyrd:
Thanks! Do you know if there's a test available for this feature somewhere? If not I'll see if I can write one at some point.
As for usage metrics, no need to worry about them, those are automatically generated.

augbog:
Hmm not that I know of, I'll try to look around and will post back if I find any :) thanks for the fast response!

augbog:
After some digging I found web-platform-tests repo have some tests?

https://github.com/web-platform-tests/wpt/tree/master/clear-site-data

Browsers also have tests in their respective repositories but still digging through that -- was the above what you were looking for?

PR #4675, opened 2018-12-13
input-event: Consistent support status

kiding:
IE9 and Firefox 3 have been considered "partial support" due to the note 1 and 2, respectively. The other notes 3 4 5 are not much different from 1 and 2, so the corresponding browsers should also be "partial support."

#### Technical rationale
- Definition: `input` event is ["fired at controls when the user changes the value."](https://html.spec.whatwg.org/multipage/input.html#the-input-element:event-input-2)
- Note 3: [`select` element fires `input` event.](https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element:event-input)
- Note 4: [`<input type="file">` fires `input` event.](https://html.spec.whatwg.org/multipage/input.html#file-upload-state-(type=file):event-input-3)
- Note 5: [Radio button fires `input` event.](https://html.spec.whatwg.org/multipage/input.html#radio-button-state-(type=radio):event-input-2)
- See [the WHATWG issue](https://github.com/whatwg/html/issues/601) for more discussion.


Fyrd:
It's often tricky to decide what counts as partial support and what's just a bug...the reason it's the way it is is because I believe the "input" event on text fields is significantly more useful and used by developers than the event is on those other fields (where the "change" event works well enough).

Because of this, to save developers from spending time figuring out why support for most browsers is "partial" it's just considered a bug for the other fields.

Do you have reason to believe the bugs are significant enough to warrant partial support?

kiding:
Thanks for the comment. I have a feeling this might be blown into a totally different one, but nonetheless here are my thoughts.

I do understand where your thoughts come from. However, to me the reason itself sounds a bit arbitrary and specific only to this feature. How much of a feature should a browser support to be considered a full support? How much usage should that specific portion of a feature have? Maybe we should gather the usage data to determine the threshold... These questions are so tricky to answer, and the solutions themselves get too much complicated.

Personally, something that shows "Partial Support" has been an immediate yellow flag (quite literally) that forces me to learn more before writing any more code. I believe that the immediacy—and the repulse that follows—is what makes Can I Use incredibly useful to web developers. This case as an example: Those who implement `<input type=file>` with `input` event simply can not avoid spending more time if they want cross-platform support. Coming back to debug would be tedious enough, but wouldn't it be nice to at least show it more apparently right away?

I think the issue boils down to the fact that "Partial Support" is somewhat ill-defined. In my opinion, a browser supports a feature "partially" when it _omits_ a sub-feature of a given feature, where the feature itself is described by the reference specification, which might change over time. The feature has "a bug/issue" when the implemented sub-feature behaves wrongly or erratically (returning unexpected results) or maybe differently (unlike other browsers). I see most of the features on Can I Use already fit this definition. Some features that came all the way from the old days of _no spec at all_ would be very tricky for sure though.

I'd love to hear any thoughts on this. I hope my intention went through the language barrier intact.

kiding:
https://github.com/Fyrd/caniuse/blob/27a65d6a111435ccb9d676da5962c963045cfcf3/features-json/datalist.json#L374-L378

Adding to this as a reference: Some of the entries read "partial support due to a bug" which is not consistent with the (implicit) policy.

PR #4687, opened 2018-12-18
Correct URL for readonly spec

spaceninja:
The spec link for the `readonly` attribute was pointing to the wrong place. This commit fixes the link.


Fyrd:
Thanks!

PR #4688, opened 2018-12-19
FF 65: break-* now available

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/65#CSS>

> The break-* properties have been implemented, and the legacy page-break-* properties have been aliased to them (bug 775618):
> - break-before is now an alias for page-break-before.
> - break-after is now an alias for page-break-after.
> - break-inside is now an alias for page-break-inside.

<https://bugzilla.mozilla.org/show_bug.cgi?id=775618>


Fyrd:
Thanks!

PR #4695, opened 2018-12-25
iOS >= 11.3 are supporting Subresource Integrity

dadon-david:
Hi,
Safari on iOS 11.3 and above are supporting Subresource integrity

"Safari 11.1 ships with iOS 11.3 and macOS 10.13.4. It is also available for macOS 10.12.6 and 10.11.6."
Source: https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html#//apple_ref/doc/uid/TP40014305-CH14-SW7

Merry Christmas !


Fyrd:
Thanks!

PR #4697, opened 2018-12-27
Add AVIF (AV1 Image File Format) support

### LINKED ISSUE #4684: Add .avif (AV1 Image File Format) support
The AV1 Image File Format (AVIF) is a new image format based on the AV1 video codec. It would be great if support for this format could be added.

The AVIF spec can be found here: https://aomediacodec.github.io/av1-avif/

Netflix published a bunch of test images: http://download.opencontent.netflix.com/?prefix=AV1/Chimera/AVIF/

EwoutH:
Added support for the AV1 Image File Format (AVIF). It's currently not supported in any browser, but Mozilla and Microsoft are working on it.

Closes #4684.


Fyrd:
I prefer not to include features that have no support at all yet. The Mozilla bug doesn't seem to indicate it's actually being worked on yet ( FLIF has had [a bug open for 3 years](https://bugzilla.mozilla.org/show_bug.cgi?id=flif) without being added yet) and with Edge switching to Chromium it doesn't seem as likely that that version will include AVIF support.

So I'm going to close this for now, once there's support (even behind a feature flag) in a browser I'll be happy to reopen and add to the site.

PR #4701, opened 2018-12-30
Chromium supports border-image-repeat: space

SelenIT:
Chromium supports border-image-repeat: space since version 56 (according to MDN), so other browsers based on corresponding Chromium versions should do also.

It also may be worth noting that Blink and Edge seem to have "overfixed" the issue from the first note, reading the exception in the old spec too broad and creating another bug ([spec discussion](https://github.com/w3c/csswg-drafts/issues/655), [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=767352)). Should it be added to the table?


Fyrd:
Thanks! Yeah that seems worth noting, though probably could just be added to the "bugs" section.

Issue #4699, opened 2018-12-29
Update AV1 info (links and notes)


DeeDeeG:
_Edit 4th February 2019: This original comment has been addressed, and the data is largely up-to-date now. Keeping this issue open though, since AV1 support is still changing quickly._

_For more up-to-date info, see comments at the bottom, and/or check the history of [Pull Requests involving AV1 in this repository...](https://github.com/Fyrd/caniuse/pulls?utf8=%E2%9C%93&q=is%3Apr+AV1)_

_Original comment is left below for reference, though it is out-of-date and can be safely ignored at this point._

---

Hi,

Now that AV1 post-1.0 has been out for a while, support is getting more and more stable in more browsers.

As such, a few of the links and notes can be updated for [`av1.json`](https://github.com/Fyrd/caniuse/blob/091eaea35a8d5549936be18eb1f697a5caf5aaff/features-json/av1.json):

### Links

```json
      "url":"https://bitmovin.com/demos/av1",
      "title":"Sample video (Firefox Nightly only)"
```

- ~~This works in Chrome stable, and also Firefox stable if the relevant flag is enabled.~~

_Edit: This has been addressed in #4708 ([this line](https://github.com/Fyrd/caniuse/pull/4708/files#diff-bf0a7e7f79fea88f9aebafb1c8412bf3R13))._

---

```json
      "url":"https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/34544281-support-aomedia-video-1-av1",
      "title":"Microsoft Edge feature request on UserVoice"
```

- ~~I believe the Windows Nov 2018 update introduced AV1 support across Windows, including Edge. Wikipedia says so, anyway: https://en.wikipedia.org/wiki/AV1#Software Maybe remove this note?~~

On further reading, AV1 in Edge still requires the AV1 extension pack for Windows, which is still listed as "beta". https://www.microsoft.com/en-us/p/av1-video-extension-beta/9mvzqvxjbq9v?activetab=pivot:overviewtab

However, [Edge will be switching to be a Chromium-based browser](https://blogs.windows.com/windowsexperience/2018/12/06/microsoft-edge-making-the-web-better-through-more-open-source-collaboration/) soon, and while they could conceivably disable AV1 decode, I would not expect them to. So Edge may see AV1 support via the Chromium code base some time soon.

---

```json
      "url":"https://www.facebook.com/330716120785217/videos/330723190784510/",
      "title":"Sample video from Facebook"
```

[...]

PR #4708, opened 2019-01-03
Update av1.json

DeeDeeG:
Addresses #4699 for now (although AV1 support in browsers is still under active development, so might update again soon.) I can't easily test the mobile browsers, or things like Baidu browser, etc, but this has all the updates for the "Big four" and Opera on desktop to the best of my knowledge.

Summary of changes:
* Opera 57+ now supports AV1 out-of-the-box [[1]](https://www.chromestatus.com/features/5729898442260480)
* Firefox Beta and Release 63+ now support AV1, still off by default and behind a preference [[2]](https://bugzilla.mozilla.org/show_bug.cgi?id=1478005)
* Bitmovin demo is now usable with multiple browsers 
  * (Because Chrome, Opera and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1445683#c137) all use AV1 1.0 now, and the Bitmovin demo also uses AV1 1.0 (I have to assume the support in Edge is AV1 1.0...))
  * In any case, I have personally tested the demo with Firefox and Chrome. It does work cross-browser.

### DIFF
--- a/features-json/av1.json
+++ b/features-json/av1.json
@@ -10,7 +10,7 @@
     },
     {
       "url":"https://bitmovin.com/demos/av1",
-      "title":"Sample video (Firefox Nightly only)"
+      "title":"Sample video from Bitmovin"
     },
     {
       "url":"https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/34544281-support-aomedia-video-1-av1",
@@ -44,7 +44,7 @@
       "15":"n",
       "16":"n",
       "17":"n",
-      "18":"n d #3"
+      "18":"n d #4"
     },
     "firefox":{
       "2":"n",
@@ -110,10 +110,10 @@
       "60":"n d #1",
       "61":"n d #1",
       "62":"n d #1",
-      "63":"n d #1",
-      "64":"n d #1",
-      "65":"n d #1",
-      "66":"n d #1"
+      "63":"n d #2",
+      "64":"n d #2",
+      "65":"n d #2",
+      "66":"n d #2"
     },
     "chrome":{
       "4":"n",
@@ -179,9 +179,9 @@
       "64":"n",
       "65":"n",
       "66":"n",
-      "67":"n d #2",
-      "68":"n d #2",
-      "69":"n d #2",
+      "67":"n d #3",
+      "68":"n d #3",
+      "69":"n d #3",
       "70":"y",
       "71":"y",
       "72":"y",
@@ -262,7 +262,7 @@
       "54":"n",
       "55":"n",
       "56":"n",
-      "57":"n"
+      "57":"y"
     },
     "ios_saf":{
       "3.2":"n",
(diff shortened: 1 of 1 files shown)


Eric-01:
Since you're working on it, I have a small request for you. Could you update the info about AV1 support in Firefox to include the changes in version 65.0 and later? In these versions the support for AV1 is enabled by default for Windows users ([bug on Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1452146)).

My suggestion is to add note 3 with the following message: `Enabled by default only for Windows users` (or similar) next to the note 2 (example: `"65":"n d #2"` → `"65":"n d #2 #3"`) in information about Firefox version 65.0 & 66.0. But it's just _my_ suggestion, you can do what you want :)

DeeDeeG:
I am not sure if they will keep it enabled in 65 past beta, but i guess we will see soon.

And unfortunately support details that fine can get complicated quickly. I'm not sure if any of the support mentioned in the data extends to Windows 32-bit yet, for example, since it has been not compiled for that platform for a while now. Hmmm.... So I'm not sure if it is worth that many notes. Maybe we should add the link to the [tracking bug](https://bugzilla.mozilla.org/show_bug.cgi?id=AV1) for Firefox to the links or notes section?

DeeDeeG:
Here's the bog for Windows 32-bit support: https://bugzilla.mozilla.org/show_bug.cgi?id=1475564

Eric-01:
You are probably right about the support on 32-bit Firefox on Windows. The appropriate bug is still open and I also checked whether the 32-bit Firefox Nightly (updated to the newest build) on 32-bit Windows 7 is able to decode AV1 (on the built-in default decoder and dav1d) or not. The result was expected – "not".

The best  section for the info about support only on 64-bit Firefox on Windows would be "notes_by_num", since having max 3 notes next to version 65.0 or 66.0 is not too much, but on the other hand "notes" section also looks great for it. Link to the tracking bug (with text e.g. `Firefox implementation bug`) could go to the "links" section, I think.

Eric-01:
You are probably right about the support on 32-bit Firefox on Windows. The appropriate bug is still open and I also checked whether the 32-bit Firefox Nightly (updated to the newest build) on 32-bit Windows 7 is able to decode AV1 (on the built-in default decoder and dav1d) or not. The result was expected – "not".

The best section for the info about support only on 64-bit Firefox on Windows would be "notes_by_num", since having max 3 notes next to version 65.0 or 66.0 is not too much, but on the other hand "notes" section also looks great for it. Link to the tracking bug (with text e.g. `Firefox implementation bug`) could go to the "links" section, I think.

Eric-01:
You are probably right about the support on 32-bit Firefox on Windows. The appropriate bug is still open and I also checked whether the 32-bit Firefox Nightly (updated to the newest build) on 32-bit Windows 7 is able to decode AV1 (on the built-in default decoder and dav1d) or not. The result was expected – "not".

The best section for the info about support only on 64-bit Firefox on Windows would be "notes_by_num", since having max 3 notes next to version 65.0 or 66.0 is not too many, but on the other hand "notes" section also looks great for it. Link to the tracking bug (with text e.g. `Firefox implementation bug`) could go to the "links" section, I think.

Fyrd:
Merging this as-is, any further changes can be added in new PRs. Thanks!

PR #4713, opened 2019-01-09
Updates AV1.json to reflect support in Firefox 65+

EwoutH:
AV1 is enabled by default since Firefox 65: 
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1452146
 - https://github.com/mozilla/gecko/commit/977e1baae9481239475ae4eb719d4779d977d863
 - https://hg.mozilla.org/releases/mozilla-beta/rev/d9d17041ca84

PR #4716, opened 2019-01-10
Updated info about AV1 support in Firefox 65.0 and 66.0

Eric-01:
As the title says, I updated the information about AV1 support in Firefox version 65.0 and 66.0.

I'm not sure, whether the support for AV1 will "ride the trains" into release channel, or maybe it will stay in beta for a longer time, but since #4713 was merged, these information should be specified ASAP that AV1 is enabled by default ONLY for Windows users (for now).

Reference:
https://bugzilla.mozilla.org/show_bug.cgi?id=1452146

### DIFF
--- a/features-json/av1.json
+++ b/features-json/av1.json
@@ -13,12 +13,16 @@
       "title":"Sample video from Bitmovin"
     },
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #4720, opened 2019-01-11
Firefox 63 for Android has TLS 1.3 support

### LINKED ISSUE #4709: Firefox for Android TLS 1.3 support
I'm using Firefox for Android 64 (release) right now.
Caniuse says that 63 is the current version still.

And according to qualys and cloudflare 64 supports tls 1.3
https://www.ssllabs.com/ssltest/viewMyClient.html
https://www.cloudflare.com/ssl/encrypted-sni/

reading changelogs it seems to have been supported since 61 actually:
https://www.mozilla.org/en-US/firefox/android/61.0/releasenotes/
(though it doesn't name the exact draft number, but on-by-default)

reading
https://blog.mozilla.org/security/2018/08/13/tls-1-3-published-in-firefox-today/
61 was the final draft but only 63 and after use the correct version number, you might also want to update the infos on Firefox desktop accordingly
(#4153 also is about updating this kind of info; support is only draft 28 and higher)

daytonlowell:
Source: https://blog.mozilla.org/security/2018/08/13/tls-1-3-published-in-firefox-today/

Resolves #4709


Fyrd:
Thanks!

PR #4722, opened 2019-01-14
Adds node_js versions 8 and 10 to travis.yml

EwoutH:


EwoutH:
Maybe also remove v0.12 and v4 since they're out of maintenance/LTS.

Fyrd:
Looks good. Thanks!

PR #4726, opened 2019-01-17
FX65+ will support Enviromental Variables

HM100:
Thanks to [this page](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/65#CSS), I intend to change unknown support to supported. Thank you!!!!


Fyrd:
Thanks!

PR #4727, opened 2019-01-18
Add info about Safari display: contents bugs

mgol:
Safari support is buggy, see
* Toggling `display: contents` to `display: none` fails to hide the element https://bugs.webkit.org/show_bug.cgi?id=188259
* `::before`/`::after` elements not filling their grid cell when container has display: contents  https://bugs.webkit.org/show_bug.cgi?id=193567


Fyrd:
Thanks!

PR #4731, opened 2019-01-22
Android Pie web views support rrggbbaa hex color notation

daytonlowell:
Source: https://bugs.chromium.org/p/chromium/issues/detail?id=618472


Fyrd:
Thanks!

PR #4733, opened 2019-01-24
Marking IE support as partial

faizanv:
IE has a ton of known issues with `calc` which will cause it to not work properly and as such it should be classified as partially supported.


HM100:
Same thing was in Flexbox as well

Fyrd:
Seems fair. Thanks!

PR #4741, opened 2019-01-26
Updated HTTP2 notes

daytonlowell:
I moved the numbered note about only supported via TLS(https) to a general note as every implementation of HTTP2 only supported it via TLS. Having that numbered note added a lot of visual clutter and seemed unnecessary.


Fyrd:
Good move, thanks!

PR #4743, opened 2019-01-28
Adds vowAll, an ES5 alternative to Promise.All

Paul-Browne:


Fyrd:
Seems specific to just one use of promises and only for network requests. While I'm sure it's useful for particular use cases it doesn't seem quite appropriate here.

PR #4748, opened 2019-01-30
Update multicolumn.json

mattbrundage:
Mozilla recently aliased the page-break-* properties to break-* https://bugzilla.mozilla.org/show_bug.cgi?id=775618
We're still waiting for column breaks as part of the CSS Fragmentation Module Level 3: https://bugzilla.mozilla.org/show_bug.cgi?id=549114


HM100:
They suck on implementing some properties Microsoft has already added Pm(In this case, multi-column)

Fyrd:
Thanks!

Issue #4749, opened 2019-01-30
Update HEIF support


marzn:
Actually https://caniuse.com/#feat=heif states, that HEIF is not supported by any browser. At least Firefox and Chrome support HEIF. I will provide a PR later.

marzn:
May my tests were not correct, cause I tested it with https://compare.rokka.io/_compare/ which draws the HEIF images via a canvas element and therefor it is no native support by the browser.

Fyrd:
Alright, thanks for confirming.

PR #4750, opened 2019-01-30
#4749 add browser support for HEIF image format

marzn:
I updated the browser support of HEIF image format for the browsers I was able to test it. It may work in some more browsers.


marzn:
May my tests were not correct, cause I tested it with https://compare.rokka.io/_compare/ which draws the HEIF images via a canvas element and therefor it is no native support by the browser.

PR #4756, opened 2019-02-01
Add note about `replace()` on classList for IE11

sklirg:
I found out that IE11 does not support `classList.replace()`, so I figured I'd let others know 😄. I don't have IE10 available to test on.

Not sure if this relates to #3326 specifically.


HM100:
You can test IE10 directly from IE11 by setting document mode to IE10 standard mode. @sklirg 
Same for IE5 to IE9 apart from IE6

Fyrd:
Thanks!

PR #4759, opened 2019-02-02
AV1: data fix (enabled by default in Firefox 65+, not 66+)

DeeDeeG:
After one of the data update commits, a change in Firefox support for AV1 got moved in the data from "65+" to "66+" instead.

This PR corrects the error introduced with the "data update" commit. 

_(Here is the commit that introduced the error: https://github.com/Fyrd/caniuse/commit/6c8252213b0f1d741be3e56ad8de21d67a787a34#diff-bf0a7e7f79fea88f9aebafb1c8412bf3)_

- Edit to add: Release notes showing this change was in 65: https://www.mozilla.org/en-US/firefox/65.0/releasenotes/#note-787847


DeeDeeG:
By the way this PR's change is identical to the newer #4761.

I should have mentioned this PR in my issue (#4699), but to be honest I didn't think anyone was reading it! I just tried to collect the sources and state of things there in case it was helpful!

Fyrd:
Yeah, that was my mistake. Was doing some quick testing on Firefox 65 on Mac and noticed there was no support, so overwrote it without paying enough attention. Just fixed this with #4745 though, so closing this one.

PR #4763, opened 2019-02-05
css-display-contents: make links clickable

mgol:
I understand Markdown links are supported based on https://github.com/Fyrd/caniuse/blob/master/features-json/css-snappoints.json


Fyrd:
Thanks!

PR #4765, opened 2019-02-07
Update do-not-track.json

mattbrundage:
Safari 12.1 removes support for DNT: https://developer.apple.com/documentation/safari_release_notes/safari_12_1_release_notes


HM100:
If this happens, then  browsers will follow that route as well

Fyrd:
Thanks!

PR #4766, opened 2019-02-10
av1.json: Firefox for Android can play AV1

DeeDeeG:
### Summary and context

The AV1 feature is compiled for Firefox for Android now.

In my testing the browser does attempt to play AV1 files now. That said, AV1 decoding is still off by default, behind a user preference flag.

### Changes

This PR changes the support status of Firefox for Android 64 from "None" to "Off by default" and tags it with the note as follows:

- "3":"Can be enabled in (all editions of) Firefox via the `media.av1.enabled` flag in `about:config`"

### Source

Source (bugzilla.mozila.org): 
- [Bug 1368843 - libaom doesn't build for android](https://bugzilla.mozilla.org/show_bug.cgi?id=1368843)
  - Status: Resolved - Fixed in Firefox 64


Fyrd:
Thanks!

PR #4778, opened 2019-02-19
AV1: More-accurate Firefox history

DeeDeeG:
### Changes made

- Updated the text of some notes, to separate out "AV1 being available only in Nightly" to its own note, and there being "a preference to flip (to enable AV1)" as its own note.
- Updated which versions are tagged with which notes.
  - Nightly has AV1 starting with 55, rather than 58, which is what the data say now.
  - Some older releases have AV1 but no pref, so don't mention the pref for those (55 through 60).

### Info supporting changes

AV1 has been available in Firefox Nightly for a while (back to Nightly 55).
- [Bug 1370978 - Enable av1 decoding in Nightly on mac and linux](https://bugzilla.mozilla.org/show_bug.cgi?id=1370978)
  - Status: `RESOLVED FIXED in Firefox 55`
- [Bug 1369932 - libaom doesn't build for windows](https://bugzilla.mozilla.org/show_bug.cgi?id=1369932)
  - Status: `RESOLVED FIXED in Firefox 56 `

The user preference for AV1 ("media.av1.enabled") goes back to Nightly 61.
- [Bug 1451926 - Add a pref around AV1 decoding](https://bugzilla.mozilla.org/show_bug.cgi?id=1451926)
  - Status: `RESOLVED FIXED in Firefox 61`


DeeDeeG:
Another source to back up this change: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#WebRTC_and_media

---

(I read that page, saw it didn't match caniuse, and was inspired to double-check against actual Nightly releases. Here are my testing results...)

```dyalog
54.0a1 (2017-01-24): no av1 pref, clips don't play
55.0a1 (2017-03-07): no av1 pref, clips don't play
55.0a1 (2017-04-19): no av1 pref, clips don't play
* 55.0a1 (2017-06-10): no av1 pref, clip plays: aadbb0251996c8ebb8310567bea330ab7ae9abe4
56.0a1 (2017-06-13): no av1 pref, clip plays: aadbb025
57.0a1 (2017-08-03): no av1 pref, clip plays: aadbb025
58.0a1 (2017-09-22): no av1 pref, clip plays: f5bdeac22930ff4c6b219be49c843db35970b918
59.0a1 (2017-11-14): no av1 pref, clip plays: e87fb2378f01103d5d6e477a4ef6892dc714e614-configure
60.0a1 (2018-01-23): no av1 pref, clip plays: e87fb2378f01103d5d6e477a4ef6892dc714e614-configure
61.0a1 (2018-03-13): no av1 pref, clip plays: e87fb2378f01103d5d6e477a4ef6892dc714e614-configure
* 61.0a1 (2018-04-07): av1 pref set to "false" by default, clip plays when enabled: e87fb2378f01103d5d6e477a4ef6892dc714e614-configure
62.0a1 (2018-05-08): av1 pref set to "true" by default, clip plays: e87fb2378f01103d5d6e477a4ef6892dc714e614-configure

extra:

63.0a1 (2018-07-25): av1 pref set to "false" by default, all AV1 post-v1.0.0 clips in Webm containers will play.
```

First `*` shows that AV1 decode was in fact added late in the 55 cycle.
Second `*` shows that the preference `media.av1.enabled` was added late-ish in the 61 cycle.

If anyone wants my test clips made with the old libaom that Firefox nightly had at the time, I can figure out a way to share them. (They're the first 15 seconds of [Big Buck Bunny](https://en.wikipedia.org/wiki/Big_Buck_Bunny))

(It appears the pref went from off by default, to on by default, to off by default again somehwere between 61 and 63. Accurately capturing this in the caniuse.com data would be a headache, and people aren't going back and trying to use old nightlies as their daily driver browsers, or at least they shouldn't do that, so I think it's not worth trying to express that much detail.)

Fyrd:
Thanks!

PR #4782, opened 2019-02-25
Overflow hidden bug

StepFlow:
https://github.com/w3c/csswg-drafts/issues/865

### DIFF
--- a/features-json/css-sticky.json
+++ b/features-json/css-sticky.json
@@ -35,6 +35,9 @@
     },
     {
       "description":"A parent with overflow set to `auto` will prevent `position: sticky` from working in Safari"
+    },
+    {
+      "description":"One of the all of nearest element ancestor with a scrolling box children with overflow set to `hidden` will prevent `position: sticky` from working"
     }
   ],
   "categories":[


Fyrd:
I think this should be rephrased: "One of the all of nearest element ancestor"...not clear to me what that means.

jonjohnjohnson:
How about this?

"Any ancestor between the sticky element and it's user scrollable container with overflow computed as anything but visible/clip will effectively prevent sticking behavior."

And this would consolidate this third description with the second about safari and auto overflow.

Fyrd:
Is that actually a browser bug or just spec-defined behavior? If the latter it should be listed under Notes instead. If it's the former it should mention which browsers are affected.

jonjohnjohnson:
It's the latter. This clarification should be listed under Notes.

Fyrd:
Has been added under Notes, so closing this PR.

jonjohnjohnson:
@Fyrd I see you added my suggestion in the non-numbered notes, but I'd also suggest removing the second bug from the bugs list.

> A parent with overflow set to auto will prevent position: sticky from working in Safari
https://github.com/Fyrd/caniuse/blob/master/features-json/css-sticky.json#L37

First, it's not a bug as far as the spec is concerned. And two, the addition of the note encompasses what that "bug" was trying to communicate.

PR #4783, opened 2019-02-26
Add template literal bug in Safari 12

jridgewell:
Safari 12 has a [bug](https://bugs.webkit.org/show_bug.cgi?id=190756) that breaks caching of `TemplateStrings`.

Re: https://github.com/kangax/compat-table/pull/1424

### DIFF
--- a/features-json/template-literals.json
+++ b/features-json/template-literals.json
@@ -14,7 +14,9 @@
     }
   ],
   "bugs":[
-    
+    {
+      "description":"Safari 12 [sometimes garbage collects](https://bugs.webkit.org/show_bug.cgi?id=190756) the cached TemplateStrings used by Tagged Template Literals"
+    }
   ],
   "categories":[
     "JS",
@@ -200,9 +202,9 @@
       "10.1":"y",
       "11":"y",
       "11.1":"y",
-      "12":"y",
-      "12.1":"y",
-      "TP":"y"
+      "12":"a",
+      "12.1":"a",
+      "TP":"a"
     },
     "opera":{
       "9":"n",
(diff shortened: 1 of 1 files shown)


features-json/template-literals.json
```diff
@@ -200,9 +202,9 @@
       "10.1":"y",
       "11":"y",
       "11.1":"y",
-      "12":"y",
-      "12.1":"y",
-      "TP":"y"
+      "12":"a",
+      "12.1":"a",
```
chicoxyzzy: it's fixed in 12.1. PR to compat-table is here https://github.com/kangax/compat-table/pull/1427

chicoxyzzy:
```suggestion
      "12.1":"y",
```

features-json/template-literals.json
```diff
@@ -200,9 +202,9 @@
...
+      "TP":"a"
```
chicoxyzzy:
```suggestion
      "TP":"y"
```

features-json/template-literals.json
```diff
@@ -275,8 +277,8 @@
       "10.3":"y",
       "11.0-11.2":"y",
       "11.3-11.4":"y",
-      "12.0-12.1":"y",
-      "12.2":"y"
+      "12.0-12.1":"a",
+      "12.2":"a"
```
chicoxyzzy:
```suggestion
      "12.2":"y"
```

features-json/template-literals.json
chicoxyzzy: Actually bug is still there so LGTM

Fyrd:
Usually bugs don't really warrant changing support to partial so I'd suggest switching these values back to `y`. If you still feel strongly about it, please add a numbered note so it's clear what partial support refers to.

Fyrd:
Updated this one manually.

PR #4790, opened 2019-03-03
Remove duplicate html5tutorial.info link

cmthornton:
The `http://www.html5tutorial.info/html5-color.php` link existed twice.


Fyrd:
Thanks!

PR #4792, opened 2019-03-04
flip `shown` to true for server-timing

cvazac:


Fyrd:
This feature hadn't been enabled yet because the support data had not yet been reviewed. Have just done some extra research on server timing and updated the data (notable Firefox support has been added)

PR #4793, opened 2019-03-04
W3C released Webauthn on March 4th, 2019

jeroenhabets:
Updated status and URL after W3C released Webauthn as a standard ("W3C Recommendation")


Fyrd:
Thanks!

PR #4806, opened 2019-03-10
Remove fixed Firefox scroll bug

Y0hy0h:
Removes the statement about dragging not causing scrolling.

The [linked bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=41708) has been marked as resolved and fixed. I have tested that it works with scrolling the window. However, I haven't tested scrollable regions embedded in the window.


Fyrd:
Thanks!

PR #4814, opened 2019-03-13
Chrome 73: Progressive Web Apps / Web App Manifest everywhere(?)

Schweinepriester:
<https://developers.google.com/web/updates/2019/03/nic73>

>  In Chrome 73, we’ve added support for macOS, bringing support for Progressive Web Apps to all desktop platforms - Mac, Windows, Chrome OS and Linux, as well as mobile, simplifying web app development.

Was not sure if <http://tests.caniuse.com/?feat=web-app-manifest> => <http://tests.caniuse.com/web-manifest.php> really suffices, thus I tested it using <https://www.chromestatus.com>:

Bildschirmfoto 2019-03-14 um 00 31 23

Bildschirmfoto 2019-03-14 um 00 25 49

(Snippet to test `appinstalled` from <https://developer.mozilla.org/en-US/docs/Web/Events/appinstalled>, executed before pressing install)

So… seems to work everywhere?


Fyrd:
Agreed. Thanks!

PR #4824, opened 2019-03-20
Safari TP 78: Pointer Events by default

Schweinepriester:
<https://webkit.org/blog/8676/release-notes-for-safari-technology-preview-78/>

> ### Pointer Events
> - Enabled support for Pointer Events by default (r242232)
> - Added support for Pointer Events on macOS (r242137)

<https://bugs.webkit.org/show_bug.cgi?id=195156#c0> and <https://bugs.webkit.org/show_bug.cgi?id=195008#c0> suggest this is (coming) in iOS too, but I guess we'll have to check once Safari 12.1/iOS 12.2 is released.


Fyrd:
Thanks!

PR #4826, opened 2019-03-21
Streams: update support for Firefox

myakura:
* ReadableStream is enabled by default in Firefox 65
* `pipeTo` and `pipeThrough` are not implemented


Fyrd:
Looks good, thanks!

PR #4828, opened 2019-03-23
Updated notes section in Shadow DOM v1

Eric-01:
I removed info that Shadow DOM v0 is being implemented by Chrome and Opera, unlike other browsers, since it is no longer true.


Fyrd:
Good one. Thanks!

PR #4832, opened 2019-03-25
Link to the released version of a spec instead of to a draft, if available

slokhorst:


Fyrd:
The draft specification URLs are used intentionally, so people can see what the actual latest version of the specification looks like. This matches what MDN does, for example:
https://developer.mozilla.org/en-US/docs/Web/API/CSS_Counter_Styles#Specifications
https://developer.mozilla.org/en-US/docs/Web/CSS/unicode-bidi#Specification

slokhorst:
Ah okay, but this is quite inconsistent right now. A lot of other standards do point to the latest release instead of to the draft. For example:

- [credential-management](https://github.com/Fyrd/caniuse/blob/master/features-json/credential-management.json) points to https://www.w3.org/TR/credential-management-1/ instead of https://w3c.github.io/webappsec-credential-management/
- [css-conic-gradients](https://github.com/Fyrd/caniuse/blob/master/features-json/css-conic-gradients.json) points to https://www.w3.org/TR/css-images-4/#conic-gradients instead of https://drafts.csswg.org/css-images-4/#conic-gradients

Would it be preferable to change those instead?

Fyrd:
It would, yes! I don't think that many people bother to check the specs so consistency hasn't been a priority. Fixes are appreciated though!

slokhorst:
Ok.. looks like 90% of pages now link to the published spec though, so it will be quite some work to convert them all. I'll see if I can find some time in the future.

Also I'm not entirely conviced that we should be linking to drafts, as web browsers should implement standardized and stable specs and not drafts that might change often. As caniuse is mostly used to see which browsers implement what, it thus would make sense to link to the stable spec that is most likely implemented.

PR #4833, opened 2019-03-26
macOS 10.14.4 has TLS 1.3 support by default

### LINKED ISSUE #4677: TLS 1.3 can be enabled in Safari
According to multiple sources, TLS 1.3 can be enabled in Safari with the following command

```sh
sudo defaults write /Library/Preferences/com.apple.networkd tcp_connect_enable_tls13 1
```

It can be enabled in iOS version as well (more info in the sources at the end)

That is not shown @ caniuse.com

image

https://superuser.com/questions/1217220/how-do-i-enable-tls-1-3-in-ios-and-macos
https://geekflare.com/enable-tls-1-3-in-browsers/
https://www.ietf.org/mail-archive/web/tls/current/msg23599.html

daytonlowell:
Fixes #4677

### DIFF
--- a/features-json/tls1-3.json
+++ b/features-json/tls1-3.json
@@ -206,8 +206,8 @@
       "11":"n",
       "11.1":"n",
       "12":"n",
-      "12.1":"n",
-      "TP":"n"
+      "12.1":"y #5",
+      "TP":"y #5"
     },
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #4837, opened 2019-03-27
Update server-timings data to reflect Safari support

stephenreay:


Fyrd:
Doesn't seem to pass my test. What's the source for this?

stephenreay:
Opening safari and looking in the network tab? What other test would you expect this to pass?

Fyrd:
Aah, I see. My tests currently only cover the PerformanceServerTiming API as defined at https://www.w3.org/TR/server-timing/#the-performanceservertiming-interface, which Safari does not (yet) support. So will need to change this to partial support as it's limited to manually checking the dev tools.

stephenreay:
I see. I didn't even know there was a JS API for it!

Fyrd:
Great, thanks!

PR #4838, opened 2019-03-27
Added issue for IE

liamhubers:
Just like Firefox, IE doesn't support the use of calc inside color functions. Firefix says to have fixed this, but I haven't tested it


Fyrd:
Thanks!

PR #4839, opened 2019-03-28
fullscreen api working on ipad

HolgerJeromin:
Source: own experiments and
https://dev.to/delahoc/comment/7pfh

https://developer.apple.com/videos/play/wwdc2018/234/ at 33:35 min


Fyrd:
Confirmed, thanks!

PR #4840, opened 2019-03-28
Update datalist.json

mfranzke:
added the supporting 12.2 version to the list


HM100:
Invalid PR. It adds new browser versiom

Fyrd:
Yeah, until Safari 12.2 is properly added you can't include it. Also for now "Safari TP" is sufficient to predict what'll be in 12.2.

PR #4856, opened 2019-04-06
WebM: Safari's WebRTC-only VP8 support

kiding:
Safari 12.1 / iOS 12.2 introduced support for [VP8 but only when used in WebRTC.](https://webkit.org/blog/8672/) 

Couldn't come up with a better phrasing than this. Edits are welcomed.


Fyrd:
Thanks!

PR #4858, opened 2019-04-07
Chrome >=73 supports prefers-color-scheme

polyzen:
https://bugs.chromium.org/p/chromium/issues/detail?id=850098


polyzen:
The bug is to do with the theme of Chrome itself. Mistakenly thought it covered `prefers-color-scheme` as well.

PR #4859, opened 2019-04-08
Adds MediaRecorder support changes for Safari iOS and Safari TP

rgeraldporter:
MediaRecorder in the latest TP is no longer something that needs to be enabled, and the iOS Safari does have MediaRecorder behind a flag in 12.1+.


Fyrd:
Thanks!

PR #4866, opened 2019-04-10
Improved page about HTTP/2 support

Greg-21:
**In this commit I did the following:**
* updated link and changed link title
* changed "HTTP2" to "HTTP/2"
* removed doubled note.

### DIFF
--- a/features-json/http2.json
+++ b/features-json/http2.json
@@ -1,12 +1,12 @@
 {
   "title":"HTTP/2 protocol",
   "description":"Networking protocol for low-latency transport of content over the web. Originally started out from the SPDY protocol, now standardized as HTTP version 2.",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
+1 for being more precise across the board.

One note tho, probably addressed @Fyrd: "http2" should still find this entry. Maybe this is given by the feature name or by how the search works. Otherwise "http2" could be added as keyword?

Greg-21:
To be honest, I also don't know how exactly the search works. If it is necessary, I can add "http2" to the keywords.

Fyrd:
Search works on: feature ID, title, description & keywords. Since "http2" is the feature id that should  already take care of it in this case. :)

Thanks!

PR #4869, opened 2019-04-15
Corrected info about TLS 1.3 support in Samsung Internet 9

Greg-21:
In this commit I corrected the info about TLS 1.3 support in Samsung Internet 9.

Since Samsung Internet 9 is based on Chromium 67 and the version 67 doesn't support TLS 1.3 by default, we can safely assume that TLS 1.3 isn't supported in Samsung Internet 9.


Fyrd:
Ah, yeah I meant that to be a `n` when I updated it. Thanks!

PR #4871, opened 2019-04-15
Removed doubled note in support table concerning HTTP/2

Greg-21:
As the title says, in this commit I removed the doubled note in support table concerning HTTP/2.


Fyrd:
Sorry about those accidents, the data comparing code sometimes doesn't do very well when there's a large amount of changes. Made sure all your corrections got in now though. Thanks!

PR #4879, opened 2019-04-17
Update u2f support for Firefox

chicoxyzzy:
https://twitter.com/ciphercoffee/status/1113879931484655616

image


chicoxyzzy:
It turns out that Firefox 67 also has U2F support already

image

Fyrd:
Thanks!

PR #4881, opened 2019-04-18
Update CSS.supports information

erikgaal:
According to https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports#Browser_compatibility, the compatibility in Edge is not complete and should be noted. This PR adds the notes explaining the partial support in Edge.


HM100:
This also happens to some older versions of chrome and Firefox. Can you mention these browsers as well?

erikgaal:
Hi @HM100, do you happen to have a source for this information that I could refer to?

HM100:
It is on MDN as well. On the same page you linked

erikgaal:
Oh, thank you. I will update the PR.

Fyrd:
Confirmed, thanks!

PR #4883, opened 2019-04-20
[intrinsic-width] Firefox 66 is still affected by #1 (only width supported)

polarathene:
Verified on both my linux desktop with version 66, and my android phone with latest update as version 66.  `-moz-available` does not behave as expected, but is recognized overriding prior fallback.

I haven't actually tried with width, I can just confirm that it's still not working with height on version 66.


HM100:
Does min and max content work with height property on FX66?

HM100:
Strangely, [https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/66](FX66 changelog) mentions that you use can use intrinsic sizing on height property but only for `-min-content` and '-max-content`currently. Mind correcting this? While still being affected by note1, it is only for `-moz-fit-content` and `-moz-available`. Have a nice day!!! 😉

polarathene:
@HM100 I think that's what you wanted?

- I added a 5th note with the text you gave(should it link to the FF66 release?)
- Updated versions 66 and newer that are listed(afaik there hasn't been any changes for this support since 66) to note 5 and reverted flags from `y` to `a x` like prior versions.

polarathene:
Firefox Focus does not have it's own support target/section listed on caniuse or MDN afaik. I've just verified that the CSS support differs from the regular Firefox Android app when trying to support a workaround to my CSS-in-JS solution applying prefixes that I have no control over:

Raised an issue on the FF Focus repo: https://github.com/mozilla-mobile/focus-android/issues/4252

Fyrd:
Thanks!

PR #4892, opened 2019-04-23
Small improvements in support table concerning TLS 1.3

Greg-21:
**In this commit I did the following:**
* removed unnecessary note reference from Firefox versions 60 – 62 (its existence in this place was my mistake and it should never take place)
* moved note 1 into place 2 and vice versa
* improved a little bit description concerning Safari.


Fyrd:
Looks good, thanks!

PR #4894, opened 2019-04-24
Added Safari bug description for Beacon API

ozgkrc:
As described extensively in the [link](https://bugs.webkit.org/show_bug.cgi?id=188329): Desktop Safari 11.1 - 12 and iOS Safari 11.3 - 12.1 have bug with sendBeacon which throws error when it is used in a pagehide event callback. This issue has been fixed in iOS 12.2 and macOS 10.14.4.

This change proposes adding this bug information under bugs array in beacon.json


Fyrd:
Thanks!

PR #4897, opened 2019-04-25
Clarify IE limitation for blob urls

aharpervc:
- the blob url can still be referenced inside the iframed document, just not as the iframe src (apparently)


Fyrd:
Thanks!

PR #4900, opened 2019-04-29
Updated WebP support in Firefox Mobile

Eric-01:
Firefox Mobile supports WebP since version 65.0, like desktop Firefox.

Reference:
https://www.mozilla.org/en-US/firefox/android/65.0/releasenotes/


Fyrd:
Confirmed, thanks!

PR #4902, opened 2019-04-30
Fix typos in webvr notes by rewording

gildesmarais:
Reworded to remove product names, as more than one manufacturer of VR HMDs exist.


Fyrd:
Looks good, thanks!

PR #4908, opened 2019-05-04
Update Chrome support for prefers-color-scheme

marto55555:
Support for prefers-color-scheme has landed in Chrome Canary


Fyrd:
Thanks!

PR #4917, opened 2019-05-14
Replace IE10 / IE11 testing link

oatkiller:
The "not being supported" link which is found in the notes about IE10 / IE11 points to codepen. Codepen does not support IE11, therefore this link isn't useful to see what is supported by IE11. This commit replaces that link with one that points to glitch.com (I have copied the test script from codepen to glitch.) Glitch's 'view' feature loads the source as is, and will work in IE11. The glitch version can be used to see support in IE11.

# Before
image
and when accepting the warning:
image
And when trying to view it in 'full' mode:
image

# After
image

### DIFF
--- a/features-json/indexeddb.json
+++ b/features-json/indexeddb.json
@@ -365,7 +365,7 @@
   },
   "notes":"",
   "notes_by_num":{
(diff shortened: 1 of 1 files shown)


oatkiller:
Here is a link to the 'glitch' I made. Maybe that glitch should be owned by someone with authority over caniuse (instead of me)
https://glitch.com/~indexdb-support-test

Fyrd:
Thanks! I'm fine with your ownership though I appreciate the concern.

PR #4918, opened 2019-05-16
Fix speech-synthesis.json: Firefox for Android already supports this in v62

EbiEbiEvidence:
Here it is: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/62#APIs


Fyrd:
Can confirm support in Firefox for Android, but does not appear to work in the Android WebView.

EbiEbiEvidence:
@Fyrd That's right! I was wrong! I fixed it

EbiEbiEvidence:
Moreover, i cant find that opera supports it.
Opera only supports Speech Synthesis.
More at: https://blogs.opera.com/desktop/changelog-27/

Fyrd:
Thanks!

PR #4921, opened 2019-05-18
BUGFIX: Opera doesn't support SpeechRecognition

EbiEbiEvidence:
Opera supports only SpeechSynthesis.

https://help.opera.com/en/opera-version-history/#27
We can find `Synthesis` in this page, but we cannot find `Recognition`.
Moreover, I confirmed it today that SpeechRecognition doesn't works in latest Opera.


Fyrd:
Yeah, curiously it does seems to recognize the `SpeechRecognition` API and recognition can be "started" but none of the events seem to fire. Will add a note on that. Thanks!

PR #4927, opened 2019-05-22
Fix correct link to Mozilla/Bugzilla bug for :focus-visible

emilbjorklund:
The current link leads to a bug for missing _developer documentation_ on `:focus-visible`, and has been closed since MDN added docs.

This change updates the URL and tile to match the implementation bug instead.


Fyrd:
Thanks!

PR #4928, opened 2019-05-22
Update lazyload.json w/ link to the new standard

Malvoz:
The [legacy `lazyload`](https://caniuse.com/#feat=lazyload) attribute has poor implementations amongst browsers and I don't think there are any good reasons for developers to use it.

To avoid confusion I have updated the feature description with a link to the new [`loading`](https://caniuse.com/#feat=loading-lazy-attr) attribute.

Is it clear enough?...

### DIFF
--- a/features-json/lazyload.json
+++ b/features-json/lazyload.json
@@ -1,6 +1,6 @@
 {
   "title":"Resource Hints: Lazyload",
-  "description":"Gives a hint to the browser to lower the loading priority of a resource.",
+  "description":"Gives a hint to the browser to lower the loading priority of a resource. Please note that this is a legacy attribute, see the [`loading`](https://caniuse.com/#feat=loading-lazy-attr) attribute for the new standardized API.",
   "spec":"https://w3c.github.io/web-performance/specs/ResourcePriorities/Overview.html",
   "status":"unoff",
   "links":[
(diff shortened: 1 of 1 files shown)


features-json/lazyload.json
```diff
@@ -1,16 +1,12 @@
 {
   "title":"Resource Hints: Lazyload",
-  "description":"Gives a hint to the browser to lower the loading priority of a resource.",
+  "description":"Gives a hint to the browser to lower the loading priority of a resource. Please note that this is a legacy attribute, see the [`loading`](https://caniuse.com/#feat=loading-lazy-attr) attribute for the new standardized API.",
```
Malvoz: @Fyrd Since I [added a note](https://github.com/Fyrd/caniuse/pull/4928/files#diff-c5441aae512c0f4a638011b54ed30267R340), is it overkill to have a warning in the feature description aswell?

Fyrd: I'd suggest removing it from the notes and leaving it in the description, since folks are more likely to notice it there.

Malvoz: Removed note in  https://github.com/Fyrd/caniuse/pull/4928/commits/e68eb84b46e67ba87d8d32d075f7116b254c95f4. Thanks! :)

Fyrd:
Thanks!

PR #4931, opened 2019-05-23
Firefox 68: ::marker pseudo-element

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/68#CSS>:

> Support [`::marker`](https://developer.mozilla.org/en-US/docs/Web/CSS/::marker "The ::marker CSS pseudo-element selects the marker box of a list item, which typically contains a bullet or number.") pseudo-element ([bug 205202](https://bugzilla.mozilla.org/show_bug.cgi?id=205202 "FIXED: Support ::marker pseudo-element")).

Test using <https://developer.mozilla.org/en-US/docs/Web/CSS/::marker#Example> works:

image


Fyrd:
Thanks!

PR #4938, opened 2019-05-28
Safari `-webkit-background-clip: text`

guoyunhe:
All versions of Safari doesn't support `-webkit-background-clip: text` for buttons. But you can put a span inside button to achieve the same result. Codepen example https://codepen.io/guoyunhe/pen/byjOeN


Fyrd:
Thanks!

PR #4947, opened 2019-06-04
iOS 13 Safari adds support for prefers-color-scheme

fishcharlie:


fishcharlie:
Along with @Schweinepriester's comment above.

I tested this to verify it on one of my websites that uses `prefers-color-scheme`, and the Safari 13 beta release notes say the following:

> Added opt-in dark mode support for websites in Safari for iOS.

https://developer.apple.com/documentation/safari_release_notes/safari_13_beta_release_notes

Fyrd:
Thanks!

PR #4949, opened 2019-06-05
Mojave and later support TLS 1.3

daytonlowell:
Minor update on the note about Safari TLS 1.3 support being limited to Mojave and later.


Fyrd:
Thanks!

PR #4950, opened 2019-06-06
Firefox 69 will support Subgrid

jensimmons:
Best we can tell, Firefox will be shipping Subgrid in Firefox 69. It is currently in Firefox Nightly (without a flag).


jensimmons:
@Fyrd Do you also need an issue for this pull request? In the past, I've opened issues as well, and linked them to each pull-request. I think because otherwise it felt like the PR wasn't seen... but maybe that's not needed. I'm happy to do it. But also, hey, maybe that's overkill.

Fyrd:
@jensimmons Nah, no need for an issue. PRs are always seen and are generally merged within a few days provided no further info/changes are required. Thanks!

PR #4951, opened 2019-06-07
web-share updates

edent:
Tested on my phone. Now works on Android's Opera and Samsung.


dartess:
> Only the support value strings can be modified; additional versions cannot be added.

> Currently it is not possible to: Add a new browser or browser version

https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md

edent:
@dartess so what should I do? Wait for a new release with updated browser versions, or stick them on the most recent one? Happy to do whatever makes sense :-)

dartess:
@edent 
a) Do not modify `Opera Mobile`. It looks like not updated.
b) Check for work in the earliest available version of Samsung Internet 9.2 (9.2.0.x). To do this, you need to independently find a suitable APK and install it instead of the current version. After that, you can change `"9.2": "n #3"` => `"9.2": "y"` and run tests. But... But perhaps this is not enough.
c) If you see a discrepancy of support and want to make corrections, it is better to check other browsers (if possible). For example, a quick search in Google gave information that this api has been supported in Chrome Mobile for quite some time (https://www.chromestatus.com/features/5668769141620736). This information can also be recorded in the table (preferably, check before). Do not forget that the Samsung Browser is built on the basis of Chromium, and it is likely that in Samsung Browser this function will work in approximately the same version of Chromium as in the Chrome itself. This means that it is desirable to find the version of the Samsung browser where version 61 or newer of Chromium began to be used, find the appropriate APK and check the operation of this function there. For example, Samsung Browser 4 uses Chromium version 44. This information can be found in the database of user agents.

It looks complicated, but in my opinion the result is worth it.

And yes, everything voiced above is only my opinion and may not coincide with the opinion of the main developers of the project. I could say something extra or forget something. I just noticed an error in the request.

PR #4954, opened 2019-06-10
Firefox shipping Scroll Snap

jensimmons:
Firefox is shipping support for Scroll Snap in Firefox 68. And removing support for the older specification.

(BTW, the name has changed, and you might want to also change the name of the JSON file. "Snap points" is no longer what it's called. But I'll let you handle that yourself, separate from this pull request.)


Fyrd:
Thanks! I'll consider renaming the json file, unfortunately that would also break existing links to the feature so I prefer to avoid it.

PR #4960, opened 2019-06-14
Update backdrop-filter for MS Edge 75, 76

mbartlett21:
Backdrop-filter now works by default.

Do I have to do anything special to add MS Edge Version 76?
EDIT: It seems I can't add a new version.


Schweinepriester:
FYI, in case you didnt find that already:

<https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#unsupported-changes>

> 
> -   Add a new browser or browser version

---

Are you sure backdrop-filter will work in Edge 75? Isn't it coming with 76, like Chrom(e|ium)?
If the latter, I believe Fyrd is taking the Chrome data as basis for new Edge versions, thus this should appear automatically from our viewpoint.

Fyrd:
Version 75 has now been updated to 76 (with support) so closing this issue.

PR #4962, opened 2019-06-18
Update ios support for download attribute

rajsite:
The feature is available for testing in iOS 13 developer beta 1 according to https://bugs.webkit.org/show_bug.cgi?id=167341


Fyrd:
Thanks!

PR #4965, opened 2019-06-19
Object.values Edge support versions

radeno:
Based on official documentation https://docs.microsoft.com/en-us/microsoft-edge/dev-guide/whats-new/edgehtml-14


Fyrd:
Thanks!

PR #4966, opened 2019-06-19
Update promise-finally.json: Edge doesn't support

balexand:
See https://developer.microsoft.com/en-us/microsoft-edge/platform/status/promiseprototypefinally/

It's still in development.


Fyrd:
Works in my test. I suspect they just forgot to update it on that page.

PR #4968, opened 2019-06-20
Updated Flexbox support

jabcreations:
Added notes part five for Chrome's partial support.


Fyrd:
I don't believe this warrants partial support, `fieldset` elements aren't used that much so I don't think this issue affects too many developers. I'd suggest adding it in the bugs list instead.

jabcreations:
That is a flat out lame excuse. This is the second issue after the whole IE not supporting getBoundingClientRect() entirely. The information immediately available is not accurate and as a whole it is ultimately not trustworthy. You and others have diminished the value of caniuse as a reference tool by choice and therefore I can not justify attempting further attempts to contribute to this project.

PR #4969, opened 2019-06-21
Remove fixed bugs from css-media-interaction.json

cletusw:
https://bugzilla.mozilla.org/show_bug.cgi?id=1035774 fixed in 2018-09-13 for Firefox 64
https://bugs.chromium.org/p/chromium/issues/detail?id=442418 fixed in July 2017


Fyrd:
Thanks!

PR #4971, opened 2019-06-22
Add resources on Safari’s Beacon API problems

da2x:


Fyrd:
I think including the same link in both bugs as well as links is unnecessary, can you remove the one under "links"? Thanks!

da2x:
> can you remove the one under "links"?

Sure, it’s done.

Fyrd:
Thanks!

PR #4973, opened 2019-06-24
ResizeObserver: Add additional polyfill for latest specification.

TremayneChrist:
As the original polyfill does not support the latest version of the specification and doesn't seem to be maintained as much recently. I'd like to propose this additional polyfill, which adds support for `box-size` options.


Fyrd:
Thanks!

PR #4979, opened 2019-06-26
Safari TP 86: tab-size with units

Schweinepriester:
<https://webkit.org/blog/9375/release-notes-for-safari-technology-preview-86/>:

> Implemented `tab-size` with units ([r246193](https://trac.webkit.org/changeset/246193/webkit/))

I also tested it using <https://tests.caniuse.com/?feat=css3-tabsize&ua=saf12.1>:

image


Fyrd:
Thanks!

PR #4984, opened 2019-07-01
Firefox 68: BigInt enabled by default

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/68>:

> The new [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt "BigInt is a built-in object that provides a way to represent whole numbers larger than 253, which is the largest number JavaScript can reliably represent with the Number primitive. BigInt can be used for arbitrarily large integers.") primitive is enabled by default ([bug 1527902](https://bugzilla.mozilla.org/show_bug.cgi?id=1527902 "FIXED: Enable javascript.options.bigint by default")).

(Using Firefox 68.0b14 (64-Bit) shows <https://tests.caniuse.com/?feat=bigint> green as well.)


Fyrd:
Thanks!

PR #4993, opened 2019-07-07
Tweak SVG `spec` urls and `status`

takenspc:
At Can I use, some entries reference SVG 2 and others reference SVG 1.1. This may confuse readers.

This PR makes entries references SVG 2 as much as possible though number or entries which references SVG 1.1 increased.


Fyrd:
Looks good, thanks!

PR #4994, opened 2019-07-08
Updates notes for Network Information API

coliff:
RE: https://caniuse.com/#feat=netinfo

> In Chrome the API is enabled on Android only

This is a bit misleading as the API is available on desktop too and supports the **downlink**, **effectiveType** & **rtt** values (as point 4 in the notes mentions). You can see the output from the API here here: https://googlechrome.github.io/samples/network-information/. I think it's better to remove this statement.

> Support is also available on Firefox OS.

No longer relevant, so safe to remove.

### DIFF
--- a/features-json/netinfo.json
+++ b/features-json/netinfo.json
@@ -344,7 +344,7 @@
       "2.5":"a #2"
     }
   },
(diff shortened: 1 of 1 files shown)


Fyrd:
Good reasoning, I agree. Thanks!

PR #5002, opened 2019-07-16
Safari 10.3 doesn't fully support async functions

paulirish:
A few months ago, it was discovered that async arrow functions aren't supported in Safari 10.1 to 10.3: https://github.com/kangax/compat-table/pull/1420


Fyrd:
Thanks!

PR #5003, opened 2019-07-17
inherit issue with `<details>` in Chrome

MartijnCuppens:
This is not necessarily an issue with `box-sizing` but it is an issue which occurs when this snippet is used:

```css
html {
  box-sizing: border-box;
}

*,
::before,
::after {
  box-sizing: inherit;
}

```

Because that technique is used quite a lot, it might be worth mentioning this in the bugs section?


Fyrd:
Thanks!

PR #5012, opened 2019-07-23
Update UC Browser in keyboardevent-key.json

cletusw:
Tested via BrowserStack on Samsung Galaxy S7, UC Browser 12.10 with https://jsbin.com/jiyazimoha/1


HM100:
Invalid PR since it adds new browser version

Fyrd:
Just tested it myself with the latest version of UC Browser and it fails in my test:
https://tests.caniuse.com/?feat=keyboardevent-key
And yes, can't add a version, though I'm about to do an update.

PR #5021, opened 2019-08-01
Update meta-theme-color.json

timvango:
Theme Color supported partially in Chrome: PWA standalone mode. Added note for this.


Fyrd:
Thanks!

PR #5023, opened 2019-08-05
Lazy-load: new blog post, keywords and Chrome flag

Malvoz:
Adding link to new informational article: https://web.dev/native-lazy-loading
&plus; `#enable-lazy-frame-loading` flag and some relevant keywords.


Fyrd:
Thanks!

Issue #4963, opened 2019-06-18
inputmode attribute update


coliff:
RE: https://caniuse.com/#feat=input-inputmode

This page needs some updates:

- inputmode is fully supported on both Opera Mobile and Opera Mini for Android. I don't know the version number that support was added though. (presumably a similar sort of timeframe as when support was added to Chrome for Android)
- I suggest to remove the note '_Firefox OS is reported to have partial support for this property as x-inputmode._' as Firefox OS is no longer relevant
-  I suggest to remove the note '_Blink-based browsers appear to recognize the inputMode property on input & textarea fields though it is unclear what effect it currently has as the feature is still in development._' as it is no longer true.
- I suggest to remove the note '_MS Edge status: Under Consideration_' as it is no longer relevant. Edge is moving to Chromium and will be supported then.
- I think it'd be good to link to my article on CSS-Tricks for more detailed info and screenshots: https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/

PR #5028, opened 2019-08-07
Update input-inputmode.json

coliff:
This fixes: #4963

- adds support for Opera Mobile and Opera Mini
- remove the note 'Firefox OS is reported to have partial support for this property as x-inputmode.' as Firefox OS is no longer relevant
- remove the note 'Blink-based browsers appear to recognize the inputMode property on input & textarea fields though it is unclear what effect it currently has as the feature is still in development.' as it is no longer true.
- remove the note 'MS Edge status: Under Consideration' as it is no longer relevant. Edge is moving to Chromium and will be supported then.
- replace the very old wufoo article with newer up-to-date one from CSS tricks


Fyrd:
Looks good, except it doesn't seem to be support in Opera Mini. Note that "Opera Mini" refers to the feature phone version, not the default behavior of Opera Mini on Android/iOS (need to enable "Extreme data savings" mode there to test it).

coliff:
Hi @Fyrd - thanks for checking on this. I updated the PR to change Opera Mini support back to 'n'

Fyrd:
Thanks!

PR #5031, opened 2019-08-13
Firefox 69: `user-select` unprefixed

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/69>:

> The [`user-select`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select "The user-select CSS property controls whether the user can select text. This doesn't have any effect on content loaded as chrome, except in textboxes.") property---which specifies whether or not the user is able to select text in the affected element---has been unprefixed in Firefox 69 ([bug 1492739](https://bugzilla.mozilla.org/show_bug.cgi?id=1492739 "FIXED: Consider unprefixing -moz-user-select")).

I've checked https://tests.caniuse.com/?feat=user-select-none using Firefox `69.0b13 (64-Bit)`.


Fyrd:
Thanks!

PR #5032, opened 2019-08-14
Update KeyboardEvent.code with mobile support

Schweinepriester:
Based on <https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Browser_compatibility> or more precise the individual commits here <https://github.com/mdn/browser-compat-data/blame/65740c63cb8c861f4511f29b3fa3cb75365f6878/api/KeyboardEvent.json#L244>.

While I'm fairly confident that this is correct even for the versions on caniuse, which are not the same as they checked, I didn't test it 0:-)


Fyrd:
From my own tests (using the onscreen virtual keyboard) it still seems to fail...just tried it with https://codepen.io/denilsonsa/pen/epmoma and it just returns an empty string. According to the specification at https://w3c.github.io/uievents/#code-virtual-keyboards :

> In general, if a virtual (or remote control) keyboard is mimicking the layout and functionality of a standard keyboard, then it MUST also set the code attribute as appropriate" which doesn't seem to be happening.

I haven't tried using a physical keyboard with these Android browsers though, I guess if that works we should go for partial support.

Schweinepriester:
Hrmpf, I just confirmed it using the android emulator and the onscreen keyboard as well as the physical keyboard connected to my PC:

<https://streamable.com/2thil> (alternative https://i.imgur.com/ebdOPdR.gifv)

> I haven't tried using a physical keyboard with these Android browsers though, I guess if that works we should go for partial support.

Seems to work, at least with my setup. But I dont think we _should_ go for partial support, we _could_, but that might give the wrong impression… thus closing.

PR #5040, opened 2019-08-18
Update cautionary note on CSS line-clamp

FND:
now fairly well supported across modern browsers, so the phrasing seems outdated


Fyrd:
Agreed, thanks!

PR #5046, opened 2019-08-21
IE11 doesn't upport const in for-in and for-of loops

DominikPalo:
As you can see on http://kangax.github.io/compat-table/es6/#test-RegExp.prototype_properties, IE11 doesn't support using `const` in `for-in` and `for-of` loops (replicated on my IE11 - it fails with a script error)


Fyrd:
Thanks!

PR #5056, opened 2019-08-23
Correct Android Chrome support (Fix #5042)

### LINKED ISSUE #5042: Chrome for Android - wrong values
According to this https://caniuse.com/#feat=dragndrop, Chrome for Android supports drag-and-drop.

However, when I check these demos in Chrome for Android, they do not work: https://www.w3schools.com/html/html5_draganddrop.asp

How exactly do you achieve these values? I think many of them may be incorrect.

*** It seems like it is not supported on iOS Safari either.

Malvoz:
Fix #5042.


Malvoz:
Closing this for now, it _seems_ that drag and drop is or has been supported in Android, see resolution in https://github.com/Fyrd/caniuse/issues/5042.

PR #5060, opened 2019-08-26
Firefox 69: CSS `contain`

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/69>:

> The [`contain`](https://developer.mozilla.org/en-US/docs/Web/CSS/contain "The contain CSS property allows an author to indicate that an element and its contents are, as much as possible, independent of the rest of the document tree. This allows the browser to recalculate layout, style, paint, size, or any combination of them for a limited area of the DOM and not the entire page.") property allowing to define that an element and its contents are mostly independent of the rest of the DOM tree got implemented [bug 1487493](https://bugzilla.mozilla.org/show_bug.cgi?id=1487493 "FIXED: Enable pref for CSS Containment (layout.css.contain.enabled)").

<https://tests.caniuse.com/?feat=css-containment> using `69.0b16 (64-Bit)`:

image


Fyrd:
Confirmed, thanks!

Issue #3948, opened 2018-01-02
Renaming WebVR to WebXR


Utopiah:
cf https://github.com/immersive-web/webxr

> The WebXR Device API is intended to completely replace the legacy WebVR specification when finalized. In the meantime multiple browsers will continue to expose the older API.

Utopiah:
@Fyrd anything I could help with here?

PerspectivesLab:
is webxr officially included as release? or is it still in beta test for google canary

Utopiah:
So I updated the page accordingly but it's not actually not really correct. Most browsers listed there support WebVR but not WebXR yet... (yes, just criticizing my own changes here).

PR #5064, opened 2019-09-01
feat: Split WebXR and WebVR

ExE-Boss:
**Chrome 79** has removed support for **WebVR**.

Without this, it’s impossible to display this in the “Can I use…” data.

---

Reverts: #4456
See also: #3948

/cc @jpmedley

### DIFF
--- a/features-json/webxr.json
+++ b/features-json/webxr.json
@@ -1,33 +1,17 @@
 {
-  "title":"WebVR API",
-  "description":"API for accessing virtual reality (VR) devices, including sensors and head-mounted displays.",
-  "spec":"https://w3c.github.io/webvr/",
-  "status":"unoff",
+  "title":"WebXR Device API",
+  "description":"API for accessing virtual reality (VR) and augmented reality (AR) devices, including sensors and head-mounted displays.",
+  "spec":"https://immersive-web.github.io/webxr/",
+  "status":"other",
   "links":[
     {
-      "url":"https://webvr.rocks/",
-      "title":"Detailed device support information"
-    },
-    {
-      "url":"https://github.com/googlevr/webvr-polyfill",
-      "title":"WebVR polyfill"
-    },
-    {
-      "url":"http://aframe.io",
-      "title":"WebVR framework"
-    },
-    {
-      "url":"https://webvr.info/",
-      "title":"WebVR info"
-    },
-    {
-      "url":"https://developer.mozilla.org/en-US/docs/Web/API/WebVR_API",
-      "title":"MDN Web Docs - WebVR API"
+      "url":"https://developer.mozilla.org/docs/Web/API/WebXR_Device_API",
+      "title":"MDN Web Docs - WebXR Device API"
     }
   ],
   "bugs":[
     {
-      "description":"WebVR is in development and is still changing. Because of this there could be bugs inside the standard of the W3C."
+      "description":"WebXR is in development and is still changing. WebXR was formally developed as WebVR. Because of this there could be bugs inside the standard of the W3C."
     }
   ],
   "categories":[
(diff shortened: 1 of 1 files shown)


Fyrd:
Yeah, seems like the right thing to do. Thanks!

Fyrd:
Looks like XR data needs some updating... the numbered notes should probably be removed and I'm not seeing support in Chrome 76.

features-json/webxr.json
```diff
@@ -0,0 +1,368 @@
...
+      "70":"n",
+      "71":"n",
+      "72":"n",
+      "73":"n",
+      "74":"n",
+      "75":"n",
+      "76":"y",
+      "77":"y",
+      "78":"y",
+      "79":"y"
```
jpmedley: Anything before 79 should be `"n"`. The same would apply for Android Chrome and Android WebView. (I don't know how you do those.)

features-json/webxr.json
```diff
@@ -0,0 +1,368 @@
+{
+  "title":"WebXR Device API",
+  "description":"API for accessing virtual reality (VR) and augmented reality (AR) devices, including sensors and head-mounted displays.",
```
jpmedley: This definition ignores AR use cases which have applications outside of games. The spec was designed specifically (so it's author tells me) to address a spectrum of immersion so that a site could use as little or as much immersion as it wanted. See [my article](https://developers.google.com/web/updates/2018/05/welcome-to-immersive) for more on that.

features-json/webxr.json
```diff
@@ -0,0 +1,368 @@
...
+  "status":"other",
+  "links":[
+    {
+      "url":"https://developer.mozilla.org/docs/Web/API/WebXR_Device_API",
+      "title":"MDN Web Docs - WebXR Device API"
+    }
+  ],
+  "bugs":[
+    {
+      "description":"WebXR is in development and is still changing. WebXR was formally developed as WebVR. Because of this there could be bugs inside the standard of the W3C."
```
jpmedley: >WebXR was formally developed as WebVR.

There's some historical truth to this. The spec has evolved so much in more than two years that I'm not sure it's relevant to users of the API. Neither Google nor MDN feel the need to mention this in WebXR documentation. 

>Because of this there could be bugs inside the standard of the W3C.

What does this mean?

features-json/webxr.json
```diff
@@ -0,0 +1,368 @@
...
+      "1.2":"n"
+    },
+    "baidu":{
+      "7.12":"n"
+    },
+    "kaios":{
+      "2.5":"n"
+    }
+  },
+  "notes":"For a WebXR experience a head mounted display (VR HMD) is required.",
```
jpmedley: Not true. Look at [`XR.requestSession()` and `XR.supportsSession()`](https://immersive-web.github.io/webxr/#xr-interface). The API supports a session mode called `"inline"` with which content can be rendered in a page.

features-json/webxr.json
ExE-Boss: https://www.chromestatus.com/feature/5680169905815552 says 76, so… ¯\\\_(ツ)\_/¯

ExE-Boss:
@jpmedley To all your questions:

I just copied the `webvr.json` file from #4456, before I reverted that PR, so take it up with @Utopiah

features-json/webxr.json
jpmedley: It says 'origin trial'. That means that individual sites would require a token to activate the API. Origin trial tokens will be unavailable when the API ships in 79.

Fyrd:
I've updated the feature data with help from @jpmedley's comments and made the feature available on the site now. For any further suggested changes please submit a new PR. Thanks all!

PR #5069, opened 2019-09-04
macOS Safari is capable of ChaCha20-Poly1305, too

Schweinepriester:
Bildschirmfoto 2019-09-05 um 00 47 14

Note sure why this is missing from <https://caniuse.com/#feat=chacha20-poly1305>? Probably is available since Safari 11, just like iOS 11, but maybe just for specific macOS versions?

But since i dont have the means to test that, i only changed what i could test (assuming 13 inbetween will work as well) :)


Fyrd:
Not sure either, but yes, confirmed! Thanks.

PR #5070, opened 2019-09-05
Chrome for Android supports prefers-reduced-motion

SamWarmuth:
Chrome for Android 74+ fully supports CSS media query `prefers-reduced-motion`. In the current `prefers-reduced-motion.json`, Android for Chrome 75 is incorrectly marked as `n`.

Here's the Chrome Status page: https://www.chromestatus.com/feature/5597964353404928. "Enabled by default in: [...] Chrome for Android release 74".

The only version in `prefers-reduced-motion.json` is `and_chr` 75, and the contributing guidelines say "Currently it is not possible to: [...] Add a new browser or browser version", so I updated 75 (rather than adding 74).


Fyrd:
Thanks!

PR #5076, opened 2019-09-11
added link for WebShare API

docluv:
Added a link to an article that demonstrates a couple of practical applications of the WebShare API to use as a reference.


Fyrd:
Thanks!

PR #5080, opened 2019-09-12
note that background-filter exists behind a flag in FF >= 70

gabrielgrant:
Feature implementation issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1178765

Issue to activate feature by default: https://bugzilla.mozilla.org/show_bug.cgi?id=1578503

Also noted at https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter#Browser_compatibility


Fyrd:
Thanks!

PR #5082, opened 2019-09-14
Fix link in shadowdom.json

Majr-:


Fyrd:
Wow, can't believe that was wrong. Thanks for catching & fixing!

PR #5088, opened 2019-09-16
Added bugs for iOS13 PointerEvents. 

IanBellomy:
These issues have been reported and are being investigated but there is no date for fixes so I'm guessing they'll be in there when iOS13 releases Sep. 19th.


Fyrd:
Thanks!

kiding:
Both bugs are fixed in the WebKit source code. The next release should include the patches.

IanBellomy:
@kiding Any idea when that'll be?

kiding:
For that, I'm clueless as you. iOS 13.1.2 has been released, but I wouldn't give much hope as [the tag isn't even available on WebKit.](https://trac.webkit.org/browser/webkit/releases/Apple) I can only presume nothing has changed since 13.1.

kiding:
For that, I'm clueless as you. iOS 13.1.2 has been released, but I wouldn't give much hope as [the tag isn't even available on WebKit.](https://trac.webkit.org/browser/webkit/releases/Apple) I can only presume nothing has changed since 13.1. At least one of the patches made it to [Safari 608.3.4 just last week.](https://trac.webkit.org/export/250540/webkit/tags/Safari-608.3.4/LayoutTests/ChangeLog)

PR #5089, opened 2019-09-16
Lazyload supported in Chrome 77

laukstein:
Announcement https://developers.google.com/web/updates/2019/09/nic77
Demo with images https://mathiasbynens.be/demo/img-loading-lazy


Fyrd:
This file's data is for the `lazyload` attribute, only supported by IE & Edge. The support you're referring to is for the `loading` attribute, which is covered at https://caniuse.com/#feat=loading-lazy-attr (and already notes Chrome's support)

PR #5090, opened 2019-09-17
Revert "Add link"

webb04:
Reverts Fyrd/caniuse#5086.

The markdown doesn't like the double bracket: `#setForceDark(int))`


Fyrd:
No need, have fixed this manually by replacing the inner parens with %28 %29.

PR #5098, opened 2019-09-20
Web Animations still experimental in Safari 13

stevenvachon:


Fyrd:
Confirmed, thanks!

stevenvachon:
Safari 13 is also the current release, but is not displayed as such on the site. I'm not sure how to correct this.

PR #5100, opened 2019-09-23
Change ios_saf 11.0-11.2 to n for Promise.prototype.finally

Andy2K11:
https://bugzilla.mozilla.org/show_bug.cgi?id=1467561
This bug reports suggests that Promise.prototype.finally()
 wasn't implemented on iOS Safari until 11.3 which concurs with my testing on iOS Safari 11.1


Fyrd:
Confirmed, thanks!

PR #5108, opened 2019-10-01
Add correct support for native lazyload on Chrome for Android 76

alexmaurizio:
As reported in the feature announcement https://www.chromestatus.com/feature/5645767347798016
and after testing the functionality, correctly report it as active for this browser


Fyrd:
Confirmed, thanks!

PR #4941, opened 2019-05-29
Safari TP 83: WebAuthN by default & `display: flow-root`

Schweinepriester:
<https://webkit.org/blog/8967/release-notes-for-safari-technology-preview-83/>:

> -   Enabled WebAuthN by default on macOS ([r245589](https://trac.webkit.org/changeset/245589/webkit/))

> -   Implemented a modern "clearfix" with `display: flow-root` ([r245494](https://trac.webkit.org/changeset/245494/webkit/))

PR #5117, opened 2019-10-04
Updates Safari support for U2F

gboone:
Safari 13 for MacOS "added support for FIDO2-compliant USB security keys with the Web Authentication standard" https://developer.apple.com/documentation/safari_release_notes/safari_13_release_notes


kiding:
It looks to me the release note refers to #4941. As far as I know, FIDO U2F API, [`u2f.sign`](https://fidoalliance.org/specs/fido-u2f-v1.0-nfc-bt-amendment-20150514/fido-u2f-javascript-api.html#high-level-javascript-api) is not supported by Safari.

Fyrd:
Agree with @kiding.

PR #5121, opened 2019-10-06
U2F: Note about `CryptoTokenExtension` on Chrome-based browsers

kiding:
On Firefox, `window.u2f` is always available, implemented using native codes. On the other hand, on Chrome-based browsers, a FIDO client is implemented in an internal extension called `CryptoTokenExtension`. [High-level JS API `window.u2f`](https://fidoalliance.org/specs/fido-u2f-v1.0-nfc-bt-amendment-20150514/fido-u2f-javascript-api.html#high-level-javascript-api) is only available after [a JS wrapper for the extension](https://github.com/google/u2f-ref-code/blob/master/u2f-gae-demo/war/js/u2f-api.js) is executed.


Fyrd:
Thanks!

PR #5126, opened 2019-10-10
Subgrid is shipping in Firefox 71

jensimmons:
Subgrid has not shipped in Firefox yet. It's shipping in ff71, on Dec 3. 
Updating the data to reflect that reality.
(We came very close to shipping in 69 & 70, but held it for polish.)


Fyrd:
Thanks!

PR #5139, opened 2019-10-17
Add a polyfill for the download attribute in IE11

jelmerdemaat:
This PR adds a polyfill for the a[download] functionality for IE11. More info:

https://github.com/jelmerdemaat/dwnld-attr-polyfill


Fyrd:
Thanks!

PR #5140, opened 2019-10-17
Update css-placeholder.json

9brada6:
Internet explorer supports ::placeholder CSS with the vendor prefix ":-ms-input-placeholder", with a single ":". 
For example see "Try it!" from https://www.w3schools.com/cssref/sel_placeholder.asp


Fyrd:
That matches the `placeholder-shown` pseudo-class behavior which works a little differently and is tracked at https://caniuse.com/#feat=css-placeholder-shown (note that IE support is shown there)

PR #5141, opened 2019-10-18
SameSite: Chrome 80+ changes, Safari support statuses

kiding:
This rabbit hole has kept me quite occupied for several weeks now. The changes are:
- Chrome 80 will follow [Incrementally Better Cookie](https://tools.ietf.org/html/draft-west-cookie-incrementalism-00) by default
- Changed many of the Safari support statuses to partial
  - Refer to the commit messages for the detailed rationale
  - Should be revisited when a new macOS version is released
- Added more Safari bugs and useful links


Fyrd:
Great work, thanks!

PR #5146, opened 2019-10-23
Node.innerText => HTMLElement.innerText

nwtgck:
Hi developers!

I'm very new to contribute to caniuse. I modified `Node.innerText` to `HTMLElement.innerText` because the following reason. According to <https://html.spec.whatwg.org>, `HTMLElement` has `innerText` not `Node`.

## `interface HTMLElement`
image

- `interface HTMLElement`: <https://html.spec.whatwg.org/multipage/dom.html#htmlelement>)
- `interface Node`: <https://dom.spec.whatwg.org/#interface-node>

In MDN, <https://developer.mozilla.org/en-US/docs/Web/API/Node/innerText> redirects to <https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText> too.

### DIFF
--- a/features-json/innertext.json
+++ b/features-json/innertext.json
@@ -1,11 +1,11 @@
 {
-  "title":"Node.innerText",
+  "title":"HTMLElement.innerText",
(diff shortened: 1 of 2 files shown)


Fyrd:
Thanks!

PR #5148, opened 2019-10-23
WebXR is enabled by default in Chrome 79

antonjuulnaber:
WebXR is now enabled by default in Chrome:
https://www.chromestatus.com/feature/5680169905815552

The API is only partially supported. Much of its functionality is still under development:
https://docs.google.com/document/d/1R8Bz0_vK9YmY5-ZhsbLtZ9-DDJpTOgNeibFD4SdUnbw/edit


features-json/webxr.json
```diff
@@ -350,9 +350,10 @@
   },
   "notes":"",
   "notes_by_num":{
-    "1":"Chrome 66-72 supported a very early version of the spec",
-    "2":"Part of an origin trial",
-    "3":"Part of the spec can be enabled with the [webxr](chrome://flags/#webxr) and [webxr-hit-test](chrome://flags/#webxr-hit-test) flags."
+    "1":"Chrome 66-72 supported a very early version of the spec.",
+    "2":"Part of an origin trial.",
+    "3":"Part of the spec can be enabled with the [webxr](chrome://flags/#webxr) and [webxr-hit-test](chrome://flags/#webxr-hit-test) flags.",
```
jpmedley: These flags have been gone for a while. I know they're not in Chrome 78 and I have no way to check earlier versions.

features-json/webxr.json
```diff
@@ -350,9 +350,10 @@
   },
   "notes":"",
   "notes_by_num":{
-    "1":"Chrome 66-72 supported a very early version of the spec",
-    "2":"Part of an origin trial",
-    "3":"Part of the spec can be enabled with the [webxr](chrome://flags/#webxr) and [webxr-hit-test](chrome://flags/#webxr-hit-test) flags."
+    "1":"Chrome 66-72 supported a very early version of the spec.",
```
jpmedley: These versions are not usable without flags, which are only intended for developers who want to try the thing out. Since the API will soon be enabled by default, the early version isn't anything anyone should be using.

features-json/webxr.json
```diff
@@ -350,9 +350,10 @@
...
+    "2":"Part of an origin trial.",
```
jpmedley: I wouldn't mention the origin trial either. Chrome 78 is the last version for it. Participating in an OT requires a sign-up and an origin-specific token. [All tokens will be deactivated on December 3.](https://developers.chrome.com/origintrials/#/view_trial/-5670031930859454463) After that date there will be no way to use XR in versions 76, 77, or 78.

features-json/webxr.json
jpmedley: The recommendation is to remove the flags completely. Users shouldn't be touching them. The implementation differences between 66-77 and 79 versions of the spec are enough that developers should build for 79 or later.

features-json/webxr.json
jpmedley: Again, this does not help developers build for the future. I recommend removing it.

features-json/webxr.json
antonjuulnaber: Replying to all comments here. I also think this should be removed eventually, but I think you're jumping the gun on it.
My team has been developing solutions for WebXR for a while now, and keeping track of what code runs where and what spec it uses is a nightmare. This resource helps us a lot.
It will take some time for us to update our code to the stable spec, and meanwhile we will still be using the old code.
I imagine other developers are facing the same problem, and i don't think keeping the data will have any negative effect on new development. Anyone new to the game will build for 79+.

antonjuulnaber: The flags in note 3 are still present all the way up to v78. They are also present up to v80, though that might change.

jpmedley: I'm running version 78 and they're not in my installation. I just double checked.

jpmedley: These have already been removed on mac.

antonjuulnaber: Ah, that makes sense! I'm on Windows. Browserstack shows they were removed in Chrome for Mac v77.
There are other slight inconsistencies with this note as well, maybe this could work:
```suggestion
    "3":"Chrome 66-78 has had intermittent support behind the #webxr and other #webxr-* flags.",
```
I'd say it's good enough for something that will be removed in a few months?

jpmedley: Well let's say I'm not going to argue the point any further. 

This is an example of why I'm never thrilled about this information being in compatibility data. As you can see from this, once a feature ships in stable, we're rather sloppy about when and how we remove the flags. That's a reflection of the fact that we don't intend them for production use.

Fyrd:
I agree with @jpmedley that we should start losing notes on the details of flagged support in older versions. Such information should only be of use to curious developers who will be using the current or upcoming versions of a browser. There may be a good place to keep this information for historical reasons but for most users they'll just be noise and distract from more valuable information.

antonjuulnaber:
Allright, if you say so, our team will just rely on our own record, and they can be removed here. I still believe it would be better to wait a few months.

Fyrd:
Well, when I said older versions I meant versions before the current one, so I feel 78 should still include the notes. Also I would recommend the older versions appear as `n d` so it's still clear at which point Chrome started having support behind a flag (though the details are less important). 
If you're getting tired of updating this PR though let me know and I'll take care of it, I do appreciate the research and diligence you put into this. :)

antonjuulnaber:
Arh! allright :)

Fyrd:
Excellent, thank you!

PR #5149, opened 2019-10-23
Improve WebGL 2.0 data

antonjuulnaber:
Fixes a few incosistencies

- ~~Firefox added support in 52, not in 51~~ _data is mixed on this, not changing anyway_
- The Android browser does have support


features-json/webgl2.json
```diff
@@ -91,7 +91,7 @@
       "48":"n d #1 #5",
       "49":"n d #1 #5",
       "50":"n d #1 #5",
-      "51":"y",
+      "51":"n d #1 #5",
```
Greg-21: Firefox Release Notes website claims that support for WebGL 2 was added in Firefox 51 (not in 52). Also other sources like e.g. [MDN's WebGL 2 browser compatibility data](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext) and [WebGL article on Wikipedia](https://en.wikipedia.org/wiki/WebGL) claims the same.

Greg-21: Firefox Release Notes website claims that support for WebGL 2 without any additional tricks was added in Firefox 51 (not in 52). Also other sources like e.g. [MDN's WebGL 2 browser compatibility data](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext) and [WebGL article on Wikipedia](https://en.wikipedia.org/wiki/WebGL) claims the same.

antonjuulnaber: Huh, that's wierd. You're right about your sources, but if you go to [the mozilla platform status page](https://platform-status.mozilla.org/) and search for "WebGL", it will say it landed in 52 🤔

I guess 51 is probably right...

Greg-21: Hmm, it really looks strange. I have just found [this article on Mozilla Hacks](https://hacks.mozilla.org/2017/01/webgl-2-lands-in-firefox/), which I remember I read at that time, and it also claims "Firefox 51", so I think it's simply a mistake on Firefox Platform Status website.

Greg-21: Hmm, it really looks strange. I have just found [this article on Mozilla Hacks](https://hacks.mozilla.org/2017/01/webgl-2-lands-in-firefox/), which I remember I read at that time, and it also claims "Firefox 51". So I suppose it's simply a mistake on Firefox Platform Status website.

Fyrd:
Thanks!

PR #5152, opened 2019-10-25
Update text-decoration.json

mattbrundage:
``text-decoration-skip`` has been supplanted by ``text-decoration-skip-ink``. 
Firefox 70: https://bugzilla.mozilla.org/show_bug.cgi?id=1573631
Chromium 64: https://www.chromestatus.com/feature/5631679087509504


Fyrd:
The latest spec still has a `text-decoration-skip` property for other values:
https://drafts.csswg.org/css-text-decor-4/#text-decoration-skip-property
So I think the partial support should still be used.

mattbrundage:
@Fyrd , this is true. But I have a feeling that no browser will get around to supporting all of the possible values of [text-decoration-skip](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip) and that it will eventually be removed. MDN's documentation states that "Most other browsers are converging on supporting the simpler ``text-decoration-skip-ink`` property."

I think the reason that this feature is so messy right now is that [line-decoration](https://www.w3.org/TR/css-text-decor-3/#line-decoration) is an established Candidate Recommendation (CR), while [the newer ``skip`` properties](https://drafts.csswg.org/css-text-decor-4/#text-decoration-skip-ink-property) are still in Working Draft (WD). Perhaps we should split this feature into multiple parts or have sub-features like we do with HTML5 forms?

mattbrundage:
This pull request is now too out-of-date to be useful.

For this feature, I still maintain that we shouldn't indicate partial support for browsers that don't support CSS Text Decoration Module Level 4, which is still in Working Draft and still considered "experimental" by MDN. Rather, for the time being, this feature should concern itself with CSS Text Decoration Module Level 3, which is a Candidate Recommendation.

PR #5155, opened 2019-10-26
Corrected info about TLS 1.3 support in Samsung Internet 10

Greg-21:
As written in the title, in this commit I corrected the info about TLS 1.3 support in Samsung Internet 10.

I haven't tested it, but since Samsung Internet 10 is based on Chromium 71 that supports TLS 1.3 by default, we can assume that TLS 1.3 is supported also in Samsung Internet 10. I don't think they changed the default "setting" and disabled support for such important web protocol in their browser.

### DIFF
--- a/features-json/tls1-3.json
+++ b/features-json/tls1-3.json
@@ -346,7 +346,7 @@
       "7.2-7.4":"n",
       "8.2":"n",
       "9.2":"n",
-      "10.1":"n"
+      "10.1":"y"
     },
     "and_qq":{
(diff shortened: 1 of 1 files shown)


Fyrd:
Just tested Samsung Internet 10.1 at https://www.ssllabs.com/ssltest/viewMyClient.html and it shows no support for TLS 1.3. Best not to assume when it can be tested. :)

Greg-21:
Yeah, you're probably right.

PR #5180, opened 2019-11-17
Update Web Components V0

antonjuulnaber:
The old spec for `web components` is dying fast. I have made some changes to reflect that:
- Removed support from Chrome 80, 81
- Removed all comments about old flags from very old browser versions

I also added a missing Chrome ID, and updated the description for `html imports` to reflect that it is deprecated.

https://www.chromestatus.com/feature/4642138092470272
https://www.chromestatus.com/feature/4507242028072960
https://www.chromestatus.com/feature/5144752345317376


Fyrd:
Looks good, thanks!

PR #5181, opened 2019-11-19
Chrome adds support for SVG favicons

antonjuulnaber:
Chrome is adding support for SVG favicons in v80 - yay!

I've also added the corresponding chrome ID and added a comment about how Chrome is treating the image. I'm not sure if the comment is necessary, so i've made it a separate commit for you to easily discard if you so please ;)

https://www.chromestatus.com/feature/5180316371124224

_Update: Now also fixes incorrect Opera data_


fsoder:
The contents of "Note 4" ("...requires the served MIME type to be image/svg+xml") also apply to Chrome (and other Chromium-based browsers, like Opera [for version 67+]). And thanks for updating the entry =).

antonjuulnaber:
> The contents of "Note 4" ("...requires the served MIME type to be image/svg+xml") also apply to Chrome (and other Chromium-based browsers, like Opera [for version 67+]). And thanks for updating the entry =).

Thanks! I added that to the notes.

While looking, I became suspicious of the Opera support, so I did some tests on it. It turns out Opera did in fact add support in v44 ([changelog](https://blogs.opera.com/desktop/changelog-for-44/), [blog](https://dev.opera.com/blog/opera-46/)), but it was mysteriously removed again from v54 ([changelog](https://blogs.opera.com/desktop/changelog-for-54/), [blog](https://dev.opera.com/blog/opera-54/)).
The feature was mentioned when it was introduced, but I cannot find any documentation on why it was removed again. v54 did have many updates related to favicons, so maybe it was accidentally removed or just not mentioned in the changelogs ¯\\\_(ツ)\_/¯

Nevertheless, I have now removed support after v54. It will though almost certainly come back in v67 with the Chromium implementation.

fsoder:
> v54 did have many updates related to favicons, so maybe it was accidentally removed or just not mentioned in the changelogs...

IIRC it was dropped based on a maintenance/value trade-off. Because of how internal processes work, dropping a patch like that will not show up in the changelog. Thanks for checking and making the update to reflect reality.

Fyrd:
Thanks!

antonjuulnaber:
@sierkb Support for Chrome on Android will be automagically added later. Caniuse only shows support information for the current stable version.

PR #5182, opened 2019-11-19
Remove duplicate `input type="color"` tutorial

TheLastZombie:
Just something I noticed while looking up browser support.


Fyrd:
Thanks!

PR #5185, opened 2019-11-19
Update IE support for basic SVG

### LINKED ISSUE #2554: No support for CSS Animations on SVG in Internet Explorer
Internet Explorer does not support css transitions/animations on SVG elements. [Test case](http://codepen.io/ChrisPlz/pen/LZEWQe)

### LINKED ISSUE #3236: Edge fixed svg viewport scaling issue
I'm using Win 10 Pro and
#798 seems to be fixed in the current version of edge:

svg

the issue of course still persists in IE 11.576.14393.0 though

antonjuulnaber:
Internet Explorer does not support css transitions/animations on SVG elements.

Closes #2554

I have tested it on all versions of IE and Edge. The lines are a little blurred:
- IE9 breaks all SVG support completely when trying to use animations
- Edge 15 & 16 still dont perfectly support animations (scaling didn't work for me)

But this should be pretty close to the truth, and i don't think it makes sense to spend resources testing these edge cases for browsers that are essentially not used anymore.

It might be worth it to resolve ` ` #3236 though. I couldn't make any sense of it.


features-json/svg.json
```diff
@@ -49,21 +49,21 @@
...
-      "8":"p",
-      "9":"y #2",
-      "10":"y #2",
-      "11":"y #2"
+      "6":"n",
+      "7":"n",
+      "8":"n",
+      "9":"p #2 #3",
+      "10":"p #2 #3",
+      "11":"p #2 #3"
```
Fyrd: Don't think you meant to change `y` to `p` here? (since `p` refers to "supported with polyfill")

antonjuulnaber: Ah yes off course! I had no idea that that existed, confused it with `a` - `Partial Support`

Fyrd:
Thanks!

PR #5186, opened 2019-11-20
Firefox 70 dropped X-Frame-Options ALLOW-FROM

### LINKED ISSUE #5177: Firefox dropped X-Frame-Options ALLOW-FROM support
Support for ALLOW-FROM has been dropped in Firefox 70 (see https://www.fxsitecompat.dev/en-CA/docs/2019/x-frame-options-allow-from-directive-has-been-removed/). Given that this has significant security implications, as FF just ignores the header and allows framing, it would be great to have the Can I Use list updated.

Schweinepriester:
Fixes #5177.

<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/70>

> The [`X-Frame-Options`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options "The X-Frame-Options HTTP response header can be used to indicate whether or not a browser should be allowed to render a page in a <frame>, <iframe>, <embed> or <object>. Sites can use this to avoid clickjacking attacks, by ensuring that their content is not embedded into other sites.") `allow-from *uri*` directive has been removed. Use the [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy "The HTTP Content-Security-Policy response header allows web site administrators to control resources the user agent is allowed to load for a given page. With a few exceptions, policies mostly involve specifying server origins and script endpoints. This helps guard against cross-site scripting attacks (XSS).") header with the [`frame-ancestors`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors "The HTTP Content-Security-Policy (CSP) frame-ancestors directive specifies valid parents that may embed a page using <frame>, <iframe>, <object>, <embed>, or <applet>.") directive instead ([bug 1301529](https://bugzilla.mozilla.org/show_bug.cgi?id=1301529 "FIXED: Misleading console message for X-Frame-Options Allow-From mismatch (remove X-Frame-Options: allow-from)")).

Checked <https://tests.caniuse.com/?feat=x-frame-options> => <https://erlend.oftedal.no/blog/tools/xframeoptions/> using Firefox 70.0.1 as well (equal to Chrome 78.0.3904.97):

image


Fyrd:
Thanks!

PR #5199, opened 2019-11-27
Update Promise.prototype.finally firefox support

maxgomes92:
I tested on Browserstack. Version 69 is the one the support begins.


Fyrd:
Also tested in Browserstack and it works in my test:
https://tests.caniuse.com/?feat=promise-finally
Can you share the test you see failing?

PR #5203, opened 2019-12-04
Update Safari 13 MediaRecorder

jgbpercy:
MediaRecorder is still behind an experimental feature flag in Safari 13


Fyrd:
Confirmed, thanks!

PR #5208, opened 2019-12-11
fetch: add unfetch polyfil

simon04:
"Bare minimum fetch polyfill in 500 bytes."
https://github.com/developit/unfetch


Fyrd:
Thanks!

PR #5220, opened 2019-12-25
Remove reference to non-existing article

jnns:
Thankfully there's already a link to an archived version of this article.


Fyrd:
Thanks!

PR #5223, opened 2020-01-02
More String.prototype.replaceAll() data

TheLastZombie:
This adds

- more tests (using https://html.house/625h1pup.html)
- the [Chrome Platform Status ID](https://www.chromestatus.com/feature/6040389083463680)

to the `String.prototype.replaceAll()` data.


Fyrd:
Thanks!

PR #5224, opened 2020-01-03
Remove duplicate links in webp.json

TheLastZombie:
- http://antimatter15.github.io/weppy/demo.html redirects to http://antimatter15.com/weppy/demo.html, which is already listed
- https://libwebpjs.appspot.com is listed twice
- https://webpjs.appspot.com is listed twice


Fyrd:
Thanks!

PR #5226, opened 2020-01-03
Minor updates to CSS Scroll Snap

Tristan-zimmerman:
- Updated a few terms to no longer use "points" but "positions" instead to cut down on any confusion  between the newest spec and the previous version.
- Corrected the MDN link to point to the current spec description.
- Rearranged the polyfills to put the tool based on the latest spec higher up.


Fyrd:
Thanks!

PR #5227, opened 2020-01-07
Firefox 72 now has CSS Motion Paths

tyteen4a03:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/72


Fyrd:
Thanks!

PR #5230, opened 2020-01-10
Update FF versions for CSS calc bugs

GustvandeWal:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/59#CSS
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/66#CSS


Fyrd:
Thanks!

PR #5232, opened 2020-01-14
Mark Safari 13 and Safari TP bugs as resolved

dhritzkiv:
bug 188259 fixed in 13
both Safari bugs fixed in TP


Fyrd:
If there's no more bugs for Safari TP it should be `y`, right?

dhritzkiv:
@Fyrd oops, yes. Fixed.

Fyrd:
Thanks!

PR #5236, opened 2020-01-15
Firefox 73: Feature Policy (by default)

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/73>:

> Feature Policy is now enabled by default! Use the [`<iframe>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe "The HTML Inline Frame element (<iframe>) represents a nested browsing context, embedding another HTML page into the current one.") element's `[allow](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-allow)` attribute (the [`HTMLIFrameElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement "The HTMLIFrameElement interface provides special properties and methods (beyond those of the HTMLElement interface it also has available to it by inheritance) for manipulating the layout and presentation of inline frame elements.") property `allow`) to set permissions for your frames ([bug 1600883](https://bugzilla.mozilla.org/show_bug.cgi?id=1600883 "FIXED: Enable Feature Policy allow attribute and permission delegation by default")).

image

Seems to have been available behind a flag before, but didnt dig into the past.

Note `#2` 

> Only supports the `allow` attribute on iframes

could be applicable. Should I add it?


Fyrd:
Yes, I think would be appropriate.

Schweinepriester:
Done.

Fyrd:
Thanks!

PR #5241, opened 2020-01-17
Add Chrome support for CSS3 image-orientation

antonjuulnaber:
Chome is enabling this feature by default in v81

https://www.chromestatus.com/feature/6313474512650240


Fyrd:
Thanks!

PR #5242, opened 2020-01-19
SharedArrayBuffer enabled on Firefox Nightly 73+

akien-mga:
See https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/Planned_changes

I tested it on Linux with current Firefox Nightly 74, and it seems to work with the demo on MDN:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer

Firefox Beta 73 does not have it enabled by default, even though Nightly 73 apparently had it,
so it's Nightly only for now. Thus I only put the footnote for 74 since Nightly 73 is no longer available.

### DIFF
--- a/features-json/sharedarraybuffer.json
+++ b/features-json/sharedarraybuffer.json
@@ -110,7 +110,7 @@
       "71":"n d #1",
       "72":"n d #1",
       "73":"n d #1",
-      "74":"n d #1"
+      "74":"n d #1 #2"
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #5243, opened 2020-01-20
FFox droped support of HPKP in FF72

matt32106:
HPKP support has been droped in 72:
https://www.fxsitecompat.dev/en-CA/docs/2019/http-public-key-pinning-is-no-longer-supported/


Fyrd:
Thanks!

PR #5252, opened 2020-01-25
Fix data for WebXR Device API on Android Chrome

antonjuulnaber:
This API is also enabled on Chrome for Android. Not sure why it didn't trickle over from the Chrome desktop branch.

https://www.chromestatus.com/feature/5680169905815552

While I was at it, I also removed the last notes from earlier versions of Chrome, since they are not relevant for any future development.


Fyrd:
Great, thanks!

PR #5256, opened 2020-01-30
Add URL Scroll-To-Text Fragment feature

### LINKED ISSUE #5102: Add Text URL Fragments (#targetText)
- <https://github.com/WICG/ScrollToTextFragment>
- <https://chromestatus.com/feature/4733392803332096>

antonjuulnaber:
1. This looks like a pretty awesome feature that many people will want to use.
2. Chrome is shipping it [now](https://www.chromestatus.com/feature/4733392803332096), but other browsers seem hesitant to implement it - most likely creating a complicated landscape of compatibility.

I think it makes a lot of sense to have this feature as an entry in caniuse.

Closes #5102


Fyrd:
Thanks!

stefanjudis:
FYI: even though Chromestatus states this to be true I don't think it is. 👇 See screenshots of Chrome and Canary for the same URL (https://en.wikipedia.org/wiki/Cat#:~:text=As%20of%202017)

Chrome:

Screen Shot 2020-02-05 at 19 16 35

Chrome Canary:

Screen Shot 2020-02-05 at 19 17 34

antonjuulnaber:
@stefanjudis thanks for noting! I was pretty sure this would happen... @jpmedley something you could update? I assume it will be enabled in Chrome 81 instead?

https://www.chromestatus.com/feature/4733392803332096

jmedley:
Sorry for the late reply. I'll look into this.

jpmedley:
The Chrome Status entry is incorrect. It should say Chrome 82.

antonjuulnaber:
Allright, thanks for checking it out!

PR #5261, opened 2020-02-03
Clarify note

jkshapiro:
Previous wording of note 1 did not make it clear whether arrow up & down keys worked


Fyrd:
Hmm, I agree with rewording it, but I feel that only referring to "increment/decrement" doesn't make it clear what the expected forms of supporting that should be, which seems worth mentioning.

Fyrd:
Excellent, thanks!

PR #5262, opened 2020-02-03
Point out Mozilla's position on WebUSB spec

voxpelli:
Helpful in the light of eg: https://news.ycombinator.com/item?id=22228879


Fyrd:
Thanks!

PR #5267, opened 2020-02-06
Add note about Vivaldi support for `theme-color `

crozet-magenta:
Add note about vivaldi's support for the `theme-color` meta tag


Fyrd:
Thanks!

PR #5270, opened 2020-02-07
SVG Favicons not supported in Edge 79

antonjuulnaber:
Tested to not work in Edge 79, which is expected, since it is not supported in Chrome/Chromium 79.

Test used: http://xn--dahlstrm-t4a.net/svg/favicon/favicon.html

This will probably be supported in Edge 80+


Fyrd:
Thanks!

PR #5272, opened 2020-02-07
Fix initial Chrome support for URL Scroll To Text Fragment

antonjuulnaber:
Despite what the Chrome Platform Status says, this feature will not be enabled by default in Chrome until v82.

Original PR and discussion: #5256


Fyrd:
Thanks!

PR #5277, opened 2020-02-11
Mark Edge 79 as supporting Shadow DOM v0

mgol:
Edge 79 supported Shadow DOM v0 just like Chrome 79. Support was dropped
in Edge 80, just like in Chrome 80.


Fyrd:
Thanks!

PR #5279, opened 2020-02-12
Webhid is also available in Chromium based Edge behind a flag

joegaffey:
Tested in 80; assume also available in 79 based on same engine:
```console
navigator.hid
HID {onconnect: null, ondisconnect: null}
    onconnect: null
    ondisconnect: null
    __proto__: HID
```


Fyrd:
Thanks!

PR #5288, opened 2020-02-21
Add intersectionObserver V2

atjn:
Chrome has proposed an iteration on the intersectionObserver API that makes it possible to track if an element is being overlapped by another element or has filters applied to it.

I needed this information yesterday, so i did the research.

It doesn't seem like the spec will be widely supported any time soon, so i have decided to track it as a subset of the original intersectionObserver API. If it ever recieves support from Safari & Firefox, it should probably become it's own feature.

https://www.chromestatus.com/feature/5878481493688320
https://developers.google.com/web/updates/2019/02/intersectionobserver-v2

[Test used](https://github.com/Fyrd/caniuse/files/4236078/intersectionobserver-v2-test.txt)

### Tested to be enabled by default in:
Chrome, Android Chrome, Android Browser 74 [`matches doc`](https://www.chromestatus.com/feature/5878481493688320)
Edge 79 `as expected (matches chrome release)`
Opera 61/62 `as expected (matches chrome release)`

Opera Mini and Opera Mobile do have support, but not the versions listed on caniuse.

### Tested to not work in the latest version of:
Firefox, Android Firefox
Safari, IOS Safari
Samsung Internet
Android UC Browser
Internet Explorer `as expected (doesn't have V1 support either)`
(Edge 18)

The last few browsers are too old to have support.


Fyrd:
Looks good, thanks!

Issue #5286, opened 2020-02-19
Difference theme-color Meta Tag between Edge and Chrome 79/80


TQxTim:
![image](https://user-images.githubusercontent.com/43742253/74821734-56d51080-5304-11ea-8711-a0c867b9a15d.png)
How is this possible, if both use the same engines?

atjn:
Chrome and Edge are free to customize Chromium however they'd like, so it is definetely possible for them to have feature disparity, especially when it relates to the UI, which is very clearly different from browser to browser.

With that said, it seems like Edge does support theme-color, so I will take look at it. Thanks for noticing!

atjn:
Nope, sorry. Edge does not support the theme-color meta tag. It does however support the theme-color attribute in the web app manifest, which is why i was originally confused.

You can test this yourself by installing the [Google Developer webapp](https://developers.google.com) in both Chrome and Edge. Notice that the theme-color is currently set to white, because that is what is defined in the web app manifest.
Now navigate to [this page](https://developers.google.com/web/updates/2014/11/Support-for-theme-color-in-Chrome-39-for-Android) within the webapp. Chrome will change the toolbar to blue because it reads the theme-color meta tag on the page (which is set to blue), but Edge does not.

Fyrd:
Closing this issue as per @atjn's comments

PR #5289, opened 2020-02-21
Small improvements to theme-color meta tag

atjn:
While closing #5286 I learned a few things that I wanted to add.

[I found a way to test](https://github.com/Fyrd/caniuse/issues/5286#issuecomment-589801665) if the browser is actually reading the theme-color meta tag, as opposed to just the web app manifest theme-color. Turns out web app support does not inherently include theme-color meta tag support (Edge being an example). Despite this, I have tested Chrome 73+ to have support, so I updated that comment (1c09807) to reflect this.

I added a note (82c370f) about older versions of Chrome. I belive this note is important because developers will look at the desktop branch, and assume it is also showing when support was added in Chrome for Android, when in fact, support in Android was added much earlier on.

I tested all browsers with support (using the [aforementioned method](https://github.com/Fyrd/caniuse/issues/5286#issuecomment-589801665)), and they all seem to prefer the theme-color meta tag over what is defined in the web app manifest. I thought this was an important thing for developers to note (ae8dafc).

I found out that using native dark-mode on Android 10 will cause Chrome to not use the theme-color in the toolbar. That took a lot of head-scratching to figure out, so I added a quick note about it (4d44595).

I also tested the feature to not be supported (or at least not used) in Edge, Opera, Safari and Firefox, all desktop and mobile.


Fyrd:
Great research, thanks for the updates!

PR #5291, opened 2020-02-21
Chrome 82 removes support for offline apps / application cache

atjn:
https://www.chromestatus.com/feature/6192449487634432


Fyrd:
Thanks!

PR #5292, opened 2020-02-22
Update notes on SVG favicons to include newer browsers

atjn:
Just a quick iteration now that Edge is also supporting this, and I noticed the Safari note wasn't updated either.


Fyrd:
Thanks!

PR #5296, opened 2020-02-25
Firefox 75: CSS math functions min(), max() and clamp() 🎉

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/75>

> Support for the [`min()`](https://developer.mozilla.org/en-US/docs/Web/CSS/min "The min() CSS function lets you set the smallest (most negative) value from a list of comma-separated expressions as the value of a CSS property value. The min() function can be used anywhere a <length>, <frequency>, <angle>, <time>, <percentage>, <number>, or <integer> is allowed."), [`max()`](https://developer.mozilla.org/en-US/docs/Web/CSS/max "The max() CSS function lets you set the largest (most positive) value from a list of comma-separated expressions as the value of a CSS property value. The max() function can be used anywhere a <length>, <frequency>, <angle>, <time>, <percentage>, <number>, or <integer> is allowed."), and [`clamp()`](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp "The clamp() CSS function clamps a value between an upper and lower bound. clamp() enables selecting a middle value within a range of values between a defined minimum and maximum. It takes three parameters: a minimum value, a preferred value, and a maximum allowed value. The clamp() function can be used anywhere a <length>, <frequency>, <angle>, <time>, <percentage>, <number>, or <integer> is allowed.") functions has been implemented ([bug 1519519](https://bugzilla.mozilla.org/show_bug.cgi?id=1519519)).

<https://tests.caniuse.com/?feat=css-math-functions>:

image

🎉


Fyrd:
Thanks!

PR #5206, opened 2019-12-11
Signed http exchanges

bazzadp:
Fixes #4920 

Note that some articles [stated that Opera has support](https://www.bleepingcomputer.com/news/google/google-chrome-adding-support-for-signed-http-exchanges/) but [testing it](https://blog.amp.dev/2018/11/13/developer-preview-of-better-amp-urls-in-google-search/) did not show it working, despite being based on Chrome 76 and I can find no mention of that in any Opera release notes. So I have stated it is not supported.

PR #5308, opened 2020-03-08
Update SXG info

bazzadp:
This PR makes the following changes for Origin Signed HTTP Exchanges (committed in https://github.com/Fyrd/caniuse/pull/5206 - note this is currently not live yet):
- I've confirmed Opera support - couldn't get this working when I originally submitted this but confirmed it works now even back to same version (64). Not sure if it was off by a flag but definitely works now.
- Edge have changed one of their URLs so fixed that.
- Set 'Unknown' rather than 'No support' for some of the mobile browsers where they are based on a version of Chromium which _may_ support this.

FYI, support can be confirmed by:
- Launching in mobile emulator mode if on desktop
- Visiting https://www.google.com/?esrch=WebPackage::Demo
- Searching for "amp dev"
- Launching an "amp.dev" URL with lightning bolt and confirming if "amp.dev" shows in URL (it's supported) or "google.com" (it's not supported).


Fyrd:
Thanks!

PR #5320, opened 2020-03-18
Fix data in Drag & Drop feature

### LINKED ISSUE #5042: Chrome for Android - wrong values

atjn:
It turns out Chrome for Android only supports this feature in Android 7 or newer, so I have added a note about that. The real-world use of Android 6 or older [seems to be so low](https://gs.statcounter.com/os-version-market-share/android/mobile-tablet/worldwide) that it doesn't make sense to mark the feature as only partially supported.

 See following discussion for testing details:

Closes #5042

I also tested it to be supported in Opera and Android Browser, but could not make it work in Samsung Internet.
When I was not able to make the test work, I tried replacing the `textarea` with a simple `<p>` tag. Sometimes the `textarea` was prioritized over the dragndrop functionality, so it would look like it wasn't supported, even though it was. I was not able to test the `<p>` tag on Samsung Internet, so I might have mislabeled that.

On another note, I tested the feature on a few different iPhones running IOS 13, 12 and 11. I was not able to make Drag & Drop work on any of them. I also could not find any pull request or discussion in the repository regarding IOS support, so I assume it was just added by mistake during a browser version update or something like that.
I have removed support from those versions.


Fyrd:
Thanks for doing the research! Testing drag 'n drop support on mobile has always been a bit tricky so I appreciate you doing it.

PR #5321, opened 2020-03-19
Firefox 73: WebVR requires a secure context

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/73>:

> The deprecated [WebVR API](https://developer.mozilla.org/en-US/docs/Web/API/WebVR_API)---which has been supplanted by [WebXR](https://developer.mozilla.org/en-US/docs/Web/API/WebXR_Device_API), which supports both [augmented](https://en.wikipedia.org/wiki/Augmented_reality) and [virtual reality](https://en.wikipedia.org/wiki/Virtual_reality) applications---now [requires a secure context](https://developer.mozilla.org/en-US/docs/Web/API/WebVR_API#API_availability) using the [HTTPS](https://developer.mozilla.org/en-US/docs/Glossary/HTTPS) protocol to operate. This is due to the availability of sensitive input sources that may include private information ([bug 1381645](https://bugzilla.mozilla.org/show_bug.cgi?id=1381645)).

Would be probably more accurate to add an additional note to Firefox 73+ giving the added information, but since its deprecated anyway and support is dwindling…

### DIFF
--- a/features-json/webvr.json
+++ b/features-json/webvr.json
@@ -388,7 +388,7 @@
   },
   "notes":"For a WebVR experience a head mounted display (VR HMD) is required.",
   "notes_by_num":{
-    "1":"Available and enabled by default only in Firefox Windows. Enabled in Nightly for iOS.",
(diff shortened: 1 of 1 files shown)


Fyrd:
Yeah, seems fine. Thanks!

PR #5324, opened 2020-03-23
Firefox 74 - 75: Re-enabled TLS 1.0/1.1

Schweinepriester:
<https://www.mozilla.org/en-US/firefox/74.0/releasenotes/>

> ~~We have [disabled TLS 1.0 and TLS 1.1](https://hacks.mozilla.org/2020/02/its-the-boot-for-tls-1-0-and-tls-1-1/) to improve your website connections. Sites that don't support TLS version 1.2 will now show an error page.~~
>
> **We reverted the change for an undetermined amount of time to better enable access to critical government sites sharing COVID19 information.**

<https://www.fxsitecompat.dev/en-CA/docs/2020/tls-1-0-1-1-support-has-been-removed/>

> Update: Mozilla is going to temporarily re-enable the TLS 1.0/1.1 support in Firefox 74 and 75 Beta. The preference change will be remotely applied to Firefox 74, which has already been shipped. This is because many people are currently forced to work at home and relying on online tools amid the novel coronavirus (COVID-19) outbreak, but some of critical government sites still don't support TLS 1.2 yet. Also due to COVID-19, Google has [postponed the release of Chrome 81](https://blog.chromium.org/2020/03/upcoming-chrome-releases.html) that would [remove their TLS 1.0/1.1 support](https://www.chromestatus.com/feature/5759116003770368). We'll update this note when the situation has changed.

- Meta bug: <https://bugzilla.mozilla.org/show_bug.cgi?id=1227521>
- <https://bugzilla.mozilla.org/show_bug.cgi?id=1623534>
- <https://bugzilla.mozilla.org/show_bug.cgi?id=1623536>

Firefox 75+ dont seem to be locked in at the moment, but 75.0b7 currently has both:

image

Technically <https://caniuse.com/#search=TLS%201.0> could also be changed, but then again…

---

Somewhat related, looks like the flag will stick around longer in Firefox: <https://bugzilla.mozilla.org/show_bug.cgi?id=1618615>


Fyrd:
Thanks!

PR #5331, opened 2020-03-30
Update css-image-orientation.json

Matt-rempel:
Safari TP version 103 supports image-orientation CSS3 property


Fyrd:
Are you sure? Looking at https://developer.apple.com/safari/technology-preview/release-notes/ I see mentioning of using the EXIF image orientation but there's no mention of the image-orientation property. In which case it should probably match the `a #1` support that iOS Safari already has.

Fyrd:
Closing until further info is provided.

PR #5332, opened 2020-03-30
Update Web Animations data for Safari

graouts:
The latest releases of Safari (13.1 on macOS and for iOS 13.4) now support the full Web Animations API, going beyond simply `Element.animate()`, and exposing CSS Animations and CSS Transitions as Web Animations objects using `Document.getAnimations()` and `Element.getAnimations()`. Previous versions only supported Web Animations as an experimental feature.

This is my first PR for caniuse.com, so I hope it is useful. I wasn't sure whether the level of support in Safari, which is the entire API save for one feature – [effect composition](https://drafts.csswg.org/web-animations-1/#effect-composition) – warranted an `a` or `y` level of support. I went with `y` in an optimistic manner, but please advise if maintainers of this repository think that is incorrect.

### DIFF
--- a/features-json/web-animation.json
+++ b/features-json/web-animation.json
@@ -230,8 +230,8 @@
       "12":"n d #4",
       "12.1":"n d #4",
       "13":"n d #4",
-      "13.1":"a",
-      "TP":"a"
+      "13.1":"y #5",
+      "TP":"y #5"
     },
     "opera":{
       "9":"n",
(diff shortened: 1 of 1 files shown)


graouts:
There is an issue for MDN to update Web Animations data for Firefox and Safari at https://github.com/mdn/sprints/issues/2933.

Fyrd:
Thanks! I'm fine with `y` as support level here. One thing you will need to correct however is the version numbers and ranges need to match the originals, can't collapse to 13.0-13.3.

graouts:
The diff should now be clean. Let me know if there's any further work required before this could be merged.

Fyrd:
Looks good, thanks!

graouts:
Thanks @Fyrd. When are changes deployed to caniuse.com?

Fyrd:
Ah, forgot to push the button, it's there now. :)

graouts:
👍

PR #5338, opened 2020-04-03
iOS 13.4: replaceAll

Schweinepriester:
<https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/>

> These releases include new JavaScript support for the `replaceAll()` method for strings

(Safari 13.1 and TP already got `y`)


Fyrd:
Thanks!

PR #5343, opened 2020-04-04
iOS 13.4: tab-size

Schweinepriester:
Safari 13.1 and TP already got `y` per #4979.

Works also in iOS 13.4:

EC52811A-DD11-4A43-8660-F2FB87B1C094

:)


Fyrd:
Thanks!

PR #5346, opened 2020-04-04
Update keyboardevent-code.json

gorisanson:
Referred "Browser compatibility" at:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Browser_compatibility


Fyrd:
Based on my test it still fails in these browsers, so I believe that information is incorrect.

gorisanson:
When I tested (I tested on the Android Chrome browser), it worked. I tested by connecting bluetooth keyboard to an Android smartphone. 

Based on https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code,
it detects a physical key. I guess maybe the reason your test failed is because you tested on virtual keyboard. Right?

gorisanson:
> Based on my test it still fails in these browsers, so I believe that information is incorrect.

When I tested (I tested on the Android Chrome browser), it worked. I tested by connecting bluetooth keyboard to an Android smartphone.

Based on https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code,
it detects a physical key. I guess maybe the reason your test failed is because you tested on a virtual keyboard, not a physical keyboard. Right?

gorisanson:
@Fyrd I fount this site https://javascript.info/article/keyboard-events/keyboard-dump/ where we can easily test [KeyboardEvent.key](https://caniuse.com/#feat=keyboardevent-key) and [KeyboardEvent.code](https://caniuse.com/#feat=keyboardevent-code).

And when I tested the Chrome for Android, an virtual keyboard fails both KeyboardEvent.key and KeyboardEvent.code, and an physical (bluetooth) keyboard successes both KeyboardEvent.key and KeyboardEvent.code.

PR #5350, opened 2020-04-08
Update Chrome removal of TLS 1.1

atjn:
Chrome is delaying the removal of support for TLS 1.0 and 1.1 to Chrome 84.

This is an effort to not break websites that are providing critical services during the Coronavirus outbreak.
https://www.chromestatus.com/feature/5759116003770368

I tested and confirmed Chrome 81-83 support using the [Qualys SSL Labs Client test](https://www.ssllabs.com/ssltest/viewMyClient.html)


Fyrd:
Thanks!

PR #5067, opened 2019-09-04
Safari TP 91: Disabled legacy TLS 1.0 and TLS 1.1

Schweinepriester:
<https://webkit.org/blog/9526/release-notes-for-safari-technology-preview-91/>:

> Disabled legacy TLS 1.0 and TLS 1.1 ([r249019](https://trac.webkit.org/changeset/249019/webkit/))

Bildschirmfoto 2019-09-05 um 00 06 04

---

We should update TLS 1.0 (https://caniuse.com/#search=TLS%201.0) as well - maybe at least with the note from <https://caniuse.com/#feat=tls1-1>?

> TLS 1.0 & 1.1 are deprecated in Chrome, Edge, Firefox, Internet Explorer 11, & Safari. Support will be removed in March 2020.

PR #5351, opened 2020-04-08
Safari 13.1 & TP: TLS 1.1

Schweinepriester:
Couldn't really find any public statement/bug beyond

<https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/>

> Following our announcement of deprecating TLS 1.0 and TLS 1.1 connections, this release now adds a “Not Secure” warning when connecting to a site where any resource is using either of these deprecated encryption protocols.

But testing shows its also back contrary to #5067:

image

image

iOS 13.4 still got `y`, which is the current state, tested that as well.


Fyrd:
Thanks!

PR #5352, opened 2020-04-09
Adjusting several Safari 13+ notes/supports regarding min macOS version

Schweinepriester:
Per <https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility> Safari 13+ (and current TP) is not available on less than macOS 10.13 High Sierra, so adjusting notes/supports for

- https://caniuse.com/#feat=flac
- https://caniuse.com/#feat=hevc
- https://caniuse.com/#feat=brotli
- https://caniuse.com/#feat=variable-fonts


Fyrd:
Great catch, thanks!

PR #5354, opened 2020-04-09
Rework Note in Safari WebGL2

NotAFile:
I found this note hard to read. I shortened the sentences, tried to fix the grammar and added a date.


Fyrd:
Thanks! However this note should actually be updated as there has been recent WebGL 2 development, see https://developer.apple.com/safari/technology-preview/release-notes/ (look for "WebGL2" and "WebGL 2"

NotAFile:
Huh. I had looked at the source code and not found any further changes. I guess I'll update it then.

Fyrd:
Have manually updated this to only use the first line, as per my last comment.

PR #5360, opened 2020-04-14
Safari 13 no longer requires vendor prefix for position sticky

bazzadp:
Can't find a reference for but tested and v13 of Safari no longer requires vendor prefix for sticky (not required for Desktop nor iOS):
```
    position: -webkit-sticky;
    position: sticky;
```
Version 12 still requires this.

Confirmed with testing locally on v13 (Desktop and iOS) and v12 (iOS only I'm afraid as don't have desktop).


Fyrd:
Thanks!

Issue #5342, opened 2020-04-03
Add font keywords ui-serif, ui-sans-serif, ui-monospace and ui-rounded


Schweinepriester:
<https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/>

> New font keywords are available for using platform-specific fonts including `ui-serif`, `ui-sans-serif`, `ui-monospace` and `ui-rounded`.

Spec: <https://drafts.csswg.org/css-fonts-4/#ui-serif-def> (CSS Fonts Module Level 4)

Firefox:
- <https://bugzilla.mozilla.org/show_bug.cgi?id=1598879>
- <https://bugzilla.mozilla.org/show_bug.cgi?id=1598880>
- <https://bugzilla.mozilla.org/show_bug.cgi?id=1598881>
- <https://bugzilla.mozilla.org/show_bug.cgi?id=1598883>

Booligoosh:
+1

Booligoosh:
Would it be best to have these as four separate features or one feature for all of them? If it was one feature for all of them, how would it work with Firefox IDs etc?

Schweinepriester:
Now available at https://caniuse.com/#feat=extended-system-fonts.

Thanks @Booligoosh! :)

Booligoosh:
Awesome!!

One thing I thought I'd double check — I set the status to WD, since the system-ui page links to the same document and its status is WD:
https://github.com/Fyrd/caniuse/blob/692d2141603fe98ccb77b2f5ef6795bfd84826c2/features-json/font-family-system-ui.json#L4

However, the actual document says Editor's Draft as opposed to Working Draft. I searched this repo for "ed" and couldn't find anything so I assume you don't differentiate between the two?

Schweinepriester:
> (*EDIT:* Never mind --- it appears that they are in the [Working Draft](https://www.w3.org/TR/2019/WD-css-fonts-4-20191113/#valdef-font-family-ui-serif) as well. Should we be linking to that instead or is it better to link to the regularly updated Editor's Draft?)

was about to comment that 0:-)

Imho linking to the Editor's Draft is better, at least in this case, as for example `ui-sans-serif` is missing the example in the WD <https://www.w3.org/TR/css-fonts-4/#ui-sans-serif-def> but has it in the ED <https://drafts.csswg.org/css-fonts-4/#ui-sans-serif-def>.

@Fyrd, what do you think?

Fyrd:
[...]

PR #5361, opened 2020-04-16
Add ui-serif, ui-sans-serif, ui-monospace and ui-rounded values for font-family

Booligoosh:
Based on issue #5342. This is my first time contributing, so let me know if I need to fix things (although I have read the CONTRIBUTING file).

Test CodePen is here: https://codepen.io/Booligoosh/pen/YzywJBd

The ID is a bit long, but I was unsure what else to use for it.


Fyrd:
Looks good! Thought about the id, how about `font-family-system-ui-2` ? A bit weird, but it indicates it's a follow up of `font-family-system-ui` while leaving room for us to change what values it refers to if necessary in the future.

Booligoosh:
Oh yeah, that could work. Or how about something like `extended-system-fonts`? Might make it a little clearer what the ID is meant to mean.

Fyrd:
Yeah, agreed that's probably better. Go for it!

Booligoosh:
Done! I've left the title as-is but let me know if you want me to change that as well.

Fyrd:
Oh, and the file needs to end with `.json`

Booligoosh:
Done, sorry about that. Travis is now failing with the following error:
```
Browser version missing: ios_saf 13.0-13.1
```
I split it up into separate fields in the file, since support started in 13.1, but is not there in 13.0:
```
...
"12.2-12.4":"n",
"13.0":"n",
"13.1":"y",
"13.2":"y",
"13.3":"y",
"13.4":"y"
...
```
What should I do about that?

Fyrd:
For iOS the version refers to the OS version, not the browser version. I believe iOS 13 & 13.1 use the same browser engine which is why they're combined. I think iOS 13.4 is where support starts in this case.

Booligoosh:
Ohhh, that makes sense. Thanks for picking that up.

Fyrd:
Thanks!

PR #5363, opened 2020-04-17
Update Opera 67 support of link-icon-svg

antoineboulanger:
The current version of Opera (67) does support SVG favicons, you can test it on this website: http://dahlström.net/svg/favicon/favicon.html. 

I don't know if it was supported in previous versions but it is weird that it was supported in versions `44-53` and not supported in versions `54-66`: https://caniuse.com/#feat=link-icon-svg  . If someone can verify that.


atjn:
Nice catch! Unfortunately that browser version is not tracked on Caniuse just yet, and you can't add new versions manually. It's right [here](https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#unsupported-changes) in the contributing guidelines.

The temporary addition in version 44-53 is wierd indeed, and has been discussed and confirmed by Opera staff in PR #5181
Opera versions 54-66 did not support the feature.

You are welcome to keep this PR open and wait until Opera 67 is tracked, just be aware that you will need some fancy manual merging, since you will be merging two documents that have separate overlapping updates. There is also a chance that Fyrd will update this by himself, because he knows Opera 67 is based on Chrome 80, which was the first Chrome version to add support.

Please don't see this as discouragement to contribute. Contributions are much appreciated, i just want to help you get your PR's merged 😊

Fyrd:
Yup, went ahead and added Opera 67 with support set. Thanks anyway! (and thanks @atjn for your comment!)

PR #5366, opened 2020-04-18
Update Chrome removal of TLS 1.1

atjn:
A quick follow-up on #5350

Chrome 84+ is removing support. I have confirmed with a test. Also just removed the comment about support being removed in March.

Opera 67 is tested to still support TLS 1.1, meaning it's following Chromiums lead on postponing the removal.


Fyrd:
Thanks!

PR #5367, opened 2020-04-18
Chrome 84+: CSS revert value

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=579788#c37>

> This enables support for the 'revert' keyword once again

~The current Canary (84.0.4118.0) tests <http://tests.caniuse.com/?feat=css-revert-value> as not working. Therefore draft pull request, but it worked before (before the first VCS-revert) and probably will again in the next build. Will update then.~

There we go, 84.0.4120.0:

image

<https://www.chromestatus.com/feature/5644990145363968>

---

> Revert the revert of 'revert'

lol!


Schweinepriester:
Fair enough!

In this case however <https://www.chromestatus.com/feature/5644990145363968> has been updated yesterday (2020-04-21) and now shows

> **Enabled by default** ([tracking bug](https://bugs.chromium.org/p/chromium/issues/detail?id=579788)) in:
> 
> - Chrome for desktop release 84
> - Chrome for Android release 84
> - Android WebView release 84

So it seems to be coming with 84?

Sure enough delays can always happen, e.g. recently the delayed removal of TLS 1.1, and we can always care for those when they happen…

Fyrd:
@Schweinepriester Thanks!

@jpmedley Interesting, if you don't think features should be added until they're at least in beta, do you think caniuse shouldn't be tracking beyond beta in the first place? Generally folks are pretty excited to update caniuse once features appear in canary, I'd like to think most developers know it's not guaranteed for that feature to wind up in Chrome stable.

jpmedley:
When Chrome feature owners update status entries with a future Chrome version it's just as likely to be aspirational as factual.

PR #5371, opened 2020-04-20
iOS Safari 13.4 adds experimental support for loading lazy

sod:
`` support can be enabled in Settings under the Safari > Advanced > Experimental Features menu since iOS safari 13.4. and is disabled by default.

https://bugs.webkit.org/show_bug.cgi?id=200764 was the ticket to track the implementation.
https://bugs.webkit.org/show_bug.cgi?id=208094 tracks when it moves from experimental to default.

image


Fyrd:
Thanks!

PR #5374, opened 2020-04-21
Fix inconsistencies in TLS 1.2 support

atjn:
For some reason, Caniuse lists initial Chrome support for TLS 1.2 at v30, when it was actually supported since v29. This aligns with the [official documentation](https://www.chromestatus.com/feature/5197666342404096) and i also tested it across Windows 8, 7, XP and MacOS Sierra. The same applies to Opera 16.

I ran the same tests on Opera 12.1 and was not able to make TLS 1.2 run by default anywhere.

Tested using the [Qualys SSL Client Test](https://www.ssllabs.com/ssltest/viewMyClient.html)


Fyrd:
Thanks!

PR #5375, opened 2020-04-22
New browser support for TLS 1.3

atjn:
Added Android browser and Samsung browser support for TLS 1.3

Android browser 80 was tested to support it.

Samsung browser 11.1 was also tested to support it. I was not able to test 10.1, but since it is based on a Chromium version that did support it, i think it's a safe bet.

Tested using the [Qualys SSL Client Test](https://www.ssllabs.com/ssltest/viewMyClient.html)

Note `3` was a duplicate of note `1`, so i merged them.


Fyrd:
Good changes, thanks!

PR #5380, opened 2020-04-24
Extensive update to Referrer Policy data

atjn:
It seems like this data hasn't been touched in a while and needs a few updates. Some of these changes are debatable, so i have split everything into separate commits to make it easier to discuss/discard specific changes.

### History
As far as i could gather, this support table was initially generated for an [old specification](https://wiki.whatwg.org/wiki/Meta_referrer) that is very different from the spec we have today. Then when the new spec was implemented, the support data started tracking the new spec, but much of the wording in the description/notes were, and still is targeted towards the old spec.

This confused me a great deal. I started thinking that this support table was specifically aimed at tracking delivery via meta tags for the new spec, when in reality it was just the old description for the old spec... or at least that's what i think is going on. So:

### New description
The new description does not mention the meta tag, because the new spec has multiple ways of delivering the policy. This should stop the confusion that i also experienced when trying to understand what the data is tracking.

### General improvements
I updated the old notes to be more useful for current development. I also added a Chrome ID and updated the link for the spec, which is now a candidate recommendation.

### Fixes to IOS support
IOS Safari has supported the new spec for a while now, but it was not reflected in the data. I ran some tests to confirm that the new spec was added in IOS Safari 12.0.
I couldn't find any good test for it, so i just manually inspected some referrer headers when jumping between sites that use `strict-origin-when-crossorigin`.

### New note on default behaviour
Chrome is [changing the default referrer policy](https://www.chromestatus.com/feature/6251880185331712) to `strict-origin-when-crossorigin` in v84. I think we should be tracking this, because it seems like all browsers will implement it, and if they do, it will essentially mean that most developers can remove the referrer policy entirely from their site and stop worrying about it 🥳

Currently, you have to define it if you want your site to follow good general practice. This is very clearly outlined in the [securityheaders test](https://securityheaders.com/?q=caniuse.com&followRedirects=on#:~:text=should%20be%20set%20by%20all%20sites).

I will be back in a while to update the note and hopefully have all major browsers onboard.

### Room for improvement
It seems pretty unlikely that Chrome had support for the new spec all the way back in 2012, and the same goes for Opera, which should probably also have note `#2` slapped on a few old versions.

My test was pretty clunky so i didn't feel like doing rigorous testing on all the old browsers that no one uses anymore, but i still wanted to mention that there is potential for even more fixes.


Fyrd:
Excellent changes, thanks!

PR #5384, opened 2020-04-24
Fix Opera support for background sync

atjn:
It turns out Opera didn't enable support for background sync until `v42`, even though they could have done it since `v36`.

I tested it with [this test](https://jakearchibald.github.io/isserviceworkerready/demos/sync/). It's also in their [changelog](https://blogs.opera.com/desktop/changelog-for-42/#:~:text=Enable%20Background%20Sync).


Fyrd:
Thanks!

PR #5388, opened 2020-04-26
Update support for Scroll-To-Text Fragment

atjn:
Official as well as unofficial sources are still all over the place on this feature.

Thankfully, i can just test everything now:
Chrome enabled this in 81 on desktop, mobile and in the android browser
Opera enabled this in 58
Edge has not enabled this in 81

Original releases of Chrome 81 did not have this enabled, but there has since been an update to "81 v2" that enabled this and many other features that were supposed to be enabled in Chrome 82.
I don't think it's worth the hassle to disclose this. Almost every person running Chrome 81 will have the newer version.

Edge did not enable this yet, but they are very optimistic about it, so i expect it to be enabled in 83. They probably just didn't handle the wierd "81 v2" update the same way Opera did.

With these changes, all support data is trustworthy, so i'd say you can make it public now 😇


Fyrd:
Thanks for the research!

PR #5390, opened 2020-04-27
Some details about percent encoding in data URIs

Yaffle:
@Fyrd , what is the version used by the original author?
I have tested on Firefox 3.6 and Firefox 4, but I didn't try to find the bug in the https://bugzilla.mozilla.org ...

### DIFF
--- a/features-json/datauri.json
+++ b/features-json/datauri.json
@@ -23,7 +23,7 @@
   ],
   "bugs":[
     {
-      "description":"Non-base64-encoded SVG data URIs need to be uriencoded to work in IE, Edge and Firefox < 4 as according to the specification. For Edge 18 it is needed to encode '\', '^', '`', '|', and '\x7F'."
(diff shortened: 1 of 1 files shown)


features-json/datauri.json
```diff
@@ -23,7 +23,7 @@
   ],
   "bugs":[
     {
-      "description":"Non-base64-encoded SVG data URIs need to be uriencoded to work in IE and Firefox as according to the specification."
+      "description":"Non-base64-encoded SVG data URIs need to be uriencoded to work in IE, Edge and Firefox < 4 as according to the specification. For Edge 18 it is needed to encode '\', '^', '`', '|', and '\x7F'."
```
Fyrd, line 26:
```suggestion
      "description":"Non-base64-encoded SVG data URIs need to be uriencoded to work in IE, Edge and Firefox < 4 as according to the specification. For Edge 18 it is needed to encode '\\', '^', '`', '|', and '\\x7F'."
```
Needs to escape the slashes to be valid JSON

Fyrd:
@Yaffle Don't know I'm afraid. Doesn't seem too import though so wouldn't worry about it.

Fyrd:
Thanks!

Yaffle:
@Fyrd, thank you!

PR #5391, opened 2020-04-28
Update font-variant-alternates spec URL

sideshowbarker:
`font-variant-alternates` was dropped from the CSS Fonts Module Level 3 spec, so https://www.w3.org/TR/css-fonts-3/#propdef-font-variant-alternates no longer works.

`font-variant-alternates` is now just in the CSS Fonts Module Level 4 spec, https://drafts.csswg.org/css-fonts-4/#propdef-font-variant-alternates


Fyrd:
Thanks!

PR #5397, opened 2020-04-28
Update css-hyphens.json

mattbrundage:
Update URL for "Chrome bug for implementing hyphenation" (Issue was split to cover implementation in Win/Linux/ChromeOS)


Fyrd:
Thanks!

PR #5402, opened 2020-04-29
Update Date.prototype.toLocaleDateString spec URL

sideshowbarker:
https://tc39.github.io/ecma402/#sec-Date.prototype.toLocaleDateString no longer works (no such anchor). https://tc39.es/ecma402/#sup-date.prototype.tolocaledatestring works.


Fyrd:
Thanks again!

PR #5407, opened 2020-05-03
Normalize ECMA-402 features' `spec` and `categories`

takenspc:
Normalize ECMAScript Internationalization API (ECMA-402) features' `spec` and `categories`.

- Use https://ts39.es/ecma402/ for `spec` url
- Make `categories` be `[ "JS" ]` as they are JavaScript features

    ECMA-402 supersedes several functions of ECMAScript (ECMA-262) so that it should be considered as JS core features.


Fyrd:
Thanks!

PR #5408, opened 2020-05-03
Normalize ECMA-262 features' `spec`, `categories`, and `status`

takenspc:
Normalize ECMAScript (ECMA-262) features' `spec`, `categories`, and `status`

- Use https://ts39.es/ecma262/ for `spec` url

    https://www.ecma-international.org/publications/standards/Ecma-262.htm says

    > The latest drafts are available at: https://tc39.github.io/ecma262/. Reporters should generally only file bugs if the bug is still present in the latest drafts.

    Thus referencing https://ts39.es/ecma262/ is better suited for developers who read specifications.

- Update `spec` url to point https://ts39.es/ecma262/ for [finished proposal](https://github.com/tc39/proposals/blob/master/finished-proposals.md)
- Make `categories` be `[ "JS" ]` as they are JavaScript features
- Make `status` be `other` (mainline features) of `unoff` (proposals) regarding CONTRIBUTING.md

### DIFF
--- a/features-json/array-find-index.json
+++ b/features-json/array-find-index.json
@@ -1,7 +1,7 @@
 {
   "title":"Array.prototype.findIndex",
   "description":"The `findIndex()` method returns the index of the first element in the array that satisfies the provided testing function.",
-  "spec":"https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.findindex",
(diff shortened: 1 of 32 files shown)


Fyrd:
Nice updates, thanks!

PR #5409, opened 2020-05-03
Add experimental support for AVIF as of firefox 77

badcf00d:
It's disabled by default but Mozilla has added experimental AVIF image support that will be included in the release of Firefox 77 - it's already included in the Firefox nighties. It can be enabled via the `image.avif.enabled` flag in `about:config`

https://bugzilla.mozilla.org/show_bug.cgi?id=1625363


Fyrd:
Thanks!

PR #5412, opened 2020-05-04
Update Chrome support for CSS appearance

atjn:
Chrome is adding unprefixed support for CSS `appearance` in v84 - yay!

This is tested with the [Caniuse test](https://tests.caniuse.com/?feat=css-appearance) and is documented:
https://www.chromestatus.com/feature/4715298156445696

I also updated the spec link. The spec status was correct, but the link was to an older specification from before it was a working draft.


Fyrd:
Thanks!

PR #5413, opened 2020-05-04
firefox 77 avif support

Spacefish:
see https://www.phoronix.com/scan.php?page=news_item&px=Firefox-77-AVIF-Experimental


Schweinepriester:
Not correct 

> Experimental AVIF support can be enabled by turning on `image.avif.enabled` in `about:config`.

and duplicate of #5409.

Fyrd:
Yup, thanks for the info but #5409 already has it covered, with the correct status.

PR #5416, opened 2020-05-06
Update Firefox support for tls 1.1

atjn:
Firefox 76 is out, and even though the beta disabled support for TLS 1.1, the final release has re-enabled it. (Tested with the [Qualys SSL client test](https://clienttest.ssllabs.com:8443/ssltest/viewMyClient.html))

As far as i can tell, Firefox will continue to enable TLS 1.1 just before release for as many versions is needed before the COVID crisis is over. They don't seem to have a roadmap for it.
Because of that, i think it makes sense to mark future versions as unknown and just test it when they come out.


Fyrd:
Sounds good, thanks!

PR #5417, opened 2020-05-06
Added prefix tag to Chrome 29

fjdelarubia:
I'm working in a project I'm working in a project in which we need to develop for Chrome 29 :( and, when we used display:flex we detected that, without the prefix, the display property is not valid. I'm testing with a Chromiun 29 and I need to set the browserlist compatibility to Chrome 28 in order to get the correct prefixes.

Here there are some screenshot in order to verify the problem:

image

image

image

With the prefix works perfectly


Fyrd:
That doesn't match my own test using Chrome 29 nor any other information out there (checked MDN and other information sources). Perhaps there's a difference with Chromium and Chrome in that regard, since your screenshot shows Chromium?

Also may I ask why you need to support such an old version?

fjdelarubia:
We are working for an old device and  it seems that the internal browser is similar to Chrome 29.

How do you test in that Chrome version? I mean... where can I find that chrome version? not the Chromium one? I will try to figure out what is the difference between browsers but, anyway is not a big deal, we just set Chrome 28 in our browserslist and everything will work fine.

Fyrd:
I use BrowserStack, which is a paid service that allows you to test on older browser versions. Sounds like the device you refer to is using a custom version of Chromium, which could explain the difference. Glad you have a workaround anyway!

PR #5420, opened 2020-05-10
Update notes on Public Key Pinning

atjn:
Some small updates to the notes now that the feature is pretty much completely unsuable.


Fyrd:
Thanks!

PR #5421, opened 2020-05-10
Create DNSSEC table

### LINKED ISSUE #2055: DNS-based Authentication of Named Entities (DANE)
Add support for DNS-based Authentication of Named Entities (DANE)?
- https://en.wikipedia.org/wiki/Domain_Name_System_Security_Extensions#DANE_Working_Group
- https://en.wikipedia.org/wiki/DNS-based_Authentication_of_Named_Entities

### LINKED ISSUE #3380: Add DNSSEC support
Add DNSSEC support

atjn:
Based on what i had read, it sounded like DNSSEC worked everywhere. I had a grim experience slowly figuring out that my code didn't work because DNSSEC is unsupported in literally every browser. So thought i would add it here.

_Techincally speaking_, all browsers should be marked `Unsupported`, but i really don't want to end up with developers glancing at the page and thinking "oh, i should not use this".
DNSSEC works just fine, it just works outside the browser. That is also why i'm a bit verbose in the notes. I really don't want to misguide developers.

I found testing to be inconsistent, so i can't really use that, but between limited testing and poking around in forums an bugs, i am pretty confident saying that no browser has ever really supported this. The only exception is Chrome which might've actually had it enabled in stable for a few versions, but i can't find any documentation on it. Note `#2` is loosely correct and is based on the info in [this article](https://www.imperialviolet.org/2011/06/16/dnssecchrome.html).

DANE is tightly coupled to DNSSEC, which is why i combined them. If they ever get real traction, they should probably be split into separate tables.

Closes #3380
Closes #2055


Fyrd:
Looks good, I appreciate the extra notes. 👍

drzraf:
DANE and DNSSEC shouldn't have merged (even less confused with DoH)

- DoH has been the trend for most browsers (so that the CA cartel and other intermediaries can keep living)
- DNSSEC is (maybe) supported by some browsers (at least [not FF](https://bugzilla.mozilla.org/show_bug.cgi?id=1609835))
- Sadly, **DANE** (which depends upon DNSSEC) isn't (no browser AFAIK: Neither [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1339991), not even [FF](https://bugzilla.mozilla.org/show_bug.cgi?id=672600) but [extension exists](https://addons.mozilla.org/en-US/firefox/addon/dnssec-dane-validator/)). It **should be given it's own comparison page**.

Obviously, Google & co will keep arguing clients need to rely on big corporations' complex CT, OCSP (or DoH) intermediary servers rather than validating straight against the destination website (+ its registrar)
The situation is thus unlikely to change in the short-term.

atjn:
Hello @dzraf

> DANE and DNSSEC shouldn't have merged

I don't see any argument as to why they should not be merged. Can you be more specific?

You mention that there are extensions that do DANE verification, but you also mention that in order to do DANE verification, you need to first do DNSSEC verification. Thus, if you want to add the extensions to this table, there is no need to create separate tables.

> (even less confused with DoH)

DoH is never mentioned in the support table. I am not sure what you mean.

> DNSSEC is (maybe) supported by some browsers

Do you have any documentation that shows that a browser supports DNSSEC? I can't find any.

drzraf:
DNSSEC doesn't provide the same feature as DANE. And showing DANE as supported while only DNSSEC is, is at best confusing. The table about DANE would show show ❌ for almost every browsers contrary to one about DNSSEC.

PR #5422, opened 2020-05-10
Fix support data for Encrypted Media Extensions

### LINKED ISSUE #4147: Safari on iOS supports EME
Safari supports Encrypted Media Extensions (EME) as of iOS 11.2.

Currently caniuse site states that EME is not supported.

atjn:
IOS Safari has had support for this spec for a few years. It's in their [changelog](https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html) and i confirmed it with the [caniuse test](https://tests.caniuse.com/?feat=eme).

EME is now a [W3C recommendation](https://www.w3.org/TR/encrypted-media/) with a new URL to boot, and all browser support it - awesome 🥳

Closes #4147


Fyrd:
Great, thanks! Not sure how I missed that issue.

PR #5425, opened 2020-05-13
Chrome adds full support for CSS appearance

atjn:
In #5412 i completely failed to notice that Chromium is also adding support for the `auto` keyword in v83, making this a fully supported, unprefixed feature 🥳

I ran a manuel test to confirm it. Here is the reference:
https://www.chromestatus.com/feature/5913213940006912

They are also doing some other stuff with it, but nothing that i felt needed notes:
https://www.chromestatus.com/features#appearance (the first four entries)

### DIFF
--- a/features-json/css-appearance.json
+++ b/features-json/css-appearance.json
@@ -197,9 +197,9 @@
       "79":"a x #1",
       "80":"a x #1",
       "81":"a x #1",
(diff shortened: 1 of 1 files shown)


Fyrd:
Nice, thanks!

PR #5426, opened 2020-05-13
Samsung internet supports WebUSB

atjn:
Samsung Internet is based on Chromium which has had support for WebUSB since v61 ([SamsInt v8.0](https://github.com/mdn/browser-compat-data/blob/master/browsers/samsunginternet_android.json#:~:text=8.0)).

I have confirmed support with the [Caniuse test](https://tests.caniuse.com/?feat=webusb) on Samsung Internet 9.0 and 11.1

https://www.chromestatus.com/feature/5651917954875392


atjn:
Huh, i've never experienced that before 😄

Fyrd:
Validation passed on retry. :) Thanks!

PR #5427, opened 2020-05-13
Update notes on web cryptography

atjn:
I have removed/edited a lot of notes from Web Crypto. They probably made sense when they were written, but are inaccurate or confusing today:

The spec is now a W3C Recommendation

Firefox removed their extra unofficial features before they enabled crypto by default.

Only enabling web crypto over secure connections is part of the spec and not unique to Chrome.


atjn:
Hmm, seems like Travis and Github don't like each other right now.

Fyrd:
Good updates, thanks!

PR #5428, opened 2020-05-13
Content Policy Level 2 is now a W3C Recommendation

atjn:
https://www.w3.org/TR/CSP2/

Let's see if Travis fails on this one too 🙃


Fyrd:
Thanks!

PR #5432, opened 2020-05-15
Update css-read-only-write.json

mattbrundage:
Unprefixed in Firefox yesterday: https://bugzilla.mozilla.org/show_bug.cgi?id=312971


Fyrd:
Thanks!

Issue #3518, opened 2017-06-15
Older versions of Chrome on Android?


dandv:
Just curious about the status of Android Chrome versions < 59. I see [here](https://github.com/Fyrd/caniuse/issues/3308) that caniuse had Chrome 57 displayed at one point. Other browsers show a significant number of past versions, while Chrome on Android shows none.

Fyrd:
Written about this before, but I can't find the comment. My mistake for failing to put it in a centralized place!

So here's the reasoning:
1. I prefer not to include browser versions that I'm unable to test, and while old desktop versions are fairly easy to come by, old mobile versions are not (old iOS versions are on browserstack, but not chrome)
2. A large percentage of features are supported on Chrome for desktop anyway, so for historical information that tends to match up. For occasional exceptions notes can be added.
3. I don't currently have an easy way to get browser usage of different Chrome for Android versions, so that would wind up being inaccurate.
4. The site is currently set up to load all data in one go, so adding a ton more versions is not good for that. I'm actually in the process of changing this to better support larger amounts of data, so this argument will go away.

All of that said, I would still like to add them anyway at some point because the browser has huge usage and it'd be helpful to have that better covered. But that list's the reason they're not there yet.

dandv:
How about apkmirror? It's [safe](https://android.stackexchange.com/questions/135711/is-apkmirror-com-safe) and if you're paranoid, I can provide SHA-1's for binaries (I work at Google).

GitToTheHub:
Hi, i realized that the minimum Chrome and Android Browser Version is equal when you compare them on MDN. See https://github.com/Fyrd/caniuse/issues/5074 last comment.

jpmedley:
For the most part. There are a few edge cases where they are not. I've managed to keep Chrome engineers on top for this for about the last 30 or so versions of Chrome. For that reason you can rely on [Chrome Status](https://www.chromestatus.com/features) for differences. The most likely places they are to be different is when a web feature has an underlying device hardware dependency. I'm having trouble finding an example just now, but keep an eye out.

GitToTheHub:
Ok, thanks for the explanation. So it's ok to compare it on basic stuff but can differ in using some Hardware-APIs. I saw the Notifications API is enabled for Desktop but not for mobile, when i follow [your link from the other issue](https://cs.chromium.org/chromium/src/android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt). Thanks :)

[...]

PR #5433, opened 2020-05-18
Brotli: Remove note #2 for Android Browser

Schweinepriester:
> "2":"Enabled since 27 May 2016",

seems to have been applicable to [Chrome 50 only](https://github.com/Fyrd/caniuse/blob/master/features-json/brotli.json#L180) and since

- Android Browser 81 wasnt available in 2016
- Chrome for Android 81 doesnt have the note
- we dont have past versions of neither browser (kinda #3518 ;))
- its been 4 years

I think the note can be removed 0:-)


Fyrd:
Agreed, thanks!

PR #5434, opened 2020-05-18
AppCache removal delayed to Chrome 85

Schweinepriester:
Following up #5291:

<https://web.dev/appcache-removal/>

> Removal from secure contexts, except for those opted-in to the origin trial | Chrome 85

<https://bugs.chromium.org/p/chromium/issues/detail?id=582750#c69>

> appcache: bump deprecation warnings from M82 to M85

---

<https://bugs.chromium.org/p/chromium/issues/detail?id=582750#c73>

For WebView technically

> Because WebView apps can't use origin trials, AppCache will be allowed until the final removal in M90.  We still recommend that developers move off of AppCache as soon as possible.

but I don't think we should track => encourage that.


Fyrd:
Thanks!

PR #5435, opened 2020-05-19
Reword & include Safari 13 in backface-visibility prefix requirement note

Schweinepriester:
Added <https://bugs.webkit.org/show_bug.cgi?id=170983> as well.

Safari TP 106:

image

Safari on iOS 13.4.1:

A03A2C76-2075-46FF-BFBB-261C09A2D1E2


Fyrd:
Thanks!

PR #5438, opened 2020-05-22
flexbox-gap: Chrome 84+ by default

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=762679#c41>

> [css-flexbox] Enable gaps by default
>
> Enabling for chrome 84 will be at least one more finch CL.

<https://www.chromestatus.com/feature/6031643502444544>

~Draft pull request because its not yet available without the flag.~

Chrome Canary 85.0.4153.0:

image

---

Chrome Beta is still at 83 unfortunately:

image

### DIFF
--- a/features-json/flexbox-gap.json
+++ b/features-json/flexbox-gap.json
@@ -210,8 +210,8 @@
       "80":"n",
       "81":"n",
       "83":"n",
-      "84":"n d #1",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #5440, opened 2020-05-22
iOS 13 and macOS 10.15 Catalina support OpenType-CFF2 fonts

litherum:


Fyrd:
Not according to my tests. What is your source?

Fyrd:
Okay I see now you work on WebKit and fonts and you probably know what you're talking about 😅 Still would be nice to see a test case or something, I'm on 10.15.4 and tried Safari 13.1 and Safari TP and my test at http://tests.caniuse.com/?feat=variable-fonts that passes in Chrome & Firefox still seems to fail.

litherum:
macOS Catalina fails to open SourceSerifVariable-cff2.otf.woff. I've filed an internal bug about it: <rdar://problem/63636353>.

Other CFF2 fonts work, though. For example, if you try AdobeVFPrototype.otf from https://github.com/adobe-fonts/adobe-variable-font-prototype it works great.

Fyrd:
Ah, thanks! Have included that font to the test suite and indeed it passes.

PR #5443, opened 2020-05-26
Add support data for CSS color() function

### LINKED ISSUE #3134: CSS4 - Specifying profiled colors: the color() function
Found no data or issue when searching for this.
Afaik this differs from #2864, which primarily requests `color-mod()`.

This is intended for 'CSS Color Module Level 4' `color()`, allowing to specify profiled colors.

Spec: <https://drafts.csswg.org/css-color/#color-function>

Example:
```css
em {
  color: color(p3 50 250 100);
}
```

Shamelessly stolen from <https://developer.apple.com/library/prerelease/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_1.html>

Safari 10.1 (& TP) / iOS 10.3 support this, apparently.

atjn:
The CSS color function:
- Seems pretty useful
- Has a complicated support landscape
- Has recently received a lot of attention
- Has active implementation bugs in Chromium and Firefox

It's perfect for Caniuse 😉

I have tested support with the [WPT test](https://wpt.fyi/results/css/css-color/color-function-parsing.html). My results are consistent with the Chromium/Firefox implementation bugs and the [Safari 10.1 release notes](https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_1.html).

Most resources link to the new and improved [editor's draft spec](https://drafts.csswg.org/css-color/#color-function), but my limited testing suggests that at least the current Safari implementation only supports the original [working draft spec](https://www.w3.org/TR/css-color-4/#color-function), so that is what i have linked to 🤷‍♀️

Closes #3134


Fyrd:
Looks good! I think however that if Safari indeed only supports the older version of the spec its support should be changed to partial (with note), and the link should be to the latest version.

@tabatkins would you happen to know if the latest spec changes are incompatible with Safari's current support?

atjn:
Hmm i'm not sure i agree on that.
The new spec just adds a few extra tidbits like support for percentage-values, and is fully backwards-compatible with the old spec. I don't think that warrants a "partial support", especially when no other browsers support the new spec yet. If other browsers implement the new spec, it would probably make sense to add a note.

I'm also not sure why you want me to link to the editor's draft. When a working draft is available, i would assume that that is what developers should be using, and then the editor's draft will be merged into the working draft when it is deemed ready for wide adoption. Am i wrong about that?

Fyrd:
Ah, I misunderstood then, thought you meant your testing indicated that Safari doesn't support the feature at all according to the latest spec. If the changes are backward-compatible then yes I agree.

As for the spec to link to, I've been previously informed that the editor's draft is the one that implementers and developers should look at when it's still in development, as the published versions just represent older snapshots of the spec. I would however keep the status at "working draft" to indicate the status the specification in general has reached.

atjn:
I see, i'll try to be more clear on that next time 😄

I guess it sort of makes sense to always use the latest spec in development. The link is updated, this feature should be good to go 😊

Fyrd:
Great, thanks!

PR #5446, opened 2020-05-29
Firefox will soon support Link Preload

atjn:
Firefox is planning support by default for link preloading:

https://groups.google.com/forum/#!msg/mozilla.dev.platform/MShUZ4VTa6s/E4XFYfL5AQA
> ** Release plan **
> - I'm about to enable it in this cycle, Firefox 78, for all users of
> Nightly and then let it ride to Early Beta only.
> - After PI and PE is done with good results we plan to do a gradual
> release on Beta and Release using a study. 

This PR adds a note about it. I will also open a draft PR for the final changes when we know which exact version will be adding support.

### DIFF
--- a/features-json/link-rel-preload.json
+++ b/features-json/link-rel-preload.json
@@ -382,7 +382,7 @@
       "2.5":"n"
     }
   },
-  "notes":"Firefox plans to enable this by default in a version after 78",
+  "notes":"Firefox has newly rewritten their implementation and plans to enable it by default in a version closely succeeding 78",
   "notes_by_num":{
     "1":"Only cachable resources can be preloaded. This includes the following `as` values: script, style, image, video, audio, track, fetch, and font (note font/collection is not supported).",
     "2":"Can be enabled via the \"Experimental Features\" developer menu",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
While I absolutely share your enthusiasm, I'm -1 on this one, because imho the flag already signals enough that this feature is in development and coming plus the vague wording plus that this feature was available behind a flag for 20(!) versions.

Also seeing the bigger picture, we probably could add a note like this to most features as they will probably/hopefully come to every browser in active in development, but that would be of little benefit.

atjn:
Hmm i see where you're coming from @Schweinepriester.
Here's how i see it: Firefox is starting stable testing now and has actually committed to enabling this in stable in a version closely succeeding 78. In my opinion, that is enough to differentiate it from a "were working on it" flag, and i think developers that are making decisions for future applications would like to know about it now, rather than in two months when FF 79 is out and we update the support table.

I'll let it be up to @Fyrd to decide 🙂

Fyrd:
I see the note linked to mentions "This feature has been newly written in Gecko and replaces a preexisting implementation" so I think that makes it reasonable to add a note for. In fact I think it would be good to rephrase it pointing out that it's a new implementation so folks will understand it's not just stating the obvious for a feature behind a flag. :)

atjn:
Allright, is this good?

Fyrd:
Yup, looks good to me! Thanks.

PR #5447, opened 2020-05-29
Firefox supports Link preload

atjn:
Building on #5446 

When Firefox enables support by default, i will update this PR for merging.


dkniffin:
@atjn I believe Firefox now supports this by default. Here's the bug tracker: https://bugzilla.mozilla.org/show_bug.cgi?id=1626997 The last comment mentions that it wasn't working for that user, however, I just tried it in my Firefox 84 and it seems to be working 🤷‍♂️

atjn:
Thanks for the heads up @dkniffin :)
Just to be clear, is `network.preload` set to `true` by default in your `about:config` settings?

The feature was not enabled on any of the 5 Firefox 84 installs i just checked.

dkniffin:
Yep:

Screen Shot 2021-01-11 at 9 47 55 AM

Not sure that it makes a difference, but I'm on MacOS Catalina (10.15.6). My firefox version is 84.0.2 (64-bit).

atjn:
Awesome, thanks for the insight! As far as I can tell, v84 is running an experiment that enables the feature on a small percentage of browsers. It seems like it is scheduled to be fully enabled in v86, but they're debating whether they should change that to v85.

I'll wait a few more days and see where this goes. I don't think it makes sense to call v84 `supported` when it is such small subset of users :)

atjn:
As of [comment 33](https://bugzilla.mozilla.org/show_bug.cgi?id=1626997#c33), it seems that preload has been accepted for release in v85. I tried to verify this with the [caniuse test](https://tests.caniuse.com/link-rel-preload), but it seems to be broken, and never detects support in any browser.

I have instead written my own tests, and have confirmed support in v85 on Mac, Windows 10 and Ubuntu with both preload methods (HTML `link` attribute and HTTP `link` header).

Fyrd:
Thanks! Have fixed the test and can confirm.

PR #5449, opened 2020-05-29
iOS 13.4 has probably the same support for `display: contents` as Safari 13.1

Schweinepriester:
TP was changed in #5232 and the mentioned bug <https://bugs.webkit.org/show_bug.cgi?id=188259> is indeed fixed.

However, note 2 refers to <https://hiddedevries.nl/en/blog/2018-04-21-more-accessible-markup-with-display-contents> which in return (only) refers to <https://bugs.webkit.org/show_bug.cgi?id=185679> which isnt fixed. There is a testcase available <https://codepen.io/hidde/pen/gzbMeL>, I'm not sure what is the correct state though.

I think its highly likely iOS 13.4 has the same support as Safari 13.1, but not sure how to properly test that. Or maybe desktop Safari still requires note 2 and partial support...?


Fyrd:
Sounds right to me!

PR #5450, opened 2020-05-30
ServiceWorkers are now a Candidate Recommendation

atjn:
Most of the SW spec is now a candidate recommendation: https://www.w3.org/TR/service-workers/ 🥳

I also updated the link, since it's moved.


Fyrd:
Thanks!

PR #5451, opened 2020-05-30
Edge supports Scroll-To-Text fragment

atjn:
As expected, Edge 83 supports Scroll-To-Text Fragment by default 🥳
This was tested manually.

I also added a few Chromium versions that had support behind flags. Not sure why i missed that originally.  In my testing, Opera's implementation didn't work until v66, so that's why it doesn't have flag support going back to v61/62.

Is there any reason this is not live on the site? If there's anything i can help with, let me know :)


Fyrd:
Thanks! No real reason, just haven't reviewed it yet. Looks pretty straightforward so will take a look soon.

atjn:
Awesome!

PR #5452, opened 2020-05-30
Safari adds experimental support for Link Prefetch

atjn:
I have tested this with the [Caniuse test](https://tests.caniuse.com/link-rel-prefetch). Please note that i only have access to Safari 12.1 and 13.1, as well as Safari IOS 13.5. Any data outside of those exact versions are just solid guesses.


Fyrd:
Thanks!

PR #5453, opened 2020-05-31
Small updates to Native Filesystem API

atjn:
The Native filesystem API is supported behind a flag on all desktop Chromium browsers. I was also able to test that it originally became available in Chrome 74 / Opera 61.

Chrome for Android has never had support for it.


Fyrd:
Thanks!

Fyrd:
Oh btw, you'll want the ` character (backtick) instead of the ´ character for code spans.

atjn:
Oh wow, well done me 🤦‍♂️

PR #5454, opened 2020-05-31
Fix WebVR data

atjn:
The WebVR data wasn't quite up to date. I have corrected a bunch of data on desktop Chromium support behind flags, and removed the comment since they are completely deprecated by now.

I also tested Chrome on Android with the [Caniuse test](https://tests.caniuse.com/webvr) and could not confirm support in v83. The tracked version right now is v81, but that should change to 83 pretty soon.


Fyrd:
Thanks!

PR #5455, opened 2020-05-31
Opera supports WebXR

atjn:
Opera supports WebXR. It makes perfect sense, as it is based on Chromium. I have also confirmed it with the [Caniuse test](https://tests.caniuse.com/webxr). (and along the way, i found out the flag for webxr was actually added in chromium 65, not 66).


Fyrd:
Thanks!

PR #5457, opened 2020-05-31
prefers-reduced-motion is now a Working Draft

atjn:
https://www.w3.org/TR/mediaqueries-5/#prefers-reduced-motion


Fyrd:
Thanks!

PR #5460, opened 2020-06-02
Add Chrome/Edge support note for CSS "image-orientation"

ingmarh:
In Chrome and Edge, the image’s aspect ratio is not maintained when
`object-fit: cover` and `image-orientation: from-image` are used
together:

https://bugs.chromium.org/p/chromium/issues/detail?id=1082669


Fyrd:
Thanks!

PR #5463, opened 2020-06-03
Remove dead SRI hash webpage scanner link

andreas-venturini:


Fyrd:
Not just dead I see...yikes. Thanks for the update!

PR #5464, opened 2020-06-04
Feature now working on Samsung mobile browser

Mathieute:
On 11.2 Samsung Browser


Fyrd:
Thanks! Just tested and it looks like it's in 11.1 as well. As both are based on Chrome 75 I've combined both versions into one and updated manually.

PR #5469, opened 2020-06-06
Update js-regexp-lookbehind.json

blama:
Support for Regexp-lookbehind added to Firefox 78 as per https://hacks.mozilla.org/2020/06/a-new-regexp-engine-in-spidermonkey/


Fyrd:
Thanks!

PR #5470, opened 2020-06-06
General updates to Media Source Extensions

atjn:
Some general updates to MSE:
- Spec is now a W3C Recommendation
- Removed broken links
- Removed old notes on partial support
- Added support in Samsung Internet 9+

I was able to test and confirm support in Samsung Internet all the way back to v9 ([test 1](https://tests.caniuse.com/mediasource), [test 2](https://html5test.com)).
#4406 shows no support up until v7.2, which means only v8 might be wrong 🤷‍♀️


Fyrd:
Looks good, thanks!

PR #5474, opened 2020-06-08
Update web-share.json for Samsung Internet

sladex:
I've tested this feature in Samsung Internet 11.1 and 11.2 and it worked.
Also mdn claims it has been supported since 8.0 https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share
Also Samsung Internet uses Chromium engine under the hood and 8.0 updated to Chromium 63 https://developer.samsung.com/internet/release-note.html


Fyrd:
Thanks! Have just updated this one manually (see #5464)

PR #5476, opened 2020-06-08
Target A2HS instead of Web App Manifest

### LINKED ISSUE #5099: Add A2HS
Please add A2HS to caniuse

### LINKED ISSUE #5448: Rework Web App Manifest data
The support data for [WAM](https://caniuse.com/#feat=web-app-manifest) (the Web App Manifest) seems to mostly be tracking support for [A2HS](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Add_to_home_screen) (Add To Home Screen functionality).
I can see why this has happened, since WAM is pretty much only used for A2HS, but they are technically speaking two different things, and IMO it is really confusing for developers that the feature is not called A2HS when that is what it is tracking.

Another thing to note is that searching for things like `A2HS`, `add to home screen` and `progressive web app` does not return any data.

I suggest that we do one of two things about it:
1. Change the title, description and keywords of the existing WAM data to A2HS (i'm not sure how you would change the URL link without breaking existing links to the feature?)
2. Add a new feature for A2HS and migrate some of the notes from WAM to A2HS

I don't mind doing the legwork for these changes, but i felt we needed to discuss the best solution before i start making any code changes 😊

Related: #5099 https://github.com/Fyrd/caniuse/issues/1532#issuecomment-168075482

atjn:
Discussion on this PR here: #5448

I have decided to replace all `partial support` with `full support`. The `appinstalled` and `beforeinstallprompt` events are not central to the experience, and aren't used by any webapps i know of. They have also been removed from the core spec for now. They're still cool though and definetely deserve a note.

Closes #5448 
Closes #5099


Fyrd:
Looks good, one suggestion I have is to include at least a mention of the web app manifest in the description to help make it clear that "web app manifest" (as mentioned in the spec) refers to the same feature.

atjn:
Yeah, that makes sense to me. Does this look good?

Fyrd:
Yup, looks good, thanks!

PR #5479, opened 2020-06-11
Update calc.json

mattbrundage:
In Android and iOS, Opera Mini supports calc() by default. When the user sets the data savings mode to "Extreme," the rendering engine switches back to Presto, which does not support calc().
https://dev.opera.com/articles/browsers-modes-engines/

### DIFF
--- a/features-json/calc.json
+++ b/features-json/calc.json
@@ -360,7 +360,7 @@
       "13.4-13.5":"y"
     },
     "op_mini":{
-      "all":"n"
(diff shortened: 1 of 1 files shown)


Fyrd:
Opera Mini support always refers to the "extreme data savings mode" or more accurately the java version of Opera Mini used heavily on feature phones in some countries, which is what accounts for most of its usage. There's a note on this when hovering over the "Opera Mini" header in feature tables.

PR #5480, opened 2020-06-13
Update some MSDN links which redirect to MDN

weibangtuo:


Fyrd:
Great updates, thanks!

PR #5487, opened 2020-06-23
TLS 1.3: Remove note for Safari 14 & TP

Schweinepriester:
Speculative but given the history highly likely that Safari 14 wont be available below macOS 10.14 Mojave: <https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility>

I'm fine with leaving this open for some time.


Fyrd:
Seems a safe assumption. I'll go ahead and merge.

PR #5491, opened 2020-06-24
WebHID is available in Opera behind the same flag

ontracker:
WebHID is available behind the same flag in Chromium based Opera (tested on Windows version 66, 67, 68). It seems slightly broken in 69.0.3686.36 right now (no confirm button when selecting device)


ontracker:
```
navigator.hid
HID {onconnect: null, ondisconnect: null}
    onconnect: null
    ondisconnect: null
    __proto__: HID
```

Fyrd:
Thanks!

PR #5493, opened 2020-06-25
Remove old note on Safari support for WebP

atjn:
Now that Safari is adding support (🥳), i don't see any reason to link to a four year old speculative article.


Fyrd:
Agreed, thanks!

PR #5494, opened 2020-06-25
Safari does not support A2HS in WebViews

atjn:
It just occured to me that Safari on iOS doesn't support A2HS in WebViews. In most browsers, that's not a big deal, but in Safari it means no third party browser can support the feature, leaving a large chunk of `saf_ios` traffic without support. I think it should be marked `Partial support` because of this.

### DIFF
--- a/features-json/web-app-manifest.json
+++ b/features-json/web-app-manifest.json
@@ -399,7 +399,7 @@
     "1":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported.",
     "2":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported on all platforms.",
     "3":"A manifest could be used to list apps in the Microsoft Store, which would use Edge as a back-end.",
-    "4":"Safari on IOS does not support A2HS in WebViews like Chrome and Firefox"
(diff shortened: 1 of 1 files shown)


features-json/web-app-manifest.json
```diff
@@ -398,7 +398,8 @@
   "notes_by_num":{
     "1":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported.",
     "2":"The installation events, `appinstalled` and `beforeinstallprompt`, are not supported on all platforms.",
-    "3":"A manifest could be used to list apps in the Microsoft Store, which would use Edge as a back-end."
+    "3":"A manifest could be used to list apps in the Microsoft Store, which would use Edge as a back-end.",
+    "4":"Safari on IOS does not support A2HS in WebViews like Chrome and Firefox"
```
Schweinepriester, line 402: tiny improvement:
```suggestion
    "4":"Safari on iOS does not support A2HS in WebViews like Chrome and Firefox"
```

0:-)

https://en.wikipedia.org/wiki/Cisco_IOS :P

atjn: Fair 😄

Fyrd:
Yeah, makes sense. Thanks!

PR #5498, opened 2020-06-26
Chrome 86: AVIF enabled by default

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=960620#c162>

> Enable the AVIF image format by default

Draft PR to wait for Canary to come around to be able to test using <https://tests.caniuse.com/?feat=avif> to check if it really gets enabled far and wide.


Schweinepriester:
Absolute mad lads:

image

85.0.4181.9, which moved from canary to dev, doesnt have it (enabled?) though.

Fyrd:
Thanks!

PR #5499, opened 2020-06-26
Firefox Android & Samsung Browser supports prefers-reduced-motion media query

atjn:
The newest versions of Firefox for Android and Samsung Browser supports the `prefers-reduced-motion` media query.

This is tested manually with the [caniuse test](https://tests.caniuse.com/prefers-reduced-motion), and it makes sense when you look at what engine version they're based on.

Also just removed a comment that isn't used and added the Chrome ID.


Fyrd:
Thanks!

PR #5500, opened 2020-06-26
Firefox 78+ drops TLS 1.1

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/78#TLS_1.0_and_1.1_removal>

<https://clienttest.ssllabs.com:8443/ssltest/viewMyClient.html>:

image


Fyrd:
Thanks!

PR #5504, opened 2020-06-29
Major update to CSS :matches() -> :is()

atjn:
The css `:maches()` spec has been renamed to `:is()`, and `:maches()` has become a deprecated alias.

[Firefox 78](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/78#CSS) and [Safari 14](https://developer.apple.com/documentation/safari-release-notes/safari-14-beta-release-notes#CSS) are enabling `:is()` by default, and Chromium will probably follow suit in ~v87.

I have manually verified it all with [this quick test suite](https://codepen.io/atjn/full/MWKErBe).

I have made a million small tweaks - let me know if there are any specific changes you would like a comment on :)


Fyrd:
Looking good! Had this one on my list to tackle at some point, I appreciate you taking it on. :)

Issue #5505, opened 2020-06-29
Samsung Internet 12 is out-ish


atjn:
I'm not sure this is completely official yet, but my phones just updated to Samsung Internet 12.0.1.47 on the stable channel.

As expected from a browser based on Chromium 79, i have verified support for:
-  [WebXR](https://caniuse.com/#feat=webxr)
- [CSS min() max() clamp()](https://caniuse.com/#feat=css-math-functions)
- [prefers-color-scheme](https://caniuse.com/#feat=prefers-color-scheme)

I've made you a [PR](https://github.com/Fyrd/caniuse/pull/5506) for it. If it's easier for you to just make the changes yourself, you can just close it :)

PR #5506, opened 2020-06-29
Samsung 12 supports CSS min() max() clamp(), WebXR and prefers-color-scheme

atjn:
PR following #5505

As expected, samsung internet 12 has added a bunch of awesome new stuff from the new Chromium build.

I've also made a few getting-things-up-to-date changes.


Fyrd:
Thanks!

PR #5511, opened 2020-07-01
avif.json: Update AVIF support in Chrome 85

EwoutH:
The AVIF image format will be enabled by default thanks to this cherry picked commit: 
https://chromium-review.googlesource.com/c/chromium/src/+/2276542


Fyrd:
Thanks!

PR #5512, opened 2020-07-01
Safari 14: Text Fragments?

Schweinepriester:
<https://twitter.com/tomayac/status/1278356832604618752>

=> <https://www.youtube.com/watch?v=DDU9T8_ZvMo>

Further testing using e.g. <https://tests.caniuse.com/?feat=url-scroll-to-text-fragment> would be appropriate to check iOS and because @tomayac mentions limitions, so we maybe have to change it to partial support 0:-)


tomayac:
Just to clarify: this works only thanks to a (still incomplete) polyfill that the extension injects.

tomayac:
This PR can be closed for the time being.

Schweinepriester:
Whoops 🤪

PR #5513, opened 2020-07-02
Update date-tolocaledatestring.json IE11 compat

morille:
Date.prototype.toLocaleDateString is fully supported in ie11
Following tests made from the IE11 devtools:

var a = new Date();
a.toLocaleDateString("en");
// "‎7‎/‎2‎/‎2020"
a.toLocaleDateString("en", {"minute": "2-digit", "hour": "2-digit"});
// "‎7‎/‎2‎/‎2020‎ ‎09‎:‎56‎ ‎AM"
a.toLocaleDateString("fr");
// "‎02‎/‎07‎/‎2020"
a.toLocaleDateString("fr", {"minute": "2-digit", "hour": "2-digit"});
// "‎02‎/‎07‎/‎2020‎ ‎09‎:‎56"
a.toLocaleDateString("ar-EG");
// "‏٠٢‏/‏٠٧‏/‏٢٠٢٠"


Fyrd:
Thanks for pointing this out! I checked my tests which were still failing in IE11, turns out the tests were running into this issue which I didn't know about:
https://www.csgpro.com/blog/2016/08/a-bad-date-with-internet-explorer-11-trouble-with-new-unicode-characters-in-javascript-date-strings/
So I have updated the tests and added that as a bug. I also noticed there's still issues with some locales as well and the timeZone option still fails so I have updated it accordingly (still partial but different notes apply). Thanks again!

PR #5514, opened 2020-07-03
Safari 14: Clarified SameSite support across macOS versions

kiding:
- Added `#4` to 13.1: Should be the same as 13
  - Regarding 14:
    - Support for macOS 10.13 High Sierra is dropped; `#4` removed
    - Support for macOS 10.14 Mojave is _not_ dropped; `#5` holds
    - Related to #5487
  - Removed `#5` from TP: Support for 10.14 is dropped


Fyrd:
Thanks!

PR #5522, opened 2020-07-08
Bulk update links to HTTPS

atjn:
There is a surprising number of insecure external links in the database. This PR fixes a lot of them, but not all of them. I have tried to make sure that the most used links are all secure now.

All links are tested for TLS support before changing them to HTTPS.

I have for the most part not evaluated if the links should even still be there, i have simply upgraded them to HTTPS :)


Fyrd:
Great work, thanks! And yeah, up until say five years ago it didn't seem terribly important for sites to use https so many of the older data would have links like that. Nice to have it updated!

PR #5523, opened 2020-07-08
General updates to AV1, AVIF, WebM, WebP

atjn:
Just some general usability updates to the media features.

- Removed a bunch of broken / no longer relevant links and notes
- Tweaked some notes and descriptions to, in my opinion, make them more useful for developers
- Plus a few other nits


features-json/avif.json
```diff
@@ -1,9 +1,13 @@
 {
   "title":"AVIF image format",
-  "description":"Image format based on the AV1 video format.",
+  "description":"A modern image format based on the AV1 video format. AVIF has better compression than [WebP](#feat=webp), JPEG, PNG and GIF and is designed to supersede them.",
```
Schweinepriester, line 3:
```suggestion
  "description":"A new image format based on the AV1 video format. AVIF has better compression than [WebP](#feat=webp), JPEG, PNG and GIF in most cases.",
```

features-json/webp.json
```diff
@@ -1,32 +1,24 @@
 {
   "title":"WebP image format",
-  "description":"Image format that supports lossy and lossless compression, as well as animation and alpha transparency.",
+  "description":"Image format (based on the VP8 video format) that supports lossy and lossless compression, as well as animation and alpha transparency. WebP has better compression than JPEG, PNG and GIF and is designed to supersede them. [AVIF](#feat=avif) is designed to supersede WebP.",
```
Schweinepriester, line 3:
```suggestion
  "description":"Image format (based on the VP8 video format) that supports lossy and lossless compression, as well as animation and alpha transparency. WebP has better compression than JPEG, PNG and GIF in most cases.",
```

atjn:
The goal of these changes is to help programmers who don't know a lot about these codecs. If i were researching this for the first time, i would really like a quick note letting me know that this codec is better/worse than these other options. For 99% of developers, this note will be true, and for those who are advanced enough that this statement might not be true, i am pretty sure they've already done their own research and have their own opinion on it.
So i generally don't see any issue with these changes.

I don't think there's anything controversial in saying that AVIF and WebP were designed to supersede older formats. The first section on WebP's official website directly calls WebP a "superior format", and then compares it directly to PNG and JPEG.

I also believe it is almost a fact that WebP and AVIF are better than PNG, GIF and JPEG. For one, their compression implementations are objectively better than the older formats. In [this video](https://www.youtube.com/watch?v=F1kYBnY6mwg), Jake Archibald from Google straight up says that "WebP is just better", and also explains why. Here is [an article from Netflix](https://netflixtechblog.com/avif-for-next-generation-image-coding-b1d75675fe4), where they ran a bunch of compression test between JPEG, WebP and AVIF. JPEG was inferior in every single test. There is also [this test dataset](https://docs.google.com/spreadsheets/d/1bG-CSTn_L79MtS7A-wR572H5_W_3kb2xX5AWagKn1ak/edit?usp=sharing) that runs a bunch of lossless compression tests. PNG is inferior to AVIF and WebP on every single image. All of this is consistent with the tests i have made myself, and the countless other articles i have read over the last couple of years. WebP and AVIF are just better.

When it comes to AVIF being better than WebP, it's slightly less clear-cut. AVIF is generally speaking still better. In [the aforementioned test from Netflix](https://netflixtechblog.com/avif-for-next-generation-image-coding-b1d75675fe4), AVIF beats WebP in every single test, and in the [aforementioned lossless dataset](https://docs.google.com/spreadsheets/d/1bG-CSTn_L79MtS7A-wR572H5_W_3kb2xX5AWagKn1ak/edit?usp=sharing), AVIF does lose to WebP once, but only by a small margin, and it is still significantly better than WebP when combining all scores.
If AVIF ends up with poor support, i plan on revisiting the notes in the WebP description.

As for documentation on caniuse, the Netflix tests are already linked on the AVIF table, and in the WebP article, i have added an article that mentions how much better WebP is, and links to a study made by Google.

In conclusion
Wile i believe your changes remove the purpose of adding the notes in the first place, i do agree that we probably could do more to hint that there's more to the story than this short note. I propose just adding a "generally":

> AVIF generally has better compression than WebP, JPEG, PNG and GIF and is designed to supersede them.

> WebP generally has better compression than JPEG, PNG and GIF and is designed to supersede them. AVIF is designed to supersede WebP.

Fyrd:
Look like some good changes! I approve of the switch to "generally" to be more accurate there and it doesn't seem biased to me to point that the formats were _designed_ to supersede existing formats (whether or not they will will remain to be seen). That information could indeed be valuable to developers.
So I'll merge this as is, happy to hear from @Schweinepriester if he has any follow-up comments.

Schweinepriester:
I like "generally" as well, better than my suggestion.

If you are fine with it, I'm fine, too.

I've just shifted my hype a bit more towards neutral because of a [previous conversation on twitter](https://twitter.com/FidonetRunes/status/1216873017999052802) or rather primarily that neither WebP nor AVIF seem to support progressive rendering (please prove me wrong!), which might be perceived faster in a web context:

<https://blog.cloudflare.com/parallel-streaming-of-progressive-images/>

> The benefits of progressive rendering are unique to JPEG (supported in all browsers) and JPEG 2000 (supported in Safari). GIF and PNG have interlaced modes, but these modes come at a cost of worse compression. WebP doesn't even support progressive rendering at all. This creates a dilemma: WebP is usually 20%-30% smaller than a JPEG of equivalent quality, but progressive JPEG *appears* to load 50% faster. There are next-generation image formats that support progressive rendering better than JPEG, and compress better than WebP, but they're not supported in web browsers yet. In the meantime you can choose between the bandwidth savings of WebP or the better perceived performance of progressive JPEG by changing Polish settings in your Cloudflare dashboard.

For the record, I saw <https://developers.google.com/speed/webp/faq#does_webp_support_progressive_or_interlaced_display>

---

Also wondering which formats(!) cloudflare meant with

> There are next-generation image formats that support progressive rendering better than JPEG, and compress better than WebP, but they're not supported in web browsers yet.

FLIF? JPEG XL?

atjn:
Yes that is true! Progressive loading is awesome, and i really hope it will be supported in AVIF or "AVIF 2".

I do however think it is fair to put it in the "advanced users" category, which falls outside the scope of a caniuse description.
JPEG doesn't support lossy compression and alpha channel, so you need to know to specifically use it only for those scenarios.
Then it comes with it's own host of issues, like the fact that it uses a lot more CPU to load and still uses much more bandwidth compared to an AVIF image.You also need to use a tool that can generate the progressive images. Juggling all these factors around is not for basic users, who would probably be better off just using WebP/AVIF for everything. And even large players like Youtube has opted to deliver WebP images, and then just detect if you are on a slow connection, and deliver smaller images to you.

PR #5525, opened 2020-07-09
getUserMedia fixed in PWA mode on iOS 13.4+

gruhn:
getUserMedia support was introduced in Safari on iOS 11 but it was not
working in Safari driven UIWebView or WKWebView components in native
apps and also not in PWA mode, i.e. when a web page is added to the
home screen. For PWAs this has been fixed in iOS 13.4. Source:

https://bugs.webkit.org/show_bug.cgi?id=185448#c84

It's still not working in UIWebView/WKWebView though.


Fyrd:
Thanks!

PR #5526, opened 2020-07-10
Chrome 86: CSS `::marker` pseudo-element

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=457718#c132>

> [css-pseudo] Ship ::marker pseudo-element in M86
> 
> This patch enables the CSSMarkerPseudoElement flag by default in M86.
> 
> Intent to Ship:\
<https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/80j6AEWI-_s>

~Doesn't work in Canary 86.0.4197.0 yet, but it's only been an hour, so probably tomorrow :D~

Test: <https://tests.caniuse.com/?feat=css-marker-pseudo>

image


Fyrd:
Thanks!

PR #5529, opened 2020-07-14
Firefox 79: Return of the Shared Array Buffer?

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/79>

> [`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) has been re-enabled in a post-Spectre-safe manner. It is available to cross-origin isolated sites ([bug 1619649](https://bugzilla.mozilla.org/show_bug.cgi?id=1619649)).

Also see https://github.com/mdn/browser-compat-data/pull/6356.

However, <https://tests.caniuse.com/?feat=sharedarraybuffer> remains red for 79.0b7 as well as 80.0a1, so maybe we need partial contrary to BCD, to represent `cross-origin isolated sites` only …?


Schweinepriester:
They followed up:

> To cross-origin isolate your site, you need to set the new [`Cross-Origin-Embedder-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy "The HTTP Cross-Origin-Embedder-Policy (COEP) response header prevents a document from loading any cross-origin resources that don't explicitly grant the document permission (using CORP or CORS).") (COEP) and [`Cross-Origin-Opener-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy "The HTTP Cross-Origin-Opener-Policy (COOP) response header allows you to ensure a top-level document does not share a browsing context group with cross-origin documents.") (COOP) headers.

> ### HTTP
>
> -   Cross-origin isolation has been implemented using the new [`Cross-Origin-Embedder-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy "The HTTP Cross-Origin-Embedder-Policy (COEP) response header prevents a document from loading any cross-origin resources that don't explicitly grant the document permission (using CORP or CORS).") (COEP) and [`Cross-Origin-Opener-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy "The HTTP Cross-Origin-Opener-Policy (COOP) response header allows you to ensure a top-level document does not share a browsing context group with cross-origin documents.") (COOP) headers. This allows you to access powerful features such as [`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) objects and unthrottled timers in [`Performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now).

=>

- https://caniuse.com/#feat=mdn-http_headers_cross-origin-embedder-policy
- https://caniuse.com/#feat=mdn-http_headers_cross-origin-opener-policy

So, what do we want for Shared Array Buffer (and `Performance.now()`)? Partial support? I'm leaning towards `y` and a note, as those headers seem to be the new normal?

Fyrd:
@Schweinepriester Yeah, I think `y` and a note seems appropriate. Thanks!

akien-mga:
I was about to submit a duplicate PR for this :)

I'd suggest removing the `#2` footnote altogether, as it was only meant to indicate that SAB is only enabled by default in Nightly, which is no longer the case now that it made its way into Beta and will be in Firefox 79 Release.

kiding:
A link to this article would be great for context: [Mozilla Hacks - Safely reviving shared memory](https://hacks.mozilla.org/2020/07/safely-reviving-shared-memory/)

Schweinepriester:
Done I think, tried to incorporate all feedback :)

While we have

- https://caniuse.com/#feat=mdn-http_headers_cross-origin-embedder-policy
- https://caniuse.com/#feat=mdn-http_headers_cross-origin-opener-policy

already, we could make cross-origin isolation a feature on its own, like <https://chromestatus.com/feature/5953286387531776>, making it somewhat easier to refer to…

Fyrd:
Looking good, thanks all!

PR #5531, opened 2020-07-14
Adjust MediaRecorder for Safari >= 13.1

### LINKED ISSUE #5530: MediaRecorder support incorrect (Safari)
I believe the [MediaRecorder API](https://caniuse.com/#feat=mediarecorder) still isn't supported by Safari. You should be able to create a new instance of `MediaRecorder` in the console, but is undefined. It's currently only supported in the TP version, albeit behind a feature flag (does this count?)

Screenshot 2020-07-14 at 22 36 04

Schweinepriester:
Fixes #5530.

I've tested Safari 13.1.1 and a fresh Safari TP 109. 
Contrary to #4859 even TP requires the feature to be enabled (again?).

Related: #5203


Fyrd:
Thanks!

PR #5532, opened 2020-07-15
Add notes on specific browser handling of TLS 1.1

atjn:
I took the time to find a random website that only supported TLS 1.1/1.0 and check how all browsers respond to it. Turns out you can pretty quickly find one by looking through the [Qualys SSL Labs 'Recent Worst'](https://www.ssllabs.com/ssltest/#:~:text=Recent%20Worst) 😄

Changes for this PR:

Chrome and Firefox's new full-page warnings are hardly "unsupported" behaviour. The Chrome implementation is getting close, but Firefox's warning only shows on the very first TLS 1.1 load, and then never displays again. Furthermore, the "dismiss" button is directly accessible and highlighted, meaning most people will just click the button without reading the warning.
Because of this, i have decided to change `unsupported` versions to `partially supported`.

Chrome 84 (released last night) does not in implement the new full-page warning. That seems to be delayed for v85.

Added notes on the small UI warnings that Firefox and Safari has already implemented.

Chromium browsers also have small UI warnings in existing releases, but they are only shown if you click on the padlock (which shows no warnings). Also, i did not get reliable results when testing it. Therefore, i have just decided to not add a note on that behaviour.

### DIFF
--- a/features-json/tls1-1.json
+++ b/features-json/tls1-1.json
@@ -218,8 +218,8 @@
       "81":"y",
       "83":"y",
       "84":"y",
-      "85":"a #4 #5",
-      "86":"a #4 #5"
+      "85":"a #3 #4",
+      "86":"a #3 #4"
     },
     "safari":{
       "3.1":"n",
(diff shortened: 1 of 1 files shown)


atjn:
FYI:

Firefox 68:
image

Chrome 84 stable: (sorry could not get it in english)
image

Chrome 85 dev: (sorry could not get it in english)
image

Fyrd:
Looks good, thanks for the research! One suggestion I have is to combine note 4 & 5 since they both seems to refer to the same browsers?

atjn:
Hmm yeah, that does seem like a brainfart on my part :P
I would suggest grouping by feature instead of by browser. This way note 3 & 4 can be combined now, and then when Safari (probably) also shows a full page warning, that can be combined into note 5 (now note 4).

I will just make the commit now so you can merge instantly if you agree.

Fyrd:
Looks good, thanks!

PR #5543, opened 2020-07-19
Updating Feature Policy

atjn:
The stable Feature Policy spec has now [been renamed to Permissions Policy](https://w3c.github.io/webappsec-feature-policy/), and most of the experimental parts of the spec is going into a new [Document Policy spec](https://w3c.github.io/webappsec-feature-policy/document-policy.html) that is very similar to Permissions Policy.

The new headers use a new structuring method, meaning they are completely incompatible with Feature Policy. Many developers already use Feature Policy, especially with the `allow` attribute on iframes, and since both Feature- and Permissions Policy use the same attribute, developers will have to closely monitor the adoption/removal of the headers to decide when to move to the new syntax.
Because of this, i propose creating a new support table for both Permissions- and Document Policy, rather than trying to add them to the existing table. Doing this also has other benefits, like easier fragmentation of new vs old documentation, less complicated notes and more accurate feature id's.

Chrome is [adding a flag for Permissions-Policy in v86](https://chromestatus.com/feature/5745992911552512). A quick test with [timewarp.atjn.dk](https://timewarp.atjn.dk) (which only uses the new header) confirms their claim. I'm hoping to more easily test this with an update to [tests.caniuse.com](https://tests.caniuse.com/feature-policy), described in #5544 

Adding a still-brewing spec like Document Policy to caniuse is a bit optimistic, but i think it makes sense, since developers have already been looking at supporting these features, and will want to check support data on the new header. Some of these features are also implemented behind flags in Chrome already, so it's pretty reasonable to assume that it will recieve support at some point.

There's a million other changes in this PR, let me know if you want a more specific comment on anything :)


Fyrd:
Looking good to me! Thanks for tackling this one. :)

PR #5545, opened 2020-07-21
docs(cssrevertvalue): updated chromium based browsers

mfranzke:
updated accordingly to https://chromestatus.com/feature/5644990145363968


Fyrd:
This got updated manually, thanks. Note that you can't add versions via PR, these will be updated once the version becomes stable.

PR #5551, opened 2020-07-28
General updates to Trusted types

atjn:
As expected, most Chromium browsers now support Trusted Types. I've tested them all with the [caniuse test](https://tests.caniuse.com/trusted-types).
Android WebView 81 doesn't support it, but i'm guessing you will update the version to 84 fairly soon, so i just added support now.

I also added Firefox's position.


Fyrd:
Thanks!

PR #5555, opened 2020-08-02
Change rel=preload Firefox bug page to meta bug page

Macley-Kun:
Changed the link from the old solved bug to the meta bug page. As Firefox is still working on this, and rendering to the meta bug page makes it clearer for the user what the status is of this feature.


Fyrd:
Nice, thanks!

Issue #5561, opened 2020-08-06
Firefox for Android 79 is released


atjn:
Firefox is finally releasing their new android app, which runs on Gecko 79 🥳

It seems like they're doing a slow rollout, but given that i and many others ([check out all the bad reviews 😅](https://play.google.com/store/apps/details?id=org.mozilla.firefox&hl=en_US)) have recieved the update on the stable channel, i think it's safe to mark it current.

I also made a PR for the new features that Gecko 79 comes with: #5562

FYI: https://support.mozilla.org/en-US/kb/firefox-android-upgrade-faqs

atjn:
This is now live on the site

PR #5562, opened 2020-08-06
Firefox for Android 79 does a lot of new stuff

atjn:
Following the release of Firefox 79: #5561

I've verified all the changes that were expected when going from Gecko 68 to Gecko 79. All changes in this PR have been manually tested with the [caniuse test](https://tests.caniuse.com/), unless otherwise noted in the commit message.

I used a prototype of a new script i'm working on, which should ensure that i have tested every single feature that might have changed since Gecko 68. It might've missed a few features that are mobile-specific, but other than that, this list is complete.

There's also a few other changes in the list - for things i stumbled upon while testing Firefox.


Fyrd:
Looking good! Not seeing the update yet myself but once I do I'll review this.

atjn:
Just stumbled across a feature that was added to mobile, but isn't in Firefox desktop. So as i said, i might have missed a few mobile-specific features in this list.

atjn:
@Fyrd large parts of this PR have been reverted in b0ed49e

features-json/loading-lazy-attr.json
```diff
@@ -131,14 +131,14 @@
...
-      "80":"y",
-      "81":"y",
-      "82":"y"
+      "75":"a #1",
+      "76":"a #1",
+      "77":"a #1",
+      "78":"a #1",
+      "79":"a #1",
+      "80":"a #1",
+      "81":"a #1"
```
Fyrd, line 140: @atjn Looks like the missing comma here prevented the data in this file from being updated. Not sure if the validator missed it or if I didn't pay enough attention to it...

Fyrd:
@atjn Thanks for pointing that out! I believe I've fixed those parts now. Looks like it was mostly the `loading-lazy-attr` that was missing as well as a few links (when both the title and url changes for a link my updating code often treats it as a new link instead of an old one being updated, will need to fix that). If there's anything else specific I missed let me know.

features-json/loading-lazy-attr.json
atjn: I see, thanks for checking that out

atjn:
Awesome, thanks! As far as i can tell, you got everything :)

PR #5569, opened 2020-08-18
WebAuthn: FIDO2 devices do not work with iOS 13 if a FIDO2 pin is set

FozzieHi:
This has been fixed in iOS 14, but I think it qualifies for "partial support" here.


FozzieHi:
I also don't think there's support for Face ID in iOS 13 but I'm not 100% sure.

Fyrd:
Thanks!

PR #5572, opened 2020-08-20
Updates to Document policy

atjn:
There has been enough changes that i think it's time for an update on Document Policy.

Chrome, as usual, missed the real release version by a few points on chromestatus:
https://www.chromestatus.com/feature/5756689661820928
https://www.chromestatus.com/feature/5744681033924608

The header is already enabled by default in Chrome 84, but it has no policies, so it can only make console errors on malformed headers.

In Chrome 85+ i have tested and confirmed that setting `Document-Policy: force-load-at-top` will force the browser to ignore any text fragment in the URL, and just load the page at the top instead.


Fyrd:
Looking good, thanks!

PR #5582, opened 2020-08-26
Chromium 86 supports Native Filesystem API

atjn:
Chromium has landed support for the native filesystem API in v86, significantly closing the gap between native and web capabilities on the desktop 🥳

In trying not to fly too close to the sun, they have decided to only roll out some functionality, and then add to it in waves. Because of that, i have marked it as `partial` support, and will add more specifics as new functions come out.
This is the samme approach as [WebXR](https://caniuse.com/#feat=webxr), which is probably due for an update pretty soon :)

I have verified basic support in Chrome 86 with [the Caniuse test](https://tests.caniuse.com/native-filesystem-api) and [this cool demo](https://googlechromelabs.github.io/text-editor/). You can also check:
https://chromestatus.com/feature/6284708426022912
https://groups.google.com/a/chromium.org/g/blink-dev/c/9Fcpl2KVfbk/m/d9VN0Fz8BwAJ

### DIFF
--- a/features-json/native-filesystem-api.json
+++ b/features-json/native-filesystem-api.json
@@ -214,9 +214,9 @@
       "83":"n d #1",
       "84":"n d #1",
       "85":"n d #1",
-      "86":"n d #1",
-      "87":"n d #1",
-      "88":"n d #1"
+      "86":"a #2",
+      "87":"a #2",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #5586, opened 2020-08-28
Navigator.sendBeacon is fixed in Chrome

adamburgess:
Fixed in chrome 81: https://bugs.chromium.org/p/chromium/issues/detail?id=720283


Fyrd:
Thanks!

PR #5587, opened 2020-08-30
chore(docs): link for testing Opera Mini in an emu

kripod:
Opera Mini support data is non-trivial to gather without an actual device. I've added a note about testing in an emulator.


kripod:
I just realized that the version in the guide may not be adequate. I'll close this PR for now.

PR #5588, opened 2020-08-30
fix: partial logical props support in Opera Mini

kripod:
Used Android Studio's emulator and [this CodePen](https://codepen.io/kripod97/pen/Rwageze) for testing. Please see mdn/browser-compat-data for [why v48 was picked](https://github.com/mdn/browser-compat-data/blob/365cf70b860bdbbdc6663593beaa54a7f92d4631/css/properties/margin-block-start.json#L55-L57).

### DIFF
--- a/features-json/css-logical-props.json
+++ b/features-json/css-logical-props.json
@@ -362,7 +362,8 @@
       "11.5":"n",
       "12":"n",
       "12.1":"n",
-      "46":"a x #2"
+      "46":"a x #2",
+      "48":"a #3 #4"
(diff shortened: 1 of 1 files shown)


Fyrd:
Will update this once I replace the latest version of Opera for Android with the latest version (see #5602)

Fyrd:
Looking more into this one I discovered the support data for all chromium browsers was incorrect, so went ahead and did a big update manually that updates opera for Android as well. Thanks!

PR #5589, opened 2020-08-30
Unprefix 'filter' CSS property for Baidu Browser

kripod:
As tested on Baidu Browser v7.11.12.0, no `-webkit-` prefix should be necessary. I downloaded the APK file from [here](https://www.hiapphere.com/app/com.baidu.browser.apps) and ran it on an emulator for safety measures. Please verify whether the source is authentic before merging!


Fyrd:
Confirmed, thanks!

PR #5591, opened 2020-09-01
Update css-focus-visible.json

mattbrundage:
Remove extra bugzilla link that only pertains to the adding of documentation -- not to the implementation of the feature in Gecko.


Fyrd:
Thanks!

PR #5598, opened 2020-09-07
Add missing information in selector list `:not()` description

sroucheray:
Add the fact that compound selector in `:not()` is not supported in Selectors level 3 spec (e.g `:not(.a.b)`).


Fyrd:
Thanks!

PR #5600, opened 2020-09-07
Updates to A2HS

atjn:
Two updates to A2HS:

1. Firefox desktop has recently made leaps and bounds, and now has a surprisingly complete solution hiding behind a flag. I think it deserves a note! 🥳

2. The `appinstalled` and `beforeinstallprompt` events have been [removed](https://github.com/w3c/manifest/pull/836) from the spec. They were only supported in Chromium, and it does not seem like they will ever receive widespread adoption. I think it's misleading to track them the way we do today, but it might make sense to track them in a separate table.


Fyrd:
Looks good, thanks!

PR #5604, opened 2020-09-09
Mark Payment Request API as fully supported on Chrome for Android 53-84.

teo-tsirpanis:
It used to be marked as partially supported, and the first version to support it was indicated on a sidenote.


Fyrd:
Thanks, have updated this one manually. For future reference note that it's not possible to add versions by PR.

PR #5617, opened 2020-09-20
Add Opera 59

mischnic:
Opera Desktop 59 was missing, for every feature, the value for 58 and 60 was identical, so 59 should be as well.

Couldn't find a better process to add a "new"/old browser version.

Came here from https://github.com/browserslist/browserslist/issues/537, @ai


ai:
@Fyrd seems like Can I Use missed Opera Desktop 59 version in the data (here is [a proof](https://blogs.opera.com/desktop/2019/01/opera-59-0-3209-0-developer-update/) that this version was released). It caused an error in Parcel and Browserslist.

Fyrd:
Opera Desktop 59 was never released to the stable channel though, which is why it was never added to the caniuse database. There's other examples of version skipping like Chrome 82. I'm not sure if breaking some process somewhere is reason for adding a version that was never really released (outside of dev & beta versions)

ai:
Thanks for your answer. We will fix it in Browserslist.

PR #5618, opened 2020-09-21
Add Cookie Store API support table

atjn:
The Cookie Store API is a new, much better way of accessing and modifying cookies in the browser. It implements a clean JS interface, whereas the old method just supplied a string that the developer had to write their own parser for. It also has tonnes of other small improvements that simply stem from it being a more modern design compared to the ancient `document.cookie` method.
My description in the support table is intentionally pretty vague, since i couldn't find a way to condense all of the new improvements down into a few sentences. Feel free to rewrite it if you think you can do it :)

Cookies are used very widely, and it's a pain for developers that they have to either write or use third-party libraries just to read cookies. I think there will be quite a demand for knowing the support status on this feature as soon as it begins shipping in browsers.
I am adding the support table now because Chromium just posted an [intent top ship](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/zOsGZGMGiM4/zdJGMq5iBwAJ). They don't seem to know exactly which version will be releasing with it yet, but i will update the table when we know.


features-json/cookie-store-api.json
```diff
@@ -0,0 +1,416 @@
+{
+  "title":"Cookie Store API",
+  "description":"An API for reading and modifying cookies. Compared to the existing `document.cookie` method, the API provides a much more mordern interface, which can also be used in service workers.",
```
Fyrd, line 3: Small typo:
```suggestion
  "description":"An API for reading and modifying cookies. Compared to the existing `document.cookie` method, the API provides a much more modern interface, which can also be used in service workers.",
```

Fyrd:
Cool, thank you! Yeah this is a cool feature, could have used this like...15 years ago.

PR #5620, opened 2020-09-21
Remove note: scrollbar-width property is at risk of being removed

Shakil-Shahadat:
No longer true, see https://github.com/w3c/csswg-drafts/issues/1958#issuecomment-685592633


Fyrd:
Thanks!

PR #5625, opened 2020-09-25
Fix missing "i" in CSS content-visibilty => visibility

Schweinepriester:
I'm afraid it's in the news as well, but I don't think we can edit that code 0:-)

image


Fyrd:
🤦‍♂️ Argh, how embarrassing. Thanks for catching that!

PR #5629, opened 2020-09-28
Updates to CSS math functions

atjn:
Gary Simons recently made a video on the new `clamp` property, where he [accidentally used the wrong support table](https://youtu.be/dg488RrpNTc?t=70). I think it's hard to blame him, since searching for `clamp` or `css clamp` first displays a non-standard property, and then shows 2-7 MDN tables before showing the thing that i believe most people will be searching for.
One of the MDN tables are actually also for the same property, so that should be removed, right?

You encounter similar issues when searching for `min` and `max`.

I don't actually know how the search engine works, but i have tried to add keywords that should help bring this support table to the top. By all means change them if you know something i don't :)

I have also tested Android Opera 59 to support this with the [caniuse test](https://tests.caniuse.com/css-math-functions). This is expected, since it's based on Chromium.

And i did a bit of tidying.

### DIFF
--- a/features-json/css-math-functions.json
+++ b/features-json/css-math-functions.json
@@ -9,14 +9,6 @@
       "title":"Test case on JSFiddle"
     },
     {
-      "url":"https://bugzilla.mozilla.org/show_bug.cgi?id=css-min-max",
-      "title":"Firefox support bug"
-    },
-    {
-      "url":"https://crbug.com/825895",
-      "title":"Chrome support bug"
-    },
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks! Yeah, getting the search to return the appropriate order can be challenging, I'll need to take a look at optimizing that.

I usually only remove the MDN table if it matches a feature exactly, in this case it's just one of three so that's why it still appears.

PR #5633, opened 2020-09-30
Opera 71 supports AVIF

atjn:
The [caniuse test](https://tests.caniuse.com/avif) confirmed that Opera 71 supports AVIF, which is expected since Chromium 85 supports it as well, and Opera doesn't run it's own media stack.

I also tested Edge 85 on the upcoming Windows 10 October 2020 update with the AV1 video extension installed, and despite all that, it still didn't work 😂


Fyrd:
You're right! I thought I tested that before but looks like I was wrong. Thanks!

PR #5567, opened 2020-08-13
web share API edge support is 81 or later. (windows only.)

juner:
Fixes #5566

PR #5634, opened 2020-10-01
Web Share: Update Chrome/Edge status up to 87

mgiuca:
Note that Edge and Chrome are sharing an implementation.

The existing data that showed Edge supported this from 81 to 85 on Windows is wrong. It hasn't been implemented yet (https://crbug.com/1035527). It should be available on both browsers in 88, in Windows.


mgiuca:
@ericwilligers is this correct?

We're also planning to ship in Chrome OS in 88 but we can update that when it ships.

atjn:
> The existing data that showed Edge supported this from 81 to 85 on Windows is wrong. It hasn't been implemented yet [...]

This is incorrect, [please see this discussion](https://github.com/Fyrd/caniuse/pull/5567#discussion_r470023521). I just retested it on Edge 85, and it works fine.

This was also somewhat reflected on [the Edge status page](https://web.archive.org/web/20200901070509if_/https://developer.microsoft.com/en-us/microsoft-edge/status/) (as v79), but has since been updated to v88.

mhochk:
atjn is correct, this has already shipped in Edge for a few releases and is being moved to a shared implementation for both browsers in time for 88.

mgiuca:
OK thanks for the feedback. I didn't realise this shipped in Edge.

Updated the PR. It is now just additions, rather than changes. It's possibly premature to commit this since the relevant changes haven't actually landed in 88 yet. But we expect it'll be true, when 88 goes out, that both Chrome and Edge will support this on Windows.

Fyrd:
Thanks for the updates! However for now I'm only updating Edge to the latest version, you can't include versions past that via PR.

mgiuca:
OK, removed 88 from the PR.

That means we're just extending Edge up to 87, and also removing Chrome's "88": "n", since 88 hasn't happened yet and is likely to be "a #3".

features-json/web-share.json
```diff
@@ -58,7 +58,9 @@
       "81":"a #3",
       "83":"a #3",
       "84":"a #3",
-      "85":"a #3"
+      "85":"a #3",
+      "86":"a #3",
```
atjn, line 62:
```suggestion

```

features-json/web-share.json
```diff
@@ -58,7 +58,9 @@
...
+      "87":"a #3",
```
atjn, line 63:
```suggestion

```

features-json/web-share.json
```diff
@@ -230,7 +232,6 @@
       "85":"n",
       "86":"n",
       "87":"n",
```
atjn, line 234:
```suggestion
      "87":"n",
      "88":"a #3"
```

features-json/web-share.json
```diff
@@ -58,7 +58,9 @@
       "81":"a #3",
       "83":"a #3",
       "84":"a #3",
-      "85":"a #3"
+      "85":"a #3",
```
atjn, line 61:
```suggestion
      "85":"a #3"
```

atjn (changes requested):
PR's cannot _at all_ alter which specific versions are tracked. This should make your changes pass.

Edit: Sorry for the mess, i haven't figured out how to use the review tool properly, if there even is a way.

mgiuca:
@atjn OK, so how do new versions ever get added? Maybe that gets done globally at some point?

In that case, this PR doesn't actually do anything. I'll just wait until 88 comes up. (Closing for now.)

atjn:
@mgiuca yes, new versions are added globally by fyrd.

But Chrome 88 is tracked, so you can add the note now :)

PR #5642, opened 2020-10-06
Updates to font-rendering-controls

atjn:
The Android browser, Opera mobile and Firefox for Android all support the `font-display` descriptor. I have done some manual testing to confirm this, and it makes sense since they're based on browser engines that support it.

I also did some general cleanup on the feature.


Fyrd:
Looks good, thanks!

PR #5652, opened 2020-10-15
preconnect stopped working in Firefox 71

da2x:
I've tested every version from 70 to 82. 70 is the last version where `<link rel=preconnect` worked in Firefox. Firefox 71+ issues no DNS requests and doesn't try to establish a TCP or TLS session to preconnect origins. (I'm aware of the difficulties in testing this properly with DNS caching and stuff. This was verified with WireShark.)


Schweinepriester:
Could be <https://bugzilla.mozilla.org/show_bug.cgi?id=1543990>?

da2x:
68, and 69 (plus 70) work. I can't currently test older versions.

Fyrd:
Don't have an easy way of confirming this but it does seem https://bugzilla.mozilla.org/show_bug.cgi?id=1543990 confirms it, so merging for now. Thanks!

da2x:
It stopped working in 66, but it worked in 65, 67, 68, 69, and 70; before it stopped working again in 71 onwards. Bug 1543990 is probably just as good of a tracking bug as any other.

PR #5656, opened 2020-10-19
Chrome 88: CSS Automatic Hyphenation for other platforms

Schweinepriester:
- <https://bugs.chromium.org/p/chromium/issues/detail?id=652964#c70>

    > This patch enables the Minikin hyphenation engine on all
platforms except Mac. Mac keeps using the engine in the OS.

- <https://groups.google.com/a/chromium.org/g/blink-dev/c/sIRXYKA4jGE/m/OfZMzveCAAAJ>
- <https://www.chromestatus.com/feature/5672891947417600>

~Currently there doesn't seem to be a target version defined yet, so I've chosen 88 for now because that is the current Canary. Also it's not in Canary yet (the change is 2 h old right now), so the PR probably should stay open for a bit.~

<https://bugs.chromium.org/p/chromium/issues/detail?id=652964#c73>:
> The current plan is to ship in M88.

~Doesn't seem to work yet tho…~

There we go:

image

Test: <https://tests.caniuse.com/?feat=css-hyphens>

Also I've updated Android and Chrome for Android, because I'd argue for those versions the note and partial support wasn't applicable anyway anymore. Or should it trickle down from Desktop?


Fyrd:
Looking good, thanks!

Issue #2745, opened 2016-08-22
CSS Logical Properties support chart is out-of-date


alanhogan:
[CSS Logical Properties](http://caniuse.com/#feat=css-logical-props) is out of date. 

**Firefox**

It indicates that current and future Firefox versions do _not_ support the `text-align: start` and `text-align: end` values (note 1), but in fact I just confirmed that Firefox does support this in version 47.0.1.

Here is my test page: https://codepen.io/alanhogan/pen/jARkXG

![screen shot 2016-08-22 at 2 25 41 pm](https://cloud.githubusercontent.com/assets/6808/17872312/54b0ca7a-6874-11e6-9941-afb572d123ce.png)

The above screen shot is of the Firefox 47.0.1 (on Mac OS X 10.11) rendering, which is totally correct and matches Chrome et al.

In fact, the MDN page suggests that `text-align: end` has been supported since Firefox 3.6! https://developer.mozilla.org/en-US/docs/Web/CSS/text-align#Browser_compatibility

⁂

**Multiple Browsers**

The chart indicates that vendor prefixes are needed, but my [test page](https://codepen.io/alanhogan/pen/jARkXG) works unprefixed in all tested browsers (except MSIE/Edge where there is no support). I tested unprefixed `text-align: start` Firefox, Chrome, and Safari. 

I’m not sure if other logical properties need a prefix, but the chart is at best misleading on this point, as it stands today.

Thank you.

Fyrd:
This was updated some time ago, so closing ticket.

PR #5666, opened 2020-10-22
Update CSS Logical Properties for Chrome 87+

kiding:
Chrome 87 now supports the properties stated in the notes `#3` `#4` by default. Specifically:
- inset, inset-block, inset-inline
- margin-block, margin-inline
- padding-block, padding-inline

**Reference** [Chrome Platform Status](https://www.chromestatus.com/feature/6618771051511808)


Schweinepriester (approved):
<https://tests.caniuse.com/?feat=css-logical-props>:

image

kiding:
Thanks for the test! They seem to be somewhat outdated—The left red one checks for `margin-start` which I believe never got standardized according to https://github.com/w3c/csswg-drafts/issues/3651, and _for historical reasons_ they work with `-webkit` `-moz` prefixes. Many other properties are also missing. Should be revised in one way or another. 

Loosely related to #2745.

Fyrd:
Yeah, these tests only pick some of the properties... and if there's a gray dotted outline it means it's deprecated or something, but is worth checking for partial support.

Thanks for the update!

PR #5671, opened 2020-10-24
Revert Chrome CSS Logical Properties support to partial

kiding:
A followup on #5666. It turns out Chrome (including 87+) has never supported the following four properties standardized by https://github.com/w3c/csswg-drafts/commit/ffa8f54a4502aeaa80d828c3ace410a26974ea2d, while Firefox has [since 66.](https://developer.mozilla.org/en-US/docs/Web/CSS/border-start-start-radius)

- `border-start-start-radius`
- `border-start-end-radius`
- `border-end-start-radius`
- `border-end-end-radius`

Should’ve checked more thoroughly before submitting a PR. Sorry about that!


Fyrd:
No biggy, thanks! Have also added one of these to the test suite.

PR #5678, opened 2020-11-02
Reflect that Safari 13.1 supports CSS image-orientation property

ialexryan:
I couldn't find a first-party source for this unfortunately (Safari's release notes leave a lot to be desired) but I noted the change at the time it came out, and just tested again on my machine running Safari 13.1.1 on macOS 10.15.5 and verified that it is supported.


Fyrd:
Confirmed, thanks!

PR #5535, opened 2020-07-16
Safari 14: WebP and HTTP/3 require macOS 11 Big Sur?

Schweinepriester:
Following up #5484, it seems WebP and HTTP/3 require macOS 11 Big Sur alike other features in the past.

However I've only tested the [recently released Safari 14 beta](https://www.macrumors.com/2020/07/15/apple-releases-safari-14-beta/) on macOS Catalina 10.15.5, where both features don't seem to work:

Bildschirmfoto 2020-07-16 um 05 51 51

Bildschirmfoto 2020-07-16 um 05 54 43

https://tests.caniuse.com/ doesn't seem to have a test for HTTP/3 yet, so I've used https://cloudflare-quic.com/, which shows HTTP/3 in Firefox (with the flag enabled). Not sure if thats a valid test.

If anyone could check that WebP and HTTP/3 work on macOS 11 that be great to validate the changes here. I'm fine with leaving this open until someone or I can check.

PR #5679, opened 2020-11-03
🎉 iOS 14 webp support update!

Snuggle:
After testing on an iPhone 11 running iOS 14.1 and Safari 14, webp images now load without showing a “broken image” icon.

### DIFF
--- a/features-json/webp.json
+++ b/features-json/webp.json
@@ -410,7 +410,7 @@
   "notes_by_num":{
     "1":"Partial support refers to not supporting lossless, alpha and animated WebP images.",
     "2":"Partial support refers to not supporting animated WebP images.",
-    "3":"Partial support in Safari refers to being limited to macOS 11 Big Sur and later."
+    "3":"Partial support in Safari refers to being limited to macOS 11 Big Sur or iOS 14 and later."
   },
   "usage_perc_y":84.33,
   "usage_perc_a":0.73,


Snuggle:
Should the tag be changed from “Almost supported” to “fully supported”?

Is desktop Safari still not supporting webp on Catalina?

kiding:
Yes, Safari 14 on Catalina does not support WebP—It’s platform-dependent feature rather than browser. Since Safari 14 is available on Big Sur, Catalina, and even Mojave, it’s been considered that “fully supported” wouldn’t be correct.

As far as I know, the iOS Safari column refers to the version of iOS itself, and 14 is already marked fully supported.

Related to #5535.

Fyrd:
Yeah, as @kiding mentioned the partial support is only for desktop Safari so I don't think this addition is necessary. Perhaps an alternative would be to instead specify "desktop Safari" to make it a little clearer.

saintyvesfsya (approved):

Schweinepriester:
> I don't think this addition is necessary

+1, as the note is on Safari only, not iOS Safari, but I find specifying "desktop Safari" acceptable as well.

saintyvesfsya (approved):

saintyvesfsya (approved):

saintyvesfsya (approved):

saintyvesfsya (approved):
Ok

Snuggle:
image

Ahah, I see. Yeah, there already exists an 'iOS Safari' section, this didn't slow earlier when I created this PR.
Thanks for the feedback and have a lovely day! 💖

PR #5682, opened 2020-11-05
Add note for old Safari implementation of webshare

mkonikov:
Safari, prior to iOS 14, did not allow for a fetch call to happen between user trigger and when native share api is called.

Can see more details here -> https://stackoverflow.com/questions/56046434/how-to-use-webshareapi-preceded-by-an-ajax-call-in-safari

After filing a bug with the WebKit team it was fixed in iOS 14 + macOS 11

I think that should be noted for others reference.

### DIFF
--- a/features-json/web-share.json
+++ b/features-json/web-share.json
@@ -253,9 +253,9 @@
       "11":"n",
       "11.1":"n",
       "12":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #5684, opened 2020-11-08
edge update for native-filesystem-api.json

naipofo:
Edge 86 has native filesystem api enabled by default, like chrome.


Fyrd:
Confirmed, thanks!

PR #5690, opened 2020-11-17
Update webp support in safari 14 from p to y

gokulkrishh:
In the latest update of safari added WebP support by default.

[Release Notes](https://developer.apple.com/documentation/safari-release-notes/safari-14-release-notes)


Schweinepriester:
Note 3:

> Partial support in Safari refers to being limited to macOS 11 Big Sur and later.

Unless that is fixed, support should stay partial imho.

gokulkrishh:
@Schweinepriester Safari 14 was released in Sep ahead of GigSur for Catalina and mojave users.

[More](https://9to5mac.com/2020/09/16/safari-14-is-now-available-as-a-standalone-update-for-macos-catalina-and-mojave-users/)

Ideally, if you have safari 14 in older OS and update it you should have webp support with a patch which will be released or already released..

Fyrd:
@gokulkrishh Do you have information about the type of patch you refer to? Occasionally features are simply tied to the OS, so I wouldn't expect this to happen. As plenty of users will be on older versions of the OS for a while I agree it should remain partial support.

gokulkrishh:
@Fyrd Nope, i was just suggesting that feature could be part of browser patch not OS. But thinking about it again (apple 🙄), i think PR should be still in partial as the feature could be tied to OS (I am not sure either ways).

Fyrd:
Alright, closing this PR then, if anything changes to that effect we can make the update then.

Macleykun:
Heya, sorry to necromance an old pull. But i wanted to ask if webp will be fully supported by safari 14 or later. I understand that only a few operating systemens will recieve this standalone update. But how does this differ?
I assume that webp on safari 14 or later will always be partially and never be fully then? Even tho it's fully supported in 14 and later?

Schweinepriester:
> I assume that webp on safari 14 or later will always be partially and never be fully then?

In the past we've changed it to fully once a version releases which isn't available below what version of macOS is required to work. So assuming we can still extrapolate from the past (https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility), that would be Safari 16 then (because presumably v16 will be available starting with macOS 11 Big Sur).

Macleykun:
Thanks for the explanation!

PR #5697, opened 2020-11-24
Firefox 84:  Application cache removed

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/84>

> -   The [application cache](https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache) has been removed---developers should use the [Service Worker API](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) instead ([bug 1619673](https://bugzilla.mozilla.org/show_bug.cgi?id=1619673)).

https://tests.caniuse.com/?feat=offline-apps contradicts this though…?


Fyrd:
Thanks!

PR #5706, opened 2020-12-02
data correction

wordlesswind:
The Chromium team decided to stay in the h3-29 version for now.
https://blog.chromium.org/2020/10/chrome-is-deploying-http3-and-ietf-quic.html


Fyrd:
Thanks!

PR #5710, opened 2020-12-08
Update the spec URL for Offline web applications

sideshowbarker:
<s>This change drops the spec URL for the "Offline web applications" feature — because that entire section has now been removed from the HTML standard.</s>

This change updates the spec URL for the "Offline web applications" feature. That entire section has now been removed from the HTML standard, so this change updates it to point to an older W3C copy.

### DIFF
--- a/features-json/offline-apps.json
+++ b/features-json/offline-apps.json
@@ -1,7 +1,7 @@
 {
   "title":"Offline web applications",
   "description":"Now deprecated method of defining web page files to be cached using a cache manifest file, allowing them to work offline on subsequent visits to the page. ",
-  "spec":"https://html.spec.whatwg.org/multipage/browsers.html#offline",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
Mh, wouldn't there be value in having a deprecated spec linked? Like the snapshot before it was removed?

Fyrd:
A specification URL of some kind is required, perhaps https://www.w3.org/TR/2011/WD-html5-20110525/offline.html#offline would be a better alternative?

sideshowbarker:
> A specification URL of some kind is required, perhaps https://www.w3.org/TR/2011/WD-html5-20110525/offline.html#offline would be a better alternative?

 I don’t 

> A specification URL of some kind is required, perhaps https://www.w3.org/TR/2011/WD-html5-20110525/offline.html#offline would be a better alternative?

WFM — amended the patch here

Fyrd:
Thanks!

PR #5712, opened 2020-12-10
Add webkit bug to filter css property

jamiebuilds:
In Safari `filter` won't apply to child elements while they are animating. [WebKit Bug](https://bugs.webkit.org/show_bug.cgi?id=219729).


Fyrd:
Thanks!

PR #5717, opened 2020-12-16
Firefox no longer supports menuitem by default

artem1919:
It was put behind a flag in version 85.
https://bugzilla.mozilla.org/show_bug.cgi?id=1680596

Also, nested items on Firefox for android were fixed when they moved from Fennec to Fenix and are now properly supported.
Can't find a source for that one and maybe it was even fixed by mistake but "it works".
Screenshot_20201216-124720_Firefox Beta

### DIFF
--- a/features-json/menu.json
+++ b/features-json/menu.json
@@ -148,8 +148,8 @@
       "82":"a #1",
       "83":"a #1",
       "84":"a #1",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #5718, opened 2020-12-18
Firefox for Android does not support H264 from v68

pbaumard:
Details are in the added note: "From [MDN - Supported video codecs](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/WebRTC_codecs#Supported_video_codecs): Firefox for Android 68 and later do not support AVC (H.264) anymore. This is due to a change in Google Play store requirements that prevent Firefox from downloading and installing the OpenH264 codec needed to handle H.264 in WebRTC connections. See [Mozilla Support article](https://support.mozilla.org/en-US/kb/firefox-android-openh264) for details."

### DIFF
--- a/features-json/mpeg4.json
+++ b/features-json/mpeg4.json
@@ -379,8 +379,10 @@
       "87":"y"
     },
     "and_ff":{
-      "67":"y"
-      "68":"n #2"
+      "1-16":"n",
+      "17-19":"a #2",
+      "20-67":"y",
+      "68-83":"n #3"
     },
     "ie_mob":{
       "10":"y",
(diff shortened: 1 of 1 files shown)


Fyrd:
That may be true for WebRTC, but this support data refers to just playing the format in the <video> tag, which seems to work fine.

pbaumard:
Right, it's just that [autoplay is disabled by default](https://support.mozilla.org/en-US/kb/playing-videos-firefox-android), so there is a black screen by default. [MDN browser-compat-data for autplay](https://caniuse.com/mdn-html_elements_video_autoplay) is misleading.

PR #5727, opened 2020-12-29
Add a known issue on CSS Font Loading API

kiding:
Added a bug explaining the recent change in the spec: https://github.com/w3c/csswg-drafts/pull/5757.


Fyrd:
Thanks!

PR #5728, opened 2020-12-29
Adding spec to CSS line-clamp

HolgerJeromin:
Not sure what to do with this strange property.  :-)

But deleting the old note seems to be best.
Feel free to edit to a better solution.


Fyrd:
Looks good, thanks!

PR #5734, opened 2021-01-07
Update pointerlock.json

supercairos:
Pointer Lock doesn't work on Safari on iOS


Fyrd:
Just tested and looks like you're right! Thanks for the correction.

PR #5736, opened 2021-01-10
AVIF is going to be supported in Firefox 86

kkk669:
Details: https://bugzilla.mozilla.org/show_bug.cgi?id=1682995


Fyrd:
Thanks!

atjn:
FYI - I was holding out on creating a PR for this because the implementation still has some major flaws that could land it in `partial support` territory, or maybe cause a delay of the launch ([see bug 1684688](https://bugzilla.mozilla.org/show_bug.cgi?id=1684688)). All major flaws are expected to be fixed before v86 launches, but that is yet to be seen ;)

The implementation also has many minor issues that won't be fixed before v86, but they are similar to bugs that are still present in Chromium, so they probably don't need to be mentioned. ([bug 1634741](https://bugzilla.mozilla.org/show_bug.cgi?id=1634741), [bug 1654461](https://bugzilla.mozilla.org/show_bug.cgi?id=1654461), [bug 1657300](https://bugzilla.mozilla.org/show_bug.cgi?id=1657300))

I'll try to follow the development and post a new PR if it becomes necessary.

Fyrd:
Thanks for the info @atjn !

atjn:
FYI, it seems like Firefox will be shipping a half-broken implementation in v86, but they already have a good number of fixes live in Nightly 87. My concern was mainly if they were going to ship AVIF and not fix these serious issues for months or years. Since they already have a fix coming, I don't think it's necessary to mention it in Caniuse, but if anyone else wants to spend the time to properly track and document these issues, be my guest :)

PR #5737, opened 2021-01-10
Mention lack of support for AVIF sequences, fix #5736

bmisiak:
Hello. Quick edit, hope I got it right.

https://gif2avif.com/browser-test/ explains it well - both Firefox and Chrome currently only support still AVIF images. Animated AVIFs, a.k.a. AVIF Sequences, don't seem to be supported.

Also, fixes #5736 - Firefox 86 removes the flag.


Fyrd:
I think since the primary use of AVIF will still be static images support should still be `y`. Though I'm fine including the note to those re: AVIF sequences.

Support behind a flag should still be `n` FYI. Thanks!

Fyrd:
Looks good, thanks!

PR #5739, opened 2021-01-11
Updates to :is()

atjn:
Chrome is now adding support for CSS `is` (yay!)
Their documentation on the feature is frustratingly bad, but I have simply brute-force [tested](https://codepen.io/atjn/pen/MWKErBe) my way to concluding that it is being enabled in v88. I will check back and make sure it's still true when the stable release rolls out :)

I have also removed a bunch of notes that described which older aliases were still available in browsers that fully support `is`. Now that all browsers will receive support, I don't see how this could be useful to anyone, and I also don't expect anyone to keep track of when the aliases might be removed.


Fyrd:
Good changes, thanks!

PR #5742, opened 2021-01-15
update(link-rel-preload): Firefox added support in 85

myakura:
it will ship in Firefox 85 (first enabled by default on 86, and then backported to 85).
https://hg.mozilla.org/mozilla-central/rev/e205e7b4a73c
https://hg.mozilla.org/releases/mozilla-beta/rev/c46b088d9a1d


atjn:
This is a duplicate, we are tracking it in #5447

I am currently testing the feature, trying to figure out why some preload tests still fail in FF85. I think it might end up being partial support.

myakura:
@atjn ah cool. thanks for the pointer! I'll close this one then.

> I am currently testing the feature, trying to figure out why some preload tests still fail in FF85. I think it might end up being partial support.

do you have public urls for the tests? (I'm just curious)

atjn:
Alright thanks, I'll make sure to finish my testing then, so I'm not blocking the update from going live.

The test on caniuse (https://tests.caniuse.com/link-rel-preload) does not detect any support in FF85. I think my conclusion is going to be that the test is wrong :)
- I'm writing my own test right now.

myakura:
@atjn oooh I've never known that caniuse has its own tests. interesting.

> The test on caniuse (https://tests.caniuse.com/link-rel-preload) does not detect any support in FF85. I think my conclusion is going to be that the test is wrong :)
> I'm writing my own test right now.

you might want to check out https://wpt.fyi/results/preload?label=experimental&label=master&aligned (source at https://github.com/web-platform-tests/wpt/tree/master/preload ). tests on wpt.fyi site are maintained by standard/browser developers.

atjn:
Thanks for the suggestion. I find it very hard to use WPT because the tests are so incredibly specific and not very well documented. I'm sure it's a great tool for someone working on the innards of the browser engine, trying to tweak some small detail somewhere, but for understanding if a feature is generally supported or not supported, I find it much efficient to use Caniuse tests or write my own tests :)

PR #5754, opened 2021-01-29
Update web-share.json

mattbrundage:
Now available on Chrome (desktop) 89: https://www.chromestatus.com/features/4777349178458112


Fyrd:
Looks like it's only for Windows and ChromeOS, not yet MacOS so should be partial support...
Although I just tested Chrome 89 on Windows too and `navigator.share` still returns undefined there too? So not sure that information is correct.

mattbrundage:
Here's Chrome on Windows: version 89.0.4389.23 (Official Build) beta (64-bit)
showing Share API user interface, as well as ``'share' in navigator`` returning ``true`` in the console. But I agree that if this isn't supported on MacOS, then it should be partial support.

image

PR #5755, opened 2021-02-02
Added bug in Safari 9.1 - 9.3 that effects nested calculations

dletorey:
This bug means that when using a previously calculated variable with calc() the new variable is not resolved.
https://bugs.webkit.org/show_bug.cgi?id=161002


Fyrd:
Since the note for no. 3 is the same as no. 2, you can just use `#2`.

dletorey:
Closing this pull request and will make new one when updated.

PR #5765, opened 2021-02-10
Adds SpeechRecognition API data for STP

apple-web-evangelist:
STP 119 provides support for the SpeechRecognition API.


Fyrd:
Thanks!

PR #5772, opened 2021-02-15
Update Firefox for Android - Pointer Events support

patrickhlauke:
x-ref https://github.com/mozilla/platform-status/issues/615 /cc @smaug----


patrickhlauke:
There's probably also some complication here as there's original Pointer Events, Pointer Events Level 2, and the upcoming Pointer Events Level 3 (and from that point on, PE will be an evergreen spec with potentially new features being added without a version number bump). not sure how that can/will be handled here...

Fyrd:
Thanks, I updated the latest version of Firefox for Android to 85 and took care of this change with that, so closing this PR. As for how updates will be handled, that usually depends on what the changes are, one option is to change the current feature to "Pointer Events (basic support)" and add new features as separate tables. May be best for a separate issue in the future.

patrickhlauke:
Thanks @Fyrd 

> May be best for a separate issue in the future

oh for sure, didn't want to cram it all in here, but thought it worth mentioning :)

PR #5776, opened 2021-02-23
Update SameSite spec to the latest

kiding:
As [Incrementally Better Cookies](https://tools.ietf.org/html/draft-west-cookie-incrementalism-01) has been adopted into rfc6265bis[-07](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-07.html#name-draft-ietf-httpbis-rfc6265bis-07), this PR updates the related information accordingly. 

Related: https://github.com/httpwg/http-extensions/issues/1241.


Fyrd:
Thanks!

PR #5777, opened 2021-02-23
Improve linking of image formats

atjn:
I have earlier tried to link the WebP and AVIF support tables to each other, to help inform users of the new image formats and make it easier to navigate what is what and what is best. I think this linking should extend to JPEG XL, once again in the interest of giving a full overview.

When I made the original links, I did not include JPEG XR and 2000, because there weren't any real replacements for them. With WebP now having a distinct advantage in browser support, and JPEG XR being an objectively better format, I think it makes sense to connect them as well, and slowly start nudging people towards some of the better alternatives.

I have also bumped the version of Firefox that supports AVIF by default. As discussed in #5736 - their current implementation is very broken, and it seems they have decided to wait until v87 so they can fix the major issues before release ([bugzilla discussion](https://bugzilla.mozilla.org/show_bug.cgi?id=1682995#c32)). I will keep an eye on the progress.


Fyrd:
Looks great, thank you for doing this! I'll merge it as is but I'm curious if it's fair to say that WebP was designed to supersede JPEG XR since WebP's introduction only happened ~6 months after JPEG XR's.

atjn:
Glad you like it! Agreed, that one sentence is somewhat funky, but its purpose is mainly to let the user know "hey, you should probably take a look at these other file formats as well since they're probably better for what you need", in as few words as possible.

I can't think of a concrete situation where this would be misleading. If you care so little about your image format that you take guidance exclusively from a Caniuse description, you are definitely better off with WebP. If you are so advanced that you would benefit from using a format that only works on a subset of devices and has poor support in the general ecosystem, you will definitely already know everything we write in that description.

PR #5780, opened 2021-02-26
Updates to permissions headers

atjn:
The permissions policy header is now enabled by default in Chromium 🥳
This is tested with the [caniuse test](https://tests.caniuse.com/permissions-policy), as well as being in the release notes.

The rest is just maintenance :+1:


Fyrd:
Looking good, thanks!

PR #5787, opened 2021-03-01
Reducing supported versions of chrome in viewport-units.json

sindrebilden:
By testing demo sites from w3schools in browserstack.com, I discovered that chrome version 20-33 does not support vw and vmin.
This is done by checking the following sites on various versions of chrome:
https://www.w3schools.com/cssref/tryit.asp?filename=trycss_unit_vw
https://www.w3schools.com/cssref/tryit.asp?filename=trycss_unit_vmin
https://www.w3schools.com/cssref/tryit.asp?filename=trycss_unit_vh

I appreciate if someone can verify this :)


Fyrd:
Just tried it and I believe the units work for most properties (e.g. width), just not the font-size property. If you like you can add another note for that and reduce support to "partial" though considering how old these versions are I'm not sure it's worth it.

Fyrd:
Closing as per my last comment, not particularly relevant for these older versions.

Issue #5660, opened 2020-10-22
@font-face descriptors to override font metrics


kiding:
`ascent-override`, `descent-override` and `line-gap-override` descriptors for `@font-face`. 

### Implementation Status
* Chrome: 86+ flagged, 87+ by default

### References
* [Chrome Status](https://www.chromestatus.com/feature/5651198621253632)
* https://github.com/w3c/csswg-drafts/issues/4792

tobireif:
It would be great if these properties could get added.

The spec: https://www.w3.org/TR/css-fonts-4/#font-metrics-override-desc

ZerdoX-x:
so what's the status of it? :eyes:

tobireif:
@ZerdoX-x See https://tobireif.com/posts/ensuring_the_correct_vertical_position_of_large_text/ -> "Browser support".
Also
https://caniuse.com/mdn-css_at-rules_font-face_ascent-override
https://caniuse.com/mdn-css_at-rules_font-face_descent-override
https://caniuse.com/mdn-css_at-rules_font-face_line-gap-override

ZerdoX-x:
@tobireif Thank you. So why this issue isn't closed yet?

PR #5788, opened 2021-03-02
Adds font-metrics-overrides file

kennethormandy:
Adds font-metrics-overrides file, based on the included links and the links in #5660.

Things you might want to change:

- The filename could also be `font-metrics-override-desc` to match https://drafts.csswg.org/css-fonts-4/#font-metrics-override-desc, or be prefixed with `fontface` instead of just font? I’ve included the relevant keywords regardless—I’m not clear on the filename conventions
- I wasn’t clear on whether this being behind a feature flag meant `"86": "n d #1",` should actually be `a`lmost instead of `n`o


features-json/font-metrics-overrides.json
```diff
@@ -0,0 +1,419 @@
...
+  "description": "Default font metrics overriding: the ascent-override, descent-override and line-gap-override descriptors",
+  "spec": "https://drafts.csswg.org/css-fonts-4/#font-metrics-override-desc",
+  "status": "unoff",
+  "links": [
+    {
+      "url": "https://docs.google.com/document/d/1PW-5ML5hOZw7GczOargelPo6_8Zkuk2DXtgfOtJ59Eo/edit",
+      "title": "New @font-face descriptors for overriding font metrics"
+    },
+    {
+      "url": "https://www.chromestatus.com/feature/5651198621253632",
```
Fyrd, line 12: This link will automatically be added thanks to the "chrome_id" value, so it can be removed here.

Fyrd:
File name looks good and `n d #1` is correct, since there's no support by default. 👍

kennethormandy:
Well I guess I made the case for the CI, so thanks for merging that ha. Let me know if you want me to squash these down to one commit.

Fyrd:
Thanks! Already squashed.

Issue #2926, opened 2016-10-24
List Chrome for Android ahead, like Desktop?


Schweinepriester:
As there is a Beta (<https://play.google.com/store/apps/details?id=com.chrome.beta>, currently 55) and a Canary version (<https://googlechromereleases.blogspot.de/2016/10/introducing-chrome-canary-for-android.html> => <https://play.google.com/store/apps/details?id=com.chrome.canary>) of Chrome for Android, what about listing it ahead like the desktop version? Maybe even Android Browser, too?

Could be usefull, e.g. for a pull request for #2925.
Or is Chrome for Android synced with Chrome when updating or the like?

Fyrd:
Sure, could do that. Mostly I'm not much in a rush to do this because 90% of the time the information's the same as Chrome Desktop and it just adds more work for little value.

Schweinepriester:
@stof tho alike, not exactly: this one requests future (beta, canary) versions while #3518 requests previous ones…

antonjuulnaber:
Just wanted to chirp in and say that it is a pretty big deal with the WebXR spec. Many of the features are out earlier on the Android version, and they have very different quirks that i would have liked to reflect when i updated the data.

@Fyrd  I am having a hard time accepting the argument that it would give "little value". Chrome for Android is currently the most used browser on the planet, so missing a small discrepancy in how it works can have huge consequences for a website.
A very useful feature in caniuse, is usage-relative view. Currently, it seems like all Android phones are up to date with the latest version, so it is easy to forget that there is a (possibly sizeable) user-base that dont have the new version, because they are being covered up and shown as if they do have the latest version.

I am both arguing for listing the versions ahead, but also backlogging (#3518).
Have you considered a model where we could add information only when it is different from desktop, and otherwise it would just fall back to the way it works right now? or fall back to just mirroring chrome for desktop?

Fyrd:
@antonjuulnaber Yeah, I do actually have some code sitting around that does what you suggest, mirrors another browser's version unless otherwise specified. At some point I'll get back to that and we'll have a better way of dealing with "mostly the same" browsers.

jpmedley:
>Mostly I'm not much in a rush to do this because 90% of the time the information's the same as Chrome Desktop and it just adds more work for little value.

[...]

PR #5790, opened 2021-03-03
avif.json: Add AVIF support for Chrome for Android 89

EwoutH:
Chrome for Android (and_chr) version 89 supports the AVIF image format.

Source: https://blog.chromium.org/2021/01/chrome-89-beta-advanced-hardware.html


Schweinepriester:
While I share your enthusiasm, the changes wont work [because per CONTRIBUTING](https://github.com/Fyrd/caniuse/blob/main/CONTRIBUTING.md#unsupported-changes):

> 
> Add a new browser or browser version

and there's currently only one version being kept for Chrome for Android.

Since AVIF is already supported in Chrome (for desktop), that state will almost certainly the one which get used when @Fyrd replaces Chrome for Android with 89.

Having said that, myself and others wish for future (#2926) and past (#3518) versions of Chrome for Android to be able to make changes like this one...

EwoutH:
Yes, let's close this and wait until Chrome for Android is updated to 89.

PR #5793, opened 2021-03-04
Edge 88: CSS Hyphenation not yet on par with Chrome

### LINKED ISSUE #5792: hyphens: auto does not work in Edge 88
Currently https://caniuse.com/?search=hyphen shows that `hyphens: auto` has full support in Microsoft Edge 88. However it is actually not supported at all - at least under Windows - not even with any of the prefixes (`-ms-` or `-webkit-`). Previous (non Chromium) versions of Edge _did_ support `hyphens: auto` with the `-ms-` specific prefix, as is indicated correctly on caniuse.

Since it does work in Chrome under Windows however, I suspect that this is more of a bug or oversight on Microsoft's part?

Schweinepriester:
Fixes #5792.

<https://tests.caniuse.com/css-hyphens>:

macOS:

image

Android:

Edge for Android just wouldn't work in my emulator (Android Studio => AVD; Firefox and Chrome Canary are fine). My assumption is that the tests are completely green as well. But it would be really good if someone could test that before merging.

Win10 (EDIT: 89.0.774.45 is the same):

image

Haven't tested Linux though.

If we cant test Edge for Android I would change the PR so the previous state (`a #1`) is returned. Note 1 is a bit unfortunate because it only mentions Chrome. But at least that would be more accurate than `y`, because thats not true (see Win10).

### DIFF
--- a/features-json/css-hyphens.json
+++ b/features-json/css-hyphens.json
@@ -53,7 +53,7 @@
       "85":"a #1",
       "86":"a #1",
       "87":"a #1",
-      "88":"y"
+      "88":"a #2"
     },
     "firefox":{
       "2":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Just checked Edge for Chrome and it's all green. Thanks!

Schweinepriester:
> Edge for Chrome

is a nice typo :D 0:-)

PR #5798, opened 2021-03-09
Updates to AVIF

atjn:
AVIF support has been further delayed in Firefox. [See comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1682995#c37).
At this point I think we should mark v88 `unknown` and wait a bit before we proclaim that it definitely will be supported :)

Also, the note on still images applies to Opera as well, since they use the same codebase.


Fyrd:
Looks good. Thanks!

PR #5806, opened 2021-03-13
Intrinsic & Extrinsic Sizing: Align Safari for iOS to Safari

Schweinepriester:
... since [test](https://tests.caniuse.com/intrinsic-width) results are the same and Safari has this state since version 11.

Bildschirmfoto 2021-03-13 um 23 56 53

Bildschirmfoto 2021-03-11 um 23 02 09

I've only changed Safari for iOS 14, since I was able to test that. I was tempted to change it back to version 11, but the support start is already different (Safari: 6.1; Safari for iOS: 7) so I've decided against that.

However, maybe we need more elaborate tests to figure the whole situation out?


Fyrd:
Thanks! I agree more tests for this would be helpful but for now this update makes sense.

PR #5810, opened 2021-03-15
Chromium supports WebHID by default

atjn:
Chromium has enabled support for WebHID by default 🥳

I have confirmed support with the [caniuse test](https://tests.caniuse.com/webhid) and it correlates with Chrome documentation. 

I have removed old comments referring to versions behind flags. They should not be necessary anymore :)


Fyrd:
Thanks!

PR #5812, opened 2021-03-15
Chrome for Android supports WebNFC by default

atjn:
Chromium for Android 89 has enabled WebNFC support by default 🥳

I have confirmed support with the [caniuse test](https://tests.caniuse.com/webnfc) and it correlates with Chrome documentation.

WebNFC is also designed to work in desktop browsers (likely with USB NFC adapters), but that feature seems to have been scrapped in Chromium at least for now. I have tested and confirmed that desktop Chromium 89+ has removed support behind a flag.


Fyrd:
Thanks!

PR #5799, opened 2021-03-10
avif.json: Add AVIF support for Chrome for Android 89

EwoutH:
Chrome for Android (and_chr) version 89 supports the AVIF image format.

Source: https://blog.chromium.org/2021/01/chrome-89-beta-advanced-hardware.html

_Note: I will rebase this PR and mark it ready for review when Chrome for Android is updated to 89 repository-wide._

I also tested it:

PR #5813, opened 2021-03-15
Updates to AV1

atjn:
I was scratching my head reading #5799 because AV1 support has never been announced for Android Chromium. Turns out it has been silently supported for quite a while.

My testing has been a bit all over the place because [Bitmovin](https://bitmovin.com/demos/av1) and [Youtube](https://www.youtube.com/playlist?list=PLyqf6gJt7KuHBmeVzZteZUlNUQAVLwrZS) seems to sometimes block AV1 playback despite it being supported, the [caniuse test](https://tests.caniuse.com/av1) uses a relatively unsupported container format (resulting in a lot of false negatives) and I don't have access to all historic mobile browsers, but I think my results are pretty solid :)

I have tested and confirmed support in Chrome, Opera and Samsung Internet 13. Support in Samsung Internet 12 is an educated guess based on the release timeline.

I have also made a few tweaks to the Firefox notes.


Fyrd:
Yeah I've found it hard to test for AV1 as well, hard to find useful samples. If there's a more representative file I can use for the test would be happy to use it. Thanks for the updates!

atjn:
I did try to see if I could find a video file to suggest for the test, but couldn't find anything readily available.

DeeDeeG:
There are some reference AV1 encodes available from various places:

- http://download.opencontent.netflix.com/?prefix=AV1/Chimera/Old/
  - Ambiguous license I suppose, but I don't think Netflix would be mad about these being used for testing purposes, as that's why these were made in the first place, to test AV1
- https://github.com/chromium/chromium/tree/master/media/test/data\
  - From Chromium, I'm pretty sure these are free/libre licensed
  - Search "AV1" on this page to find the AV1 encodes... example: [[bear-av1.webm](https://github.com/chromium/chromium/blob/master/media/test/data/bear-av1.webm)]
- https://tools.woolyss.com/html5-audio-video-tester/?u=woolyss.com/f/caminandes-1-llama-drama-av1-opus.webm
  - Point to this? https://commons.wikimedia.org/wiki/File:Caminandes-_Llama_Drama_-_Short_Movie.ogv
    - Free/libre CC By Attribution license
    - Would have to download the latest file from Wikimedia and verify to see if the copies on Wikimedia are still AV1 encodes in a WEBM container, or something else? Or use the copy at Woolyss.com?

atjn:
Oh hey, thanks for the link to Chromium test files @DeeDeeG. I have avoided using videos from Netflix and Woolyss because they are both under the not-so-safe Netflix license, but the Chromium files are perfect!

@Fyrd I'd suggest adding [bear-av1.mp4](https://github.com/chromium/chromium/blob/master/media/test/data/bear-av1.mp4) and [bear-av1.webm](https://github.com/chromium/chromium/blob/master/media/test/data/bear-av1.webm) to the test. Those are two very common formats, I would expect them to work everywhere.

DeeDeeG:
I would also be able to encode in basically any container/to any specifications (run length, resolution, filesize...) desired.

(The obvious source material would be a short clip of one of the [Blender Open Movies](https://www.blender.org/about/projects). I could upload a re-encode of some clip from one of those here under an appropriate license.)

PR #5814, opened 2021-03-15
CSS overflow: Chrome 90, Firefox 81 (align to MDN)

Schweinepriester:
<https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#browser_compatibility>

I've [test](https://tests.caniuse.com/css-overflow)ed Chrome 90:

image

and Firefox 86:

image

Otherwise copied from MDN.

Just in case: I'm arguing for `y` since <https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin> is another property and currently all browsers which support `clip` also support
> Multiple keyword syntax for `overflow-x` and `overflow-y`

EDIT:

Plus I've added <https://bugs.webkit.org/show_bug.cgi?id=198230>.


Fyrd:
Nice, thanks!

PR #5818, opened 2021-03-16
Chrome supports AVIF image sequences

wantehchang:
Chrome 85 or later supports AVIF image sequences.

The MIME type "image/avif-sequence" specified in AVIF Specification
v1.0.0 was never registered with IANA [1] and has been removed from the
AVIF specification [2]. The MIME type "image/avif" and the file
extension ".avif" are now used for both still images and image
sequences.

[1] https://www.iana.org/assignments/media-types/media-types.xhtml#image
[2] https://aomediacodec.github.io/av1-avif/


wantehchang:
Looking at the git history of features-json/avif.json, I found that the footnote #2 was added in this commit:

> commit b6d31b12aead0188ac46a78a695e34301aa40306
> Author: Brian Misiak <wickoo@gmail.com>
> Date:   Sat Jan 9 22:01:53 2021 -0800
> 
>     Mention lack of support for AVIF sequences, fix #5736
>     
>     Hello. Quick edit, hope I got it right.
>     
>     https://gif2avif.com/browser-test/ explains it well - both Firefox and Chrome currently only support still AVIF images. Animated AVIFs, a.k.a. AVIF Sequences, don't seem to be supported.
>     
>     Also, fixes #5736 - Firefox 86 removes the flag.

Using "View page source" on https://gif2avif.com/browser-test/, I found that the img tag that tests AVIF image sequence support is `Thumbs up computer kid .avifs animation`. I downloaded `sample.avifs` using the `wget https://gif2avif.com/sample.avifs` command. `sample.avifs` is an AV1 video file in the IVF file format, not an AVIF image sequence file. Here is the first four lines of the output of the `xxd sample.avifs` command:

```
00000000: 444b 4946 0000 2000 4156 3031 9001 2c01  DKIF.. .AV01..,.
00000010: 0a00 0000 0100 0000 0000 0000 0000 0000  ................
00000020: b21c 0000 0000 0000 0000 0000 1200 0a0b  ................
00000030: 0000 00fa 231f 2b10 8560 501a 1010 0227  ....#.+..`P....'
```

Note the `DKIF` signature of the [IVF file format](https://wiki.multimedia.cx/index.php/IVF) and the `AV01` codec name.

Fyrd:
Confirmed, thanks for the details and the example!

PR #5819, opened 2021-03-16
Edge 12 doesn't support classes

angieskazka:
according to this
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#browser_compatibility

also Babel thansforms ES6 classes to functions with `targets: { "edge": "12"}` in config


Fyrd:
Yup, can confirm. Thanks for the correction!

PR #5822, opened 2021-03-17
Chrome: :dir() CSS pseudo-class is back to experimental

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=576815#c31>

> Flip CSSPseudoDir back to experimental

~~Haven't tested it yet, I suppose we have to wait for tomorrows build.~~

Yup: <https://tests.caniuse.com/css-dir-pseudo>

image

Also checked enabling the flag, which made it work again.


Fyrd:
Thanks!

PR #5828, opened 2021-03-22
Chrome 91: `position: sticky` on `thead` & `tr` fixed 🎉 

Schweinepriester:
The [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=702927) from note 4:

> **Status:** Fixed (was: Started)
> 
> I believe with the above patch and tables-ng being enabled [1] this issue is now resolved.
>
> [1] <https://bugs.chromium.org/p/chromium/issues/detail?id=958381#c248>

<https://bugs.chromium.org/p/chromium/issues/detail?id=958381#c248>

> [TablesNG] turn to stable
>
> Goal is to ship in M91

The [codepen](https://codepen.io/Fyrd/pen/wegxqN) from <https://tests.caniuse.com/css-sticky> is already green in Chrome 91 (unlike 89):

Bildschirmfoto 2021-03-22 um 22 11 52

And also seems to behave like Firefox (86.0.1), so not exactly sure what was missing until today, but at the latest tomorrows Chrome Canary appears to have it all 🎉


Fyrd:
Awesome, thanks!

PR #5832, opened 2021-03-26
Add `<template>` to keywords of template

Schweinepriester:
Coming from <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template> I [searched for `<template>`](https://caniuse.com/?search=%3Ctemplate%3E) and found nothing, therefore add it to the keywords.

I found https://github.com/Fyrd/caniuse/blob/8f4d0acf750890359315e97829d8802e854fae12/features-json/html5semantic.json#L435 which made it look legit to add \<elements\> as keywords.

### DIFF
--- a/features-json/template.json
+++ b/features-json/template.json
@@ -426,7 +426,7 @@
   "usage_perc_a":0.34,
   "ucprefix":false,
   "parent":"",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #5834, opened 2021-03-28
Update note for CSS "image-orientation" bug

ingmarh:
The bug where an image’s aspect ratio is not maintained when
`object-fit: cover` and `image-orientation: from-image` are used
together is fixed with Chrome version 89.

See:
https://bugs.chromium.org/p/chromium/issues/detail?id=1082669


Fyrd:
Thanks!

PR #5836, opened 2021-03-30
Update HTTP/3 current status

bazzadp:
HTTP/3 support has now been rolled out by default to [Chrome and Edge](https://groups.google.com/a/chromium.org/g/net-dev/c/5M9Z5mtvg_Y/m/iw9co1VrBQAJ) between versions 88 and 89.

Firefox looks to have this enabled in 88.

Currently [caniuse says Safari has enabled by default](https://caniuse.com/http3), but that doesn't appear to be the case so reverted that.


bazzadp:
Confirmation that HTTP/3 is rolled out to at least 76% (and more likely nearer 100%) of up to date Chrome/Edge users: https://twitter.com/ericlaw/status/1377758311056048131?s=21

ericlaw1979:
I'm not sure this change is quite right. I think

```
   "and_chr":{
      "89":"n"
    },
```

should be "Yes" (based on the Finch setting of ~96%), but I don't recall whether Finch impacts the Android WebView.

bazzadp:
> I'm not sure this change is quite right. I think
> 
> ```
>    "and_chr":{
>       "89":"n"
>     },
> ```
> 
> should be "Yes" (based on the Finch setting of ~96%), but I don't recall whether Finch impacts the Android WebView.

Fixed thanks @ericlaw1979 - I missed there where two Android settings (“Android Browser” and “Chrome for Android”) and only updated the former not the latter. That’s corrected now.

> Also (unsure if it's worth mentioning) but Edge set to 100% in Edge 87.

Edge 87? Or 89?

Personally I think going into the whole Finch/ECS thing (explained excellently in [your post on this](https://textslashplain.com/2017/10/18/chrome-field-trials/) is a bit too detailed and should just go with the majority.

And I’ve gone with the thinking that if an older version is grabbing the new Finch/ECS settings then highly likely they are getting to upgrade to be told there is a new version to and upgrade to that.

Hence why I’ve said this is active from version 89 onwards even if earlier version might have got it after their initial release, or if ~4% of Chrome 89 users don’t have it.

Lots of other reasons people won’t have it too (e.g. stuck behind corporate proxy, antivirus, or some other proxy) so never gonna be perfect.

Hoping that makes sense and you agree with the thinking here.

ericlaw1979:
yeah, I think anything's an improvement here. We ramped to 100% back in 87, but yeah, given that 89 is now stable, older values are of historical interest at best.

bazzadp:
Just to be clear, which if these is most accurate?:
1. Edge 87 was released with it ramped to 100%
2. Edge 87 was released without it on for vast majority and then quickly ramped up to 100% before Edge 88 was released.
3. Edge 87 and 88 were released without it on for vast majority and then ramped up to 100% before Edge 89 was released.

I presumed it was 3, which is what my PR is based upon, but if it’s 1 then happy to backdate to 87, and if it’s 2 then happy to backdate to 88.

Any ideas on Chrome? From my links above it seemed to be to be defaulted to on between 88 and 89 and, as per your Twitter posts, is now on for ~96%, in which case I’ll leave that as being support from 89.

ericlaw1979:
It appears that Edge Stable went to 100% for `h3-29,h3-q50` on November 5th for v86+, which had then been in stable for almost a month. So, Edge 87 had QUIC the whole time.

For Chrome, it's hard to do the archaeology here because we only have daily snapshots (not easily diffed) but on Nov5th, 2020, their Finch config looked much like it does today, with <10% of configs with H3-29 disabled.

bazzadp:
Thanks @ericlaw1979 - based on that I’ve backdated the changes for both Edge and Chrome to 87.

Fyrd:
Looking good, thanks all!

PR #5840, opened 2021-04-03
Firefox 89: AVIF by default

Schweinepriester:
<https://bugzilla.mozilla.org/show_bug.cgi?id=1682995#c41>
plus 
<https://bugzilla.mozilla.org/show_bug.cgi?id=1682995#c43>


Fyrd:
Thanks!

atjn:
Firefox has said they would enable AVIF by default in FF83, FF86, FF88, and now FF89. Every deadline was scrapped because the implementation was too buggy. No bugs have been fixed since it was backed out from release in FF86.

As discussed in #5736 and #5798, I think FF89 should be marked `unknown` and not `supported`.

Schweinepriester:
I mean we can still change it to unsupported once FF89 is closer/out... just like we'd have to change it from `unknown` to something else when the time comes.

But if you're more like

54pr8u

I'm down for that, too :D

Schweinepriester:
Also they seem to acknowledge the current state:
<https://bugzilla.mozilla.org/show_bug.cgi?id=1682995#c46>

> The bug is correctly RESOLVED FIXED at the moment because AVIF is currently enabled by default in nightly, and absent intervention will be going through our normal release process to ship in Fx89. Yes, there are incomplete blocking issues and our expectation is that those will be resolved in time for shipping in Fx89. You can rest assured that this bug will get updated in response to any changes to that plan.

atjn:
Nice :D

It's not the end of the world, but when I see a green rectangle on caniuse, I understand it as "we tested it and it works". I don't like being told that I can build my new webapp with a new feature, only to come back three months later and find that it wasn't true. 

And in this case, I am almost certain it won't be shipping in v89 ;)

PR #5842, opened 2021-04-04
Rename Public Key Pinning to HTTP Public Key Pinning

FozzieHi:
There is a difference between Public Key Pinning and HTTP Public Key Pinning. Nearly all major browsers still support Public Key Pinning for pinning specific websites which are built into the browser.

Chrome's PKP list - https://chromium.googlesource.com/chromium/src/net/+/master/http/transport_security_state_static.pins
Firefox's PKP list - https://searchfox.org/mozilla-central/source/security/manager/tools/PreloadedHPKPins.json


Fyrd:
Thanks!

Issue #5843, opened 2021-04-05
Test for SVG favicons appears to be broken


Schweinepriester:
<https://tests.caniuse.com/link-icon-svg>

That is, no link appears to be present

<img width="845" alt="Bildschirmfoto 2021-04-05 um 17 03 27" src="https://user-images.githubusercontent.com/2644614/113588841-ef297400-9630-11eb-9494-0ccb02a80b34.png">

(#3196 Fyrd pls 0:-))

Fyrd:
Fixed!

Schweinepriester:
Noice!

PR #5844, opened 2021-04-05
SVG favicons: Add WebKit feature request bug

Schweinepriester:
Coming from <https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs#safari-pinned-icon> I wanted to check Safari (thats why #5843 exists - as a side note, didn't really find a good replacement link) and searched quite a bit since surely there had to be an issue with WebKit.

There is <https://bugs.webkit.org/show_bug.cgi?id=136059>, but the TL;DR is

> So if you're looking for SVG favicon support in Safari, you'll need to file with Apple at developer.apple.com.

But I think it still has value to add that link to spare others the time to search.


Fyrd:
Sounds good, thanks!

PR #5848, opened 2021-04-09
CSS `clip-path`: Add Chromium & WebKit bugs for shapes in external SVGs

Schweinepriester:
- https://bugs.chromium.org/p/chromium/issues/detail?id=109212
- https://bugs.webkit.org/show_bug.cgi?id=104442


Fyrd:
Thanks!

PR #5846, opened 2021-04-07
Chrome: JPEG XL behind a flag

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=1178058#c11>

> Implement image/jxl decoding behind a flag.
>
> Adds support for decoding JPEG XL image format.

PR #5850, opened 2021-04-11
Chrome does not support JPEG XL behind a runtime flag

atjn:
This is a follow-up on #5846

Now that updated versions of Chrome 91 and 92 have been released, I tried to run them on Windows and Linux to check that they actually do support jxl decoding. Unfortunately, none of them expose the `enable-jxl` flag, which means the current note is wrong. I also tried to enable some other flags, like `enable-experimental-web-platform-features`, but nothing seems to enable jxl decoding.

I am far from an expert on building binaries, but I believe this is happening because Chrome needs to be built with the `ENABLE_JXL_DECODER` flag set before it is even possible to enable the flag in `chrome://flags`, and for whatever reason, the Chrome engineers decided not to turn that build flag on yet.

Building Chrome is a monumental task that requires a lot of technical knowledge, so I don't see any reason to bloat the note with specifics on which flags to set, especially when we haven't actually tested that it would work. Let me know if you disagree.

### DIFF
--- a/features-json/jpegxl.json
+++ b/features-json/jpegxl.json
@@ -424,7 +424,7 @@
   },
   "notes":"",
   "notes_by_num":{
-    "1":"Can be enabled via the `enable-jxl` flag in `chrome://flags`"
+    "1":"Can experimentally be build with support for jxl decoding."
   },
   "usage_perc_y":0,
   "usage_perc_a":0,
--- a/features-json/jpegxl.json
+++ b/features-json/jpegxl.json
@@ -424,7 +424,7 @@
   },
   "notes":"",
   "notes_by_num":{
-    "1":"Can experimentally be build with support for jxl decoding."
+    "1":"Can experimentally be built with support for jxl decoding."
   },
   "usage_perc_y":0,
   "usage_perc_a":0,


Schweinepriester:
image

They actually removed it or rather I guess at least todays build doesn't have the flag.

Because I checked yesterday and I could enable `#enable-jxl` and paint <https://tests.caniuse.com/jpegxl> green. But I wasn't able to display the .jxl on its own, it would force the download.

But oh well, I guess it's experimental after all…

atjn:
Interesting, I didn't know that it was available yesterday, I only had time to check today.
I guess it still makes sense to merge this, but after that, maybe we should give it a few weeks to stabilise before making any updates.
It sucks for sure, was looking forward to test it :/

Schweinepriester:
The flag might be coming back: https://bugs.chromium.org/p/chromium/issues/detail?id=1178058#c15
Looking at the [diff](https://chromium.googlesource.com/chromium/src/+/bb58f5954d74ca7fe9e01e130a4a069c4083d045%5E%21/#F0) one could speculate the flag was immediately expired since 91 was the first version with it.

Time will tell of course if the flag is back tomorrow and stays for a bit :D

Schweinepriester:
It's back (for now?):

image

🎉

features-json/jpegxl.json
```diff
@@ -424,7 +424,7 @@
   },
   "notes":"",
   "notes_by_num":{
-    "1":"Can be enabled via the `enable-jxl` flag in `chrome://flags`"
+    "1":"Can experimentally be build with support for jxl decoding."
```
Fyrd, line 427:
```suggestion
    "1":"Can experimentally be built with support for jxl decoding."
```

Fyrd: Small correction :)

atjn:
It seems to be back for me as well. I still like the idea of giving it 1-2 weeks before making an update. Chrome 90 is rolling out right now, so in a week we will have a new beta, new dev and new canary to test on.

EwoutH:
I just downloaded Chrome Beta from [www.google.com/chrome/beta](https://www.google.com/chrome/beta/) to test this again, and at least for Windows it's build by default and can be enabled by the `enable-jxl` flag in Chrome 91 Beta.

Screenshot_456

PR #5852, opened 2021-04-12
Safari on iOS 14.5 supports `gap`

jimmymk23:
Indicated lack of support for `gap` in iOS Safari ends on version 14.4 and starts supporting it on Safari iOS version 14.5


Fyrd:
Thanks, have added iOS 14.5 properly and with it included support for `gap`.

PR #5856, opened 2021-04-16
CSS `display: contents`: Accessibility not yet fixed in Safari

Schweinepriester:
Coming from <https://css-tricks.com/platform-news-using-focus-visible-bbcs-new-typeface-declarative-shadow-doms-a11y-and-placeholders/>

> Until Apple fixes this [bug in Safari](https://bugs.webkit.org/show_bug.cgi?id=185679), be careful when using the `contents` value on semantic elements and test in screen readers to confirm that your pages are accessible in Safari as well.

Which links to the same unfixed bug mentioned in [the article](https://hiddedevries.nl/en/blog/2018-04-21-more-accessible-markup-with-display-contents) in note 2.

> ### Safari
>
> The list shows as 'no accessibility information' ([Safari bug](https://bugreport.apple.com/web/?problemID=39630240), [Webkit bug](https://bugs.webkit.org/show_bug.cgi?id=185679))

Just like [comment 6](https://bugs.webkit.org/show_bug.cgi?id=185679#c6) in the bug brings up, caniuse is almost certainly currently not correct.

Looks like there was a miscommunication in #5232 which made `y` from TP ride to all subsequent versions including iOS (#5449).

There is a test in [comment 4](https://bugs.webkit.org/show_bug.cgi?id=185679#c2), but I haven't tested it yet. But given the severity I think we should go the safer route and mark Safari as partial, at least until it's tested.
That test case could be added to <https://tests.caniuse.com/css-display-contents>.


Schweinepriester:
To comment further on #5232: 
Since only one of the two bugs mentioned in note 3 was fixed for Safari 13, I don't think note 3 should be removed from Safari 13. Technically we could split note 3, one note for each issue, but in the end it's partial support and since both have been fixed in TP (apparently becoming 13.1/iOS 13.4), I don't think there's much value there.

=> I'm going to add note 3 back to Safari 13 and iOS <= 13.4, alright?

Fyrd:
Looks good, thanks!

PR #5864, opened 2021-04-26
Opera Mobile supports AVIF

atjn:
I have tested and confirmed support for AVIF in Opera Mobile 63 with the [caniuse test](https://tests.caniuse.com/avif).

This should also have been supported in Opera Mobile 62, which is currently the version displayed on caniuse.


Fyrd:
Thanks!

PR #5865, opened 2021-04-26
Samsung Internet supports AVIF

atjn:
I have tested and confirmed that Samsung Internet 14.0 supports AVIF with the [caniuse test](https://tests.caniuse.com/avif).

This PR fails because you haven't added Samsung Internet 14 to the support tables yet :)


Fyrd:
Confirmed and already included in my testing of Samsung Internet 14 :)

PR #5866, opened 2021-04-27
iOS 14.5 Classes

robpalme:
iOS 14.5 contains Safari 14.1 which has Private Fields & Methods


Schweinepriester:
The changes are of course still valid and should be applied, but FYI both are currently not shown:

https://github.com/Fyrd/caniuse/blob/97c961d9ec7c742c6c720aafb27c7b59dd2d8e2a/features-json/private-class-fields.json#L437

https://github.com/Fyrd/caniuse/blob/97c961d9ec7c742c6c720aafb27c7b59dd2d8e2a/features-json/private-methods-and-accessors.json#L429

Most likely because at least partially covered via MDN: <https://caniuse.com/mdn-javascript_classes_private_class_fields>

PR #5868, opened 2021-04-29
Denote SXG support in Samsung Internet 11+

twifkak:
SXG was first supported in [Chromium M73](https://chromestatus.com/feature/5745285984681984), and Samsung 11 is the [first version running an equal or higher milestone](https://en.wikipedia.org/wiki/Samsung_Internet).

I've also confirmed that SXG works on my phone on Samsung Internet (currently ver 14, but I tested earlier too).


Fyrd:
Thanks!

PR #5871, opened 2021-04-29
Safari TP 123: Web Animations API composite modes

Schweinepriester:
While I didn't see it in <https://webkit.org/blog/11585/release-notes-for-safari-technology-preview-123/> nor found the ticket, <https://tests.caniuse.com/web-animation> speaks for itself:

Bildschirmfoto 2021-04-29 um 22 15 00

vs. Safari 14.1:

Bildschirmfoto 2021-04-29 um 22 16 16

We should probably check again once Safari 15 is out if its not just TP only for some time longer.


Fyrd:
Confirmed, thanks!

PR #5721, opened 2020-12-24
Clarify VP8/VP9/WebM status on Safari

kiding:
- Separated the duplicate portion of note 4 from note 5.
- Added note 5 to iOS Safari 14.
- Added off-by-default clause to note 5.
- Added note 6 for [VP9 + WebM + MediaSource](https://bugs.webkit.org/show_bug.cgi?id=216652#c1) support on Safari 14 + Big Sur.

Fixes #5713 as YouTube uses MediaSource/MSE for 4K HDR support [on Big Sur.](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes)

PR #5873, opened 2021-04-29
Safari 14.1 updates

Schweinepriester:
<https://webkit.org/blog/11648/new-webkit-features-in-safari-14-1/>

TODO to make sure the changes here are (more) correct:
- [x] <https://tests.caniuse.com/audio-api> on iOS 14.5
- [x] <https://tests.caniuse.com/mediarecorder> on iOS 14.5
- [x] <https://tests.caniuse.com/webm> on macOS 11 Big Sur using Safari 14.1 plus latest TP (CC @kiding because #5721)
- [x] <https://tests.caniuse.com/speech-recognition> on iOS 14.5 + Safari 14.1 on macOS 11 Big Sur (CC @apple-web-evangelist because #5765)

Can't yet test myself, since I'm not on Big Sur for [Xcode 12.5](https://developer.apple.com/documentation/xcode-release-notes/xcode-12_5-release-notes) nor already on iOS 14.5. Browserstack doesn't seem to be there either.


features-json/webm.json
```diff
@@ -255,7 +255,7 @@
       "13":"a #4",
       "13.1":"a #4",
       "14":"a #4 #5 #6",
-      "14.1":"y",
+      "14.1":"a #4 #5 #6",
```
kiding, line 258: image

Regarding webM: Safari 14.1 release notes on Catalina/Mojave suggests the feature is **not** dependent on the OS. I've been setting up my VM to test that got delayed due to work reasons. Let me get back to this soon, but in the mean time I don't mind keeping (reverting?) it partial.

Schweinepriester: Yeah, I'm actually not sure why @Fyrd added 14.1 as supported in https://github.com/Fyrd/caniuse/commit/499c067dc51b8141840847cc52ef365d92c05a73#diff-7f7bcd7c4eeb5170b1aa0c458cdfbc6c9e2708e3daf5f3a6735baa8be1861123, given that TP is in the state you set in #5721, which is still true, isn't it? 
Or is fully supported now?
I can test 14.1 & TP on Catalina, but I'm quite sure that yesterday <https://tests.caniuse.com/webm> didn't work.

Schweinepriester: On Catalina (10.15.7 (19H1030), Mid 2020 Intel MacBook Pro), neither in 14.1 nor TP any of <https://tests.caniuse.com/webm> work:

Bildschirmfoto 2021-05-01 um 00 29 27

kiding: Yep, I did my own testing and it's definitely a Big-Sur-or-later feature.

Could you add Note 7 like WebP,
https://github.com/Fyrd/caniuse/blob/13138dde08de5c00dccdd3ad6b4dcb697c7df9c4/features-json/webp.json#L435

And change the statuses like this?
```suggestion
      "14.1":"a #7",
      "TP":"a #7"
```

kiding (changes requested):

features-json/webm.json
kiding: Yep, I did my own testing and it's definitely a Big-Sur-or-later feature.

Could you add a new note 7 somewhere similar to WebP:
> Partial support in Safari refers to being limited to macOS 11.3 or later.

And change the statuses like this?
```
"14.1":"a #4 #5 #7",
"TP":"a #4 #5 #7"
```

kiding (changes requested):

features-json/webm.json
Fyrd: Yeah, I jumped the gun a bit there on the Safari support, I saw it working on my Big Sur machine and didn't think too hard about it.

Fyrd:
Can confirm `audio-api` and `mediarecorder` support on iOS 14.5

Fyrd:
Can also confirm speechrecognition only has prefixed support.

Schweinepriester:
Thansk @Fyrd for testing :)

features-json/webm.json
Schweinepriester: > Could you add a new note 7 somewhere similar to WebP

Sure!

But are note 4 and 5 still applicable? Or does every video at <https://tests.caniuse.com/webm> just work by default?
Because then it would be more like `a #7` (== partial only because "macOS 11.3 or later"), right?

Fyrd: Yeah, I'm fine with removing notes 4 & 5 since those were only worth mentioning when they were the only types of support.

kiding: There’s one more technicality to think about: VP9 profile 0 and profile 2 in WebRTC. The former is enabled by default in the TP (the note 5) then Safari on macOS Catalina/Mojave would support it too. I’ve been meaning to prepare a PR for this.

I’m not sure if we should track the WebRTC support _at all_ in this feature. Should we simply stop/remove them? This would inevitably make some of Safari’s the WebM support statuses (that rely on the note 4 and/or note 5) to unsupported. The “off-by-default” clause has bothered me anyways.

Schweinepriester: I'm going with `a #7` for now to finally complete this PR, but please follow up @kiding if necessary :)

Schweinepriester:
Thanks again @kiding & @Fyrd! :)

Afaik ready to merge.

Fyrd:
Nice work, thanks!

PR #5881, opened 2021-05-05
Update image-set data

### LINKED ISSUE #5636: Update Chrome/Safari image-set to partial support
I sent this via feedback on the site but posting here for visibility, and adding more details. The info on the `image-set` CSS property is out of date. Chrome and Safari no longer have full support as per the spec:

- Chrome (v85.0) does not support the format `-webkit-image-set("image.jpg" 1x)`, only the form with `url("image.jpg")` - [bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=630597)
- Neither Chrome nor Safari (v14.0) support the MIME type parameter, like `-webkit-image-set(url("image.jpg") 1x type("image/jpeg"))`
- The note listed for Safari should link to [this bug report instead](https://bugs.webkit.org/show_bug.cgi?id=206142)
- There seems to be three separate entries for the same property: [1](https://caniuse.com/css-image-set), [2](https://caniuse.com/mdn-css_types_image_image-set), [3](https://caniuse.com/mdn-css_properties_background-image_image-set).

I can send a PR if desired, but I don't know how you approach reverting support when the spec changes. Do we go back and set all old versions to partial support?

atjn:
The image-set() method has been highly experimental for a long time, but recently had a lot of new additions and standardization. The data in the current caniuse table is mostly based on support for the older, simpler syntax. This PR updates the data to correlate with the new syntax.

I have tested and confirmed all support data with the [web platform tests](https://wpt.fyi/results/css/css-images/image-set/image-set-parsing.html) and some local tests I hacked together to test prefixed browsers.

Notable changes:
- Chromium's implementation is very lacking, so all browsers are degraded to `partial` support.
- Safari had lacking support (so degraded to `partial`), but has almost caught up in v14.
- Firefox's implementation is almost complete, but they are still degraded to `partial` for missing a few things.

Closes #5636


Schweinepriester:
Can you argue why you removed the links http://cloudfour.com/examples/image-set/ and <https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/6606738-image-set>? 0:-)

Probably outdated? But maybe keep them for the history?

Potential merge conflicts looming because image-set changed as well in https://github.com/Fyrd/caniuse/pull/5892/files#diff-89575ac88266aadb08f994f87e4de0a7e5b3e968f462a769cba4b8537296beb2

atjn:
I have recorded my thoughts on the UserVoice link in https://github.com/Fyrd/caniuse/pull/5892#issuecomment-834082837. Depending on what Fyrd decides to do, it should be pretty easy to resolve a merge conflict between these two PR's :)

The cloud four image-set test does not use the unprefixed version of image-set, meaning it doesn't work in Firefox and might also stop working in other browsers at some point. It also doesn't test any of the new features that were recently added to the spec. I didn't want to mislead anyone, so I decided to replace that link with a link to [wpt](https://wpt.fyi/results/css/css-images/image-set/image-set-parsing.html), that I think gives a much better overview of the current status in all browsers.

Schweinepriester:
Alright, that's fair! ✔️

Fyrd:
Very nice, thanks for the research & updates!

PR #5882, opened 2021-05-05
CSS Backdrop Filter: Add WebKit bug to unprefix

Schweinepriester:
For convenience: <https://bugs.webkit.org/show_bug.cgi?id=224899>


Fyrd:
Thanks!

PR #5887, opened 2021-05-05
CSS `text-orientation`: Safari 14+ without prefix 🎉 

Schweinepriester:
I've [test](https://tests.caniuse.com/css-text-orientation)ed 14.1 and TP:

Bildschirmfoto 2021-05-05 um 22 00 58

Changed 14 as well, because it was tested in the MDN PR: https://github.com/mdn/browser-compat-data/pull/6745

---

MDN also has a bit different data for past versions of Safari, esp. Safari on iOS: https://developer.mozilla.org/en-US/docs/Web/CSS/text-orientation#browser_compatibility
But I didn't follow through, because I think the future and recent past are most important 0:-)


Fyrd:
Nice, thanks!

PR #5888, opened 2021-05-05
Updates and more links following "Is Vendor Prefixing Dead?"

Schweinepriester:
Checked the table under <https://css-tricks.com/is-vendor-prefixing-dead/#prefixing-in-2021> and updated a bit plus added links where applicable.

#5882, #5883 & #5887 should have been part of this too, sorry 0:-)

### CSS3 font-kerning

[Test](https://tests.caniuse.com/font-kerning)ed Safari on iOS only using 14.5.

<details>
 <summary>Safari on iOS 14.5 screenshot</summary>

 1525D11D-582E-4489-A570-22604F66BCB8
</details>

MDN says it's supported way longer though: <https://developer.mozilla.org/en-US/docs/Web/CSS/font-kerning#browser_compatibility>


Fyrd:
Thanks! Just tested older iOS Safari and noticed proper unprefixed support starting in 12 so updated the PR with that. :)

PR #5889, opened 2021-05-05
Edge 90: CSS Hyphenation still not on par with Chrome

Schweinepriester:
Per https://github.com/Fyrd/caniuse/issues/5792#issuecomment-833054906 the [state](https://tests.caniuse.com/css-hyphens) is still the as it was in #5793.

<details>
 <summary>Edge 90 screenshot</summary>

 image
</details>

Looks like Edge 90 was correctly added in https://github.com/Fyrd/caniuse/commit/6b4e9f5ebdf577b33297f6d8c3ad08f3f03586be#diff-e0d49bfc31ca78f3aa16681547cf2dad43fa41b07e281514e0b77318b6a67524R58, but overridden in https://github.com/Fyrd/caniuse/commit/60f111be8a6f2f907f5ff3fc41e53696ed3e43bc#diff-e0d49bfc31ca78f3aa16681547cf2dad43fa41b07e281514e0b77318b6a67524R58

If I had to guess, I'd say the "create new browser version" script respects previous states, while the "sync chromium browsers with chrome" script doesn't? 0:-)

EDIT:

Opera 75 & 76 is the same as Edge.

<details>
 <summary>Opera 76 screenshot</summary>

 image
</details>


Fyrd:
Yeah, it's possible I just did a quick test on my Mac and forgot about the note so I assumed it was just supported 🤦  Thanks for the fix!

PR #5894, opened 2021-05-09
Add Firefox EventSource service worker bug

FlyOrBoom:
Firefox currently does not support [EventSource in service workers](https://bugzilla.mozilla.org/show_bug.cgi?id=1681218)


Fyrd:
Thanks!

PR #5895, opened 2021-05-10
Speech Recognition API: More info for Safari

### LINKED ISSUE #5879: Safari SpeechRecognition API
According to this bug, https://bugs.webkit.org/show_bug.cgi?id=225298#c3

> SpeechRecognition API is not available in SafariViewController and web apps added to Home Screen for now
> There are some implementation details we need to figure out before we can enable it.

please update the page to reflect that
https://caniuse.com/?search=speechrecognition

thanks

Schweinepriester:
Fixes #5879.

… or do you want me to split the note, because it's getting lengthy?


meliixo:
Sorry what do U mean

Fyrd:
It is getting lengthy! But it's probably fine.

PR #5899, opened 2021-05-11
Update Opera Mini custom property support

bogger33:
Just tested this, opera mini supports custom properties in the current version


bogger33:
ah nevermind, just saw this discussed in pull request #4208

PR #5902, opened 2021-05-12
Firefox 88 doesn't appear to have HTTP/3 enabled by default

Schweinepriester:
I saw

- https://bugzilla.mozilla.org/show_bug.cgi?id=1710816
- https://phabricator.services.mozilla.com/D114956

and checked my Firefox, which didn't have HTTP/3 enabled by default.

<details>
    <summary>Screenshot of Firefox 88.0.1 showing <code>network.http.http3.enabled</code> set to <code>false</code> by default.</summary>

 image
</details>

Additionally tested using <https://blog.cloudflare.com/how-to-test-http-3-and-quic-with-firefox-nightly/>.

Might not even enabled by default in Firefox 89…

CC @bazzadp because #5836.

---

<https://tests.caniuse.com/http3> seems to be missing a test or a link to one 0:-)


bazzadp:
It's rolling out. I've got it set on mine:

image

From the Mozilla blog: https://hacks.mozilla.org/2021/04/quic-and-http-3-support-now-in-firefox-nightly-and-beta/

> tl;dr: Support for QUIC and HTTP/3 is now enabled by default in Firefox Nightly and Firefox Beta. We are planning to start rollout on the release in Firefox Stable Release 88. HTTP/3 will be available by default by the end of May.

And while I've seen no further official word from Mozilla, members of the HTTP community are noticing it rolling our as per above schedule: https://twitter.com/bagder/status/1389915444145008642?s=20

So it will be enabled for Firefox 88 for all users but might take a few more weeks for that to happen.

I'm not sure what caniuse's policy is for these features that are enabled remotely without an update. At the moment it's on for some but not for others so neither is right. But that's a short term issue. Personally I'd keep it as is - it'll be resolved soon.

Oh and BTW, when I first added this for Firefox 88 in #5836 , Firefox 87 was the production version and the dev and beta channels for 88 did have it enabled by default.

Schweinepriester:
Alright, fair enough!

> And while I've seen no further official word from Mozilla, members of the HTTP community are noticing it rolling our as per above schedule: https://twitter.com/bagder/status/1389915444145008642

Thanks for the link, there's [a reply](https://twitter.com/alpagot/status/1390053500013449218) linking to <https://http3.is/> which would be even better for <https://tests.caniuse.com/http3>, I think :)

> I'm not sure what caniuse's policy is for these features that are enabled remotely without an update. At the moment it's on for some but not for others so neither is right. But that's a short term issue. Personally I'd keep it as is - it'll be resolved soon.

I'll leave the PR open and let @Fyrd decide, but now I'm fine with either way :)

> Oh and BTW, when I first added this for Firefox 88 in #5836 , Firefox 87 was the production version and the dev and beta channels for 88 did have it enabled by default.

Yeah, but the [previous state](https://hg.mozilla.org/try/rev/6eedd6a47d63eb1dbc66a06d9d64a79f59c64982) makes it look like that it was dev/beta only:

```
// Http3 parameters
#if defined(EARLY_BETA_OR_EARLIER)
pref("network.http.http3.enabled", true);
#else
pref("network.http.http3.enabled", false);
#endif
```

bazzadp:
Yeah I was just pointing out why I set it for 88 when I did :-)

HTTP/3 usage (and so testing) is gonna be messy to define. Unlike HTTP/2 where we upgrade HTTP/1.1 to HTTP/2 when it's supported, HTTP/3 is much more interesting as it's not done over TCP so a completely separate connection. So, first of all your browser needs to know whether to try that, and most will just default to TCP (so HTTP/1.1 or HTTP/2) initially, and then see from `alt-svc` header for next connection. Which is why that [https://http3.is/](https://http3.is/) test page will initially load without HTTP/3.

But even after that initial load then you get into all sorts of heuristics and fallbacks while it races connections and decides which to use (kind of like IPv4 and IPv6).

So don't think there is a clean "test" for this.

Schweinepriester:
In 89.0.1 it's now enabled by default for me as well - but 88 wasn't till the end.

But since 88 isn't current anymore I guess my change is now less crucial...

@Fyrd, I think you can decide to close or merge :)

PR #5904, opened 2021-05-13
AVIF is still disabled in FF89+

saschanaz:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1682995#c47 and https://bugzilla.mozilla.org/show_bug.cgi?id=1706160.


Fyrd:
Thanks!

PR #5349, opened 2020-04-07
Remove inaccurate note and Safari support for ShadowDOM

justinfagnani:
I tested selectors like `:host > .foo` in several versions of Safari, including 10.0 and they all worked. I'm not sure where this comment originated from, but without automated tests or a bug reference to back up the note, it appears to be wrong.

Here are screenshots of a simple selector test working in Safari 10 on Sauce:

Screen Shot 2020-04-07 at 2 48 06 PM
Screen Shot 2020-04-07 at 2 48 13 PM

PR #2787, opened 2016-09-07
Adds Custom elements v1, shadow dom v1, and URLSearchParams

ebidel:
R: @Fyrd

Also updates BroadcastChannel API

PR #5906, opened 2021-05-17
Shadow DOM (V1): Remove Note 1 from Safari on iOS

Schweinepriester:
Basically another attempt to get the state correct or rather follow-up #5349.

Note 1 was intially added in #2787 for Safari 10 and TP: https://github.com/Fyrd/caniuse/pull/2787/files#diff-2d56820274d456cb75cbe61fa96708ba772ded175a5fabc58dfc5417b9b4106aR162

From there it made it's way to Safari on iOS.

Then removed in #5349, entirely for Safari only, not Safari on iOS, though.

Therefore it returned in
https://github.com/Fyrd/caniuse/commit/331389b6e2cf6388a8560faeb37a6e72e95b4043#diff-2d56820274d456cb75cbe61fa96708ba772ded175a5fabc58dfc5417b9b4106aL378 
because it was still needed.

I've now gone ahead and removed the note from Safari on iOS, except for 10.x where the plausibility is highest.

I've tried to find some bugs, but only came across some possible candidates (only for `::slotted`):
- https://bugs.webkit.org/show_bug.cgi?id=154765
- https://bugs.webkit.org/show_bug.cgi?id=149441

CC @justinfagnani because of #5349.

@ebidel: Do you remember anything regarding this note for Shadow DOM (V1)?
> Certain CSS selectors do not work (`:host > .local-child`) and styling slotted content (`::slotted`) is buggy.


ebidel:
Sorry, don't remember the details.

Schweinepriester:
Can't expect that of course - thanks for replying! :)

Fyrd:
Good research. :) Thanks!

PR #5908, opened 2021-05-17
Add note to HLS

### LINKED ISSUE #5905: HLS HTTP Live Streaming info outdated?
I'm opening an issue instead of a pull request because I want to be sure I'm not misunderstanding anything.
According to [caniuse](https://caniuse.com/?search=hls), HLS is not supported on Edge, Chrome and Firefox.

However, according to the linked [Wikipedia article](https://en.wikipedia.org/wiki/HTTP_Live_Streaming) to which it refers, HLS is supported with media source extensions on Chrome and FF (and presumable Edgium too?).
Almost every video platform, CDN, etc. seems to prefer to use HLS too, even for Chrome and Firefox clients.

Should that not warrant a "partial" indicator or a note?

Schweinepriester:
Fixes #5905.

Per your comment there. Also improved the link for DASH to skip the redirect.

### DIFF
--- a/features-json/http-live-streaming.json
+++ b/features-json/http-live-streaming.json
@@ -418,7 +418,7 @@
       "2.5":"n"
     }
   },
-  "notes":"",
+  "notes":"HLS can be used with a JavaScript library in browsers that doesn't support it natively as long as they support [Media Source Extensions](/mediasource).",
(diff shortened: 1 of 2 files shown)


Fyrd:
Nice, thanks!

PR #5910, opened 2021-05-18
Added a workaround for IE issue with border-collapse + box-shadow

ribeirobreno:


Fyrd:
Hmm, I see "This debug view expired" when following that link. https://codepen.io/ribeirobreno/pen/PopbqKo works though, maybe change to that instead?

ribeirobreno:
Thanks! Updated the URL. :)

Fyrd:
Thanks!

PR #5911, opened 2021-05-20
safari does not support background-attachment since version 14 anymore

fabb:
tested as broken on iPhone iOS 14 and macOS Safari 14

tested as working on macOS Safari 13.

see also https://github.com/mdn/browser-compat-data/pull/10542


fabb:
it looks like `fixed` is somewhat broken on Safari 14 and iOS Safari 13. the background image is not recalculated for the area outside of the current visible part of a scrolling container. when scrolling, the remaining part just has a blank background. looks like an "optimization" of Safari for scrolling.

background-attachment-safari-14

IMHO that would count as not supported, WDYT?

Fyrd:
Yeah, I agree, that's pretty bad. I'd be okay with changing it to not supported and maybe add a new note saying there's buggy behavior when attempting to use it.

Schweinepriester:
~~Are there one or more bugs in the WebKit Bugzilla? Should we open them if not?~~

There seem to be bugs ([search](https://bugs.webkit.org/buglist.cgi?quicksearch=background-attachment&list_id=7185744)), which could be useful to link:
- local: https://bugs.webkit.org/show_bug.cgi?id=219324
- fixed: maybe(!), might be a different one: https://bugs.webkit.org/show_bug.cgi?id=204574

fabb:
The webkit bug says it works on safari 13, i need to test this more in depth with browserstack.

fabb:
ok tested on macOS Safari 13, works there. then i'd mark it as unsupported with a comment for Safari 14 only.

on iOS Safari 13 it's the same as on iOS Safari 14, e.g. `fixed` has the same issue as macOS Safari 14:
ios13
also `local` does not work correctly, the salamander location is not recalculated when scrolling the outer scroll area. so i'd say it's completely broken on iOS Safari 13+14.

fabb:
i updated the PR according to my findings. please verify.

Fyrd:
Looking good, thanks!

PR #5912, opened 2021-05-21
Safari 14.1 doesn't support focus-visible

vwkd:
Just tested.


Fyrd:
I did too, but wasn't 100% sure my test was right...I'll take this as confirmation. :) Thanks!

PR #5914, opened 2021-05-22
This feature is depreciated as of Chrome 65

ozgur-dogan:
https://developers.google.com/web/updates/2018/02/chrome-65-deprecations#block_cross-origin_wzxhzdk5a_download


Fyrd:
It's only been deprecated for cross-origin downloads, same-origin downloads work fine. Maybe we should add a note for this instead?

Issue #2458, opened 2016-04-19
Multicolor Glyph support (COLR/CPAL)


xyproto:
Some browsers supports multicolored glyphs, such as the "COLR/CPAL" format.

When can I use that?

Here is one test: https://pixelambacht.nl/demo/color-font-test.html
And here is an article: https://pixelambacht.nl/2014/multicolor-fonts/

Cheers!

eosrei:
+1 

> The vote count is based on +1s in each issue's comments

src: http://caniuse.com/issue-list

Crissov:
May be relevant for emoji support. #1522

tobireif:
Newer URL: https://pixelambacht.nl/chromacheck/

tobireif:
+1

tphinney:
This isn't a subset of #1018 , it's the exact same thing.

Crissov:
@tphinney is right unless we treat this issue as limited to COLR/CPAL.

tphinney:
Well, the request was "such as" not "specifically."
Since COLR/CPAL is not the best supported color font format, and as support for the four variant formats is quite different, but ~ similar in extent (although the smart money would be betting on SVG OpenType in the long run), it would be silly for CanIUse to choose to enumerate the COLR/CPAL format without the others, at this time.

Fyrd:
Re-purposing this issue to be specifically for COLR/CPAL, subset of #1018

yisibl:
> but ~ similar in extent (although the smart money would be betting on SVG OpenType in the long run), it would be silly for CanIUse to choose to enumerate the COLR/CPAL format without the others, at this time.
@tphinney 

COLR/CPAL is currently compatible with almost all mainstream browsers, which is a very important feature. [As Chrome will soon support the COLR v1 specification](https://groups.google.com/a/chromium.org/g/blink-dev/c/XIXudwZdr44), it will also support gradients. It is very reasonable that we separately list the compatibility of COLR/CPAL.

yisibl:
I am going to submit a PR for the COLR table, who can help review?

tphinney:
> COLR/CPAL is currently compatible with almost all mainstream browsers, which is a very important feature.

It’s as if things have changed since 2017! No question that COLR/CPAL is the leading color font format. With support for gradients coming, that should cement the lead.

[...]

PR #5915, opened 2021-05-24
New: add COLR/CPAL font formats

yisibl:
The COLR/CPAL format is already very popular, and we should support it as soon as possible.

## Test Case
- https://at.alicdn.com/t/project/2433183/81b0b90c-6c73-4541-8f1e-3814b86099e1.html

## Screenshot
More: https://www.yuque.com/u7268/iconfont/colr-test

image

## Native support
* Windows 8.1+, [Doc](https://docs.microsoft.com/zh-cn/windows/win32/directwrite/what-s-new-in-directwrite-for-windows-8-consumer-preview?redirectedfrom=MSDN#rendering-color-fonts)
* macOS High Sierra(10.13)+
* iOS 11+

## Browser support test tool

- https://pixelambacht.nl/chromacheck/

## Chromium engine browser
- Chrome 71 supports COLR fonts https://www.chromestatus.com/feature/5897235770376192
- Opera: 58 ≈ Chrome 71 https://help.opera.com/en/opera-version-history/#58
- Samsung Internet: [10.2.01.10 ≈ Chrome 71](https://en.wikipedia.org/wiki/Samsung_Internet#:~:text=version%2010.2.01.10)
- Microsoft Edge: The first version of Microsoft Edge based on the Chromium engine was Chromium 79, so it supports COLR fonts. https://en.wikipedia.org/wiki/Microsoft_Edge#Anaheim_release_history

Fixes: https://github.com/Fyrd/caniuse/issues/2458

I will submit the [COLRv1 format](https://github.com/googlefonts/colr-gradients-spec) in a future PR.


features-json/colr.json
```diff
@@ -0,0 +1,445 @@
...
+      "61":"n",
+      "62":"n",
+      "63":"n",
+      "64":"n",
+      "65":"n",
+      "66":"n",
+      "67":"n",
+      "68":"n",
+      "69":"n",
+      "70":"n",
```
yisibl, line 220: @drott Is it supported before Chrome 70?

yisibl: In addition, where can I see the version of Opera corresponding to Chrome 70?

yisibl:
@Fyrd PTAL

features-json/colr.json
drott: CFF2, SBIX, CBDT/CBLC are supported [since 66](https://storage.googleapis.com/chromium-find-releases-static/7cd.html#7cd32864f21b540832c3297c68d9473f465e6341), 
COLR/CPAL is universally supported since commit 7a95ad3567708717416e53bbd9d65db9ba2bc4f1, [which means 71](https://storage.googleapis.com/chromium-find-releases-static/7a9.html#7a95ad3567708717416e53bbd9d65db9ba2bc4f1). So marking `"71"` here as `"y"` looks good to me.

See also https://bugs.chromium.org/p/chromium/issues/detail?id=882844

drott: > In addition, where can I see the version of Opera corresponding to Chrome 70?

Seems to be Opera 57 according to https://help.opera.com/en/opera-version-history/

yisibl: Thank you so much, I updated Opera's data.

yisibl: > See also https://bugs.chromium.org/p/chromium/issues/detail?id=882844

Yes, I added this bug URL in `notes_by_num 3`.

features-json/colr.json
```diff
@@ -0,0 +1,449 @@
...
+      "5.5":"n",
+      "6":"n",
+      "7":"n",
+      "8":"n",
+      "9":"n",
+      "10":"n",
+      "11":"n"
+    },
+    "edge":{
+      "12":"y",
```
yisibl, line 45: COLR is already supported in Windows 8.1, so I believe Edge also supports it. https://www.theverge.com/2015/4/29/8511169/microsoft-edge-official-name-internet-explorer-upgrade

features-json/colr.json
yisibl: @drott I tested it in Windows 10 and even Chrome 55 supports it. This is rendered by directwrite, right?
Screenshot: https://www.yuque.com/u7268/iconfont/colr-test#UHzkk

drott: > @drott I tested it in Windows 10 and even Chrome 55 supports it. This is rendered by directwrite, right?
> Screenshot: https://www.yuque.com/u7268/iconfont/colr-test#UHzkk

Yes, single platform, if the underlying platform rasteriser supports it - so on Windows 10 earlier versions of Chrome supported it. Universal support through FreeType from 71. You could put a foot note for "Windows 10 from 55" but general support was available only from 71.

yisibl: Thanks!

Fyrd:
Looks good, thanks!

PR #5918, opened 2021-05-26
fix: Firefox 32- does not support colr fonts

yisibl:
Screenshot: https://www.yuque.com/u7268/iconfont/colr-test#PKE4I
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=889401#c41


yisibl:
I will update the Chrome data in the next PR.

> On https://codereview.chromium.org/1984943002/ Color Emoji support is added to Chrome, great step forward but it is tied to platform ability on rendering
https://bugs.chromium.org/p/chromium/issues/detail?id=617379

Fyrd:
Thanks!

PR #5920, opened 2021-05-26
CSS3 3D Transforms: Adjust note

Schweinepriester:
Adjust note 2 for Safari to include Safari 14 and future versions, to stop to have to adjust that with every new version.

Checked Safari 14, still the case:

Bildschirmfoto 2021-05-27 um 00 23 20


Fyrd:
Nice, thanks!

PR #5923, opened 2021-05-29
Update Mozilla link in canvas-text.json

devinmcinnis:
Followed the redirect via http://web.archive.org/web/20190303013144/https://developer.mozilla.org/en-US/docs/Drawing_text_using_a_canvas

image


Fyrd:
Thanks!

PR #5933, opened 2021-06-04
Rename font-rendering controls to font-display

### LINKED ISSUE #5931: font-display?
I went to find the `font-display` table on Can I Use. And there isn't one. https://caniuse.com/?search=font-display

I started to think I should open an issue about adding it to Can I Use, but then realized oh, this is the kind of data that should now get pulled in from MDN. And MDN does have a compat table: https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

But somehow it doesn't render at Can I Use. Why is this data not migrating over?

Schweinepriester:
Fixes #5931.

Per discussion there first try renaming it here.
I thought the short "CSS `font-display`" title would be fitting for caniuse, mentioning `@font-face` just below that.

Feel free to suggest changes or make them yourself :)

CC @jensimmons

### DIFF
--- a/features-json/css-font-rendering-controls.json
+++ b/features-json/css-font-rendering-controls.json
@@ -1,6 +1,6 @@
 {
-  "title":"CSS font-rendering controls",
-  "description":"`@font-face` descriptor (currently defined as `font-display`) that allows control over how a downloadable font renders before it is fully loaded.",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #5935, opened 2021-06-04
Update css-descendant-gtgt.json

limeyboi01:
Tested whether >> CSS selector works in Chrome 92 and it does not.
I was told that this syntax is deprecated and probably won't be included in future updates, so I also marked it as unsupported in Chrome 93 and 94.


meliixo:
Unsubscribe I don't no this

Schweinepriester (approved):

Schweinepriester:
Since that syntax is deprecated and Chrome won't implement it I think it we can change 93 and 94 to unsupported as well, if you want to, @limeyboi01 :)
- https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator#browser_compatibility
- https://bugs.chromium.org/p/chromium/issues/detail?id=446050#c17

Fyrd:
Thanks!

PR #5942, opened 2021-06-10
CSS3 `tab-size` - Unprefixed in Firefox

Schweinepriester:
<https://bugzilla.mozilla.org/show_bug.cgi?id=737785#c36>

> Un-prefix -moz-tab-size (keeping the prefixed version as an alias for now).

We'll have to wait for the next build(s) to check precisely.

<https://tests.caniuse.com/css3-tabsize>


Fyrd:
Thanks!

Schweinepriester:
Now confirmed as well:

image

PR #5946, opened 2021-06-12
Align Logical Properties & `theme-color` in Safari to MDN

Schweinepriester:
Checked all the changes in https://github.com/mdn/browser-compat-data/pull/10898 against caniuse and Logical Properties & `theme-color` have own entries, so aligning those.
Also: 🎉 

<https://tests.caniuse.com/css-logical-props> is definitely already in TP:

Bildschirmfoto 2021-06-12 um 03 52 32

Bildschirmfoto 2021-06-12 um 03 52 51

TP 125 got the last one missing compared to Chrome green as well, so full support

Since I haven't got Safari 15 or rather macOS 12 I wasn't able to actually test <https://tests.caniuse.com/meta-theme-color> and maybe TP with that isn't available yet, but I have zero doubt it will be available if it isn't already. And per <https://developer.apple.com/videos/play/wwdc2021/10029/> I've gone full support, at least for now.


Fyrd:
Sounds reasonable to me, thanks!

PR #5948, opened 2021-06-12
Variable fonts: Drop note 7 for Safari 15+

Schweinepriester:
Drop note 7 for Safari 15+ since TP [already isn't](https://developer.apple.com/safari/download/) and Safari 15 almost certainly will not be available for macOS 10.14 Mojave, given [the history](https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility).


Fyrd:
Thanks!

PR #5952, opened 2021-06-20
Update audio-api.json

hoch:
Web Audio API is not a W3C Recommendation:
https://www.w3.org/TR/webaudio/

W3C Press release:
https://www.w3.org/2021/06/pressrelease-webaudio.html.en

### DIFF
--- a/features-json/audio-api.json
+++ b/features-json/audio-api.json
@@ -2,7 +2,7 @@
   "title":"Web Audio API",
   "description":"High-level JavaScript API for processing and synthesizing audio",
   "spec":"https://www.w3.org/TR/webaudio/",
-  "status":"wd",
+  "status":"recommendation",
   "links":[
     {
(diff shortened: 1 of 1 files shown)


features-json/audio-api.json
```diff
@@ -2,7 +2,7 @@
   "title":"Web Audio API",
   "description":"High-level JavaScript API for processing and synthesizing audio",
   "spec":"https://www.w3.org/TR/webaudio/",
-  "status":"wd",
+  "status":"recommendation",
```
Fyrd, line 5:
```suggestion
  "status":"rec",
```

Fyrd:
Thanks! I've already fixed it but for future reference the value for recommendation should be `rec`.

hoch:
Thanks for the quick response, @Fyrd! 👍

PR #5954, opened 2021-06-23
Fix MDN Docs resource URL

celsobessa:
MDN Docs resource moved to https://developer.mozilla.org/en-US/docs/Web/API/Navigation_timing_API


Fyrd:
Thanks!

PR #5955, opened 2021-06-24
Fix info about CFF2 in Firefox, add v-fonts.com to links

nicksherman:
See discussion about CFF2 support in Firefox here: https://bugzilla.mozilla.org/show_bug.cgi?id=1598072#c2


Fyrd:
Confirmed, thanks!

PR #5962, opened 2021-07-01
Add Temporal support table

### LINKED ISSUE #5961: Add Entry for Temporal
Is it too early for caniuse to provide an entry for Temporal?

- https://tc39.es/proposal-temporal/docs/
- https://2ality.com/2021/06/temporal-api.html

atjn:
`Temporal` is a complete overhaul of the `Date` API, which has been a massive pain for developers ever since JavaScript was a thing.

It is a little early to add this table, since Temporal still isn't supported anywhere, but given the massive amount of work put into it, the attention it has gotten from developers, and the fact that all browsers have already begun implementing it, I think it's safe to assume it will be a thing soon, and when it eventually lands in a browser, it would be nice to already have the support table + tests ready for the surge of developers trying to find out when/where they can use it :)

I suspect Temporal will also be added to MDN, but this is one of those massive API's where I think adding a native Caniuse table is worth it because it gives a better overall idea of the implementation status.

For now, I think a basic test is all we need on [tests.caniuse.com](https://tests.caniuse.com/):
```js
if(Temporal){
  //Temporal is supported
}
```

Closes #5961


Fyrd:
Thanks!

PR #5966, opened 2021-07-08
Update to add note

ohiowebpro:
It is important to note that the option is set in the OS and that the OS may not support this feature, thus the browsers will not suppor.


Fyrd:
Sounds reasonable, thanks!

PR #5978, opened 2021-07-22
Classify Firefox support for `image-set()` as complete

atjn:
In https://github.com/Fyrd/caniuse/pull/5972/, @yisibl pointed out that `image-set` in Firefox actually _does_ support generated images, but just not in the `content` property.

I initially marked Firefox support as `partial` because I thought it had no support for generated images whatsoever, but now I believe it should instead be marked as `complete`. The `content` property and the generated images are already both niche use cases for `image-set`, so while I think we should keep the comment, I see no reason that this minor issue should classify the entire feature as partially supported :)

I would've added these changes directly to https://github.com/Fyrd/caniuse/pull/5972/, but the branch was too old, so I had to create a new branch and open this PR instead.


Fyrd:
Sounds good, thanks!

yisibl:
Thanks!

PR #5979, opened 2021-07-24
Opera supports JPEG XL behind a flag

atjn:
As expected, coming from Chromium, Opera now also has support for JPEG XL behind a flag.

I have tested and confirmed support with the [caniuse test](https://tests.caniuse.com/jpegxl).


Fyrd:
Thanks! It is worth noting that `chrome://flags` works in both Opera and Edge too (as alias) so doesn't hurt too much to keep that. :)

PR #5986, opened 2021-07-29
Updates to favicon support

### LINKED ISSUE #5982: PNG favicon support incorrect for Safari (for macOS) and Safari for iOS
Currently, the support table for [PNG favicons](https://caniuse.com/link-icon-png) says that all versions of (desktop) Safari support PNG favicons (always preferring ICO if available) and that all versions of Safari for iOS do not use favicons at all (but "may have alternative for bookmarks, etc.").

Safari 12 added a new "Show icons in tabs" option for both [macOS Mojave](https://www.macworld.com/article/231533/article-606.html) and [iOS 12](https://www.macworld.com/article/231560/ios-12-how-to-enable-favicons-site-icons-on-tabs-in-safari.html) so it appears both columns of the support table are incorrect.

(I would open a PR to fix this but I don't have the capacity right now to test ICO/PNG preference in Safari on both platforms.)

atjn:
A few updates to the favicon support tables:

Safari 14+ has started showing favicons in tabs both on MacOS and iOS. The feature seems to have been available since Safari 12 (in the settings menu), but has been disabled by default until v14. I was able to completely confirm this for Safari on MacOS, but not completely for Safari on iOS. There are however numerous articles online which confirm that this feature was also added to Safari on iOS in v12.
I was able to use [the caniuse test](https://tests.caniuse.com/icon-test.php) to confirm that Safari for iOS supports PNG favicons, and that none of the Safari browsers support SVG favicons.

SVG support is currently marked `partial` in both Safari browsers because they have a weird non-standard implementation of pinned tabs where you can supply a very specific type of SVG to be used as an icon. Marking that as `partial` made sense at the time, because the browsers didn't have a normal implementation of favicons, but now that they do use normal favicons the way you would expect a browser to use favicons, I think it is misleading to say that they have `partial` support for SVG icons when the normal implementation has no support for SVG whatsoever.

I have tested and confirmed that newer versions of Samsung Internet do support SVG favicons. I wasn't able to exactly test which version added support, so I have based that on when SVG favicon support was added in Chromium.

I did not spend any time testing whether the browsers prefer PNG over ICO because [the caniuse test](https://tests.caniuse.com/icon-test.php) doesn't give me all the options I need to do that. I am however pretty sure that some of the notes we have right now are wrong. I am not sure how much value these notes bring to developers, and when also considering how much work it is for us to properly maintain them, I might suggest that we just remove them?

Closes #5982


Fyrd:
Thanks! I agree, looking back at these notes they were intended to be helpful to determine which format should be used & included , but I imagine linking to an article with cross-browser recommendations would be better ([this one](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7) perhaps).

Will merge the changes for now, thanks!

PR #5987, opened 2021-08-06
Firefox supports AVIF by default

atjn:
Firefox has enabled AVIF by default for version 92: https://hg.mozilla.org/integration/autoland/rev/bb07fcce3567

They've already tried to do this ~5 other times, and then eventually backtracked because their implementation wasn't ready, but this time they've fixed all major issues _before_ enabling by default, so I am pretty confident that AVIF will in fact be enabled by default this time 🥳

At some point during development, support for sequence images was added (confirmed with [the caniuse test](https://tests.caniuse.com/avif)). I don't see any reason to keep a note about sequence support when it only applies to early development implementations that are hidden behind flags, so I have removed that note entirely.

I will be following this closely and will commit a new patch if they do end up scrapping it again. We can't really know for sure until it actually lands in stable 92.

### DIFF
--- a/features-json/avif.json
+++ b/features-json/avif.json
@@ -141,22 +141,22 @@
       "74":"n",
       "75":"n",
       "76":"n",
-      "77":"n d #1 #2",
-      "78":"n d #1 #2",
-      "79":"n d #1 #2",
-      "80":"n d #1 #2",
-      "81":"n d #1 #2",
-      "82":"n d #1 #2",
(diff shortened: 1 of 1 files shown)


Schweinepriester (approved):

Schweinepriester:
> 6 minutes ago 

Damnit, too slow this time! :D

Good changes :)

Fyrd:
Thanks!

PR #5988, opened 2021-08-06
Improve favicon notes

atjn:
This is a follow-up to https://github.com/Fyrd/caniuse/pull/5986#issuecomment-892348557

A lot of the current notes on favicon support are not maintained properly, and only cover a small part of the crazy complicated landscape that is favicons in browsers. In this PR, I am removing many of these notes and instead linking more prominently to [an excellent guide on web.dev](https://dev.to/masakudamatsu/favicon-nightmare-how-to-maintain-sanity-3al7), that covers things in much more depth.

I am happy to discuss details :)


Fyrd:
Nice work, thanks!

PR #5991, opened 2021-08-10
Update avif.json

pooyaEst:
Firefox 93 released in nightly channel also does support avif


Fyrd:
Thanks! This and the othe features work in the default version on iOS and Android, but not the proxy version which is used by the majority of Opera Mini users on feature phones, which is what is tracked on caniuse.

As shown on the site:
Screen Shot 2021-08-11 at 7 54 18 PM

pooyaEst:
Hi, This pull request was for Firefox 93 not opera mini

Fyrd:
Ah you're right, looks like I put the comment in the wrong place. The Firefox 93 data has also been updated though, so no need for this PR. Thanks!

PR #5993, opened 2021-08-10
opera mini does support avif

pooyaEst:
opera mini does support avif


pooyaEst:
i've checked it on android phone using https://libre-software.net/avif-test/ 
is a screenshot needed ?

Fyrd:
Closing as per https://github.com/Fyrd/caniuse/pull/5991#issuecomment-897307885

pooyaEst:
after testing opera mini with different proxy settings i noticed that it only falls back to jpeg with Extreme data saving settings enabled and (probably only if there is a jpeg fallback) i guess it does worth to mark it as supported or maybe partial support.
nevertheless  opera mini does support the picture source structure as it does not have much to do with proxy settings.

PR #6007, opened 2021-09-01
Removing a bad example from svg-img

melvinsh:
Hi!

Removing a bad example here because this blog links to a missing SVG, which is very misleading in this context! :D 

Screen Shot 2021-09-01 at 9 58 18 PM

Page: https://caniuse.com/svg-img (Resources).


Fyrd:
Fair enough, thanks!

PR #6015, opened 2021-09-08
Firefox 93: AVIF = yes

patryk:
According to https://bugzilla.mozilla.org/show_bug.cgi?id=1682995 and beta release notes.


Fyrd:
Guess this might change again depending on whether or not https://bugzilla.mozilla.org/show_bug.cgi?id=1729071 is fixed in time, but will merge for now. Thanks!

PR #6022, opened 2021-09-16
Add support data for Cascade Layers

### LINKED ISSUE #6000: Support for CSS Cascade Layers
It looks like both [Safari](https://bugs.webkit.org/attachment.cgi?id=436491&action=diff#a/Source/WTF/Scripts/Preferences/WebPreferencesExperimental.yaml_sec1) and [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1095765) are planning to release this [new feature](https://drafts.csswg.org/css-cascade-5/#layering) soon, behind experimental feature flags.

One Chrome caveat: 

> We will not implement the 'revert-layer' keyword due to its complexity and likely negative performance impact.

And one from Safari:

> This patch adds initial support for @layer rules, including both the block and the statement syntax. No support for @import or CSSOM yet but basic functionality mostly works.

The chrome issue seems to be intentional. The Safari issue looks temporary, so we'll have to see what lands.

mirisuzanne:
Fixes #6000 

This is my first contribution, so let me know if there are improvements I can make. Cheers!

See http://wpt.live/css/css-cascade/ for various web platform tests. Cascade Layer tests all begin with the `layer-` prefix.


Fyrd:
Looks great, thanks for contributing @mirisuzanne!

Fyrd:
@mirisuzanne FYI I updated it to include the runtime flag support in Canary

mirisuzanne:
Thanks! Yeah, I wasn't sure if a runtime flag was public enough to mention. Sounds good. :)

PR #6023, opened 2021-09-16
Update Container Query specification and syntax details

mirisuzanne:
- Links to the actual draft specification
- Update description for clarity, removing mention of `contain` property from old syntax
- Note the change in syntax between Chrome 92 & 93


Fyrd:
I see the latest draft still refers to the `contain` property in a few places so I was a bit confused at first, but I'm guessing that just needs to be updated?

Anyway, thanks for the updates!

mirisuzanne:
We're still making updates to containment (which is controlled by the `contain` property) - and container queries still use that containment under the hood… But there's no longer a direct need for authors to set those values directly. They'll use the `container-type` & `container-name` (or the shorthand `container`) properties — and browsers will apply containment in the background.

PR #6029, opened 2021-09-22
Minor updates to WebGPU

### LINKED ISSUE #6019: Chrome have added WebGPU support for Linux on Dev Version
chrome version 95.0.4621.4

atjn:
Just keeping the table up to date:
- the spec is now a working draft
- WebKit has added an ID for the feature
- there is a new test site that could be interesting to check out

The only somewhat major news is that Chrome now also supports WebGPU behind a flag on Linux (confirmed with tests, and [on the official wiki](https://github.com/gpuweb/gpuweb/wiki/Implementation-Status)), and that it has started an origin trial for testing WebGPU, which is pretty sweet :)

Closes #6019


Fyrd:
Thanks!

PR #6037, opened 2021-09-29
Updates to WebXR

atjn:
Some minor updates to WebXR :)

- Opera mobile supports WebXR (expected - it's built on Chromium)
- Safari for MacOS has added an experimental feature flag for WebXR (🥳). It doesn't seem to work, but it's there.


Fyrd:
Thanks!

PR #6039, opened 2021-09-29
Safari on macOS has full support for WebP

jdatapple:
Safari on macOS has fully implemented WebP. It does require macOS Big Sur or later (2020 or later), so the note is helpful. But the "partial support" olive green implies that WebP won't fully work in Safari, when that is not true.

### DIFF
--- a/features-json/webp.json
+++ b/features-json/webp.json
@@ -276,10 +276,10 @@
       "12.1":"n",
       "13":"n",
       "13.1":"n",
-      "14":"a #3",
-      "14.1":"a #3",
-      "15":"a #3",
-      "TP":"a #3"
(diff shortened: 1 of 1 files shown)


atjn:
When something is "fully supported" on caniuse, it means you can expect it to always work in that particular browser. In this case, WebP support won't always work in Safari because it depends on which version of MacOS it runs on.

This is standard practice on caniuse. If you disagree with that standard practice and want to start a discussion, feel free to open an issue about it :)

Fyrd:
Yes, @atjn's statement is correct. Developers will be quick to point out that it doesn't work for them or for particular users due to them being on Catalina, therefore we don't want to consider it fully supported.

PR #6071, opened 2021-10-30
Update css-scroll-behavior.json

brettdonald:
Safari on iOS has the same support as macOS ... smooth scrolling is considered an experimental feature, and can be turned on in Settings > Safari > Advanced > Experimental Features > CSSOM View Smooth Scrolling.


Schweinepriester (approved):

Fyrd:
Thanks!

PR #6074, opened 2021-11-02
Update multicolumn.json

mattbrundage:
The keywords `avoid-page` and `avoid-column` are now supported for the `break-inside` property: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/92

These notes are inching closer to the truth.


Fyrd:
Thanks!

PR #6076, opened 2021-11-04
Updates to `scroll-behavior`

Schweinepriester:
Marking as ready for review now, since I want to come to an end.
Links for convenience and notes:
- https://bugs.chromium.org/p/chromium/issues/detail?id=434497
- https://www.w3.org/TR/cssom-view-1/#smooth-scrolling
- https://blog.hospodarets.com/native_smooth_scrolling is 404 thus https://hospodarets.com/native_smooth_scrolling (archive shows it's the same one: https://web.archive.org/web/20161018023805/https://blog.hospodarets.com/native_smooth_scrolling)
- Checked iOS 14.8.1 on my device, where the "CSSOM View Smooth Scrolling" flag is available also. Since Safari on iOS 14.5 released together with Safari 14.1 it is pretty safe to assume it's available since then. Also it's one entry anyway 0:-)

---

I've gone with only the changes I think can be applied without doubt but there are a lot of little paths which could be taken additionally:

1. In [the most recent spec](https://drafts.csswg.org/cssom-view/) `scroll-behavior` has been moved [since 2020-10-19](https://drafts.csswg.org/cssom-view/#changes-from-2020-10-19) to CSS Overflow Module Level 3 (<https://drafts.csswg.org/css-overflow-3/#smooth-scrolling>). But I kept <https://www.w3.org/TR/cssom-view-1/#smooth-scrolling> for now, because that will probably stay that way for longer and also directly is CSSOM.
2. https://caniuse.com/scrollintoview could be a relative link, so `/scrollintoview`? Not totally sure if that works in that context? In the description of e.g. <https://caniuse.com/avif> it works as relative links should, but since I don't know how links are parsed in notes… Would it work? Want me to change it?
3. The table of <https://caniuse.com/scrollintoview> seems to directly contradict the first half of note 1
   > Partial support refers to support everything except of `Element.scrollIntoView()`

   and I'm somewhat confident that that only applied to before it was shipped enabled by default, so for us Chrome <=60 and Opera <=47, which would mean splitting that note. But since I'm not (easily) able to test those old versions in addition to not firmly being sure what to test (just `scrollIntoView({behavior: 'smooth'})` like <https://tests.caniuse.com/scrollintoview>?) I left it…
   - <https://tests.caniuse.com/css-scroll-behavior> is green
   - <https://wpt.fyi/results/css?label=master&label=experimental&aligned&q=scroll-behavior> pretty green as well…


Fyrd:
Will merge this as-is for now, as for your notes:
1. Interesting, yeah I'm fine with leaving it
2. Relative link should work fine, though sometimes I worry about how well that'll work on a 3rd party reading the data. But I've yet to hear a complaint about it so not too worried. :)
3. Hmm, yeah we may want to revisit that one. For one thing it's weird to refer to partial support when those versions aren't marked as partial!

PR #6077, opened 2021-11-04
Adding note about intersection observer on particular iOS devices

jakearchibald:
https://twitter.com/tbroyer/status/1456297026031865864.

I'll see if I can get more specific info.


jakearchibald:
Seems like this isn't confirmed https://twitter.com/smfr/status/1456298972985708545

jakearchibald:
I'll reopen if it turns out to be an actual thing.

PR #6082, opened 2021-11-09
Update Firefox on when it followed Referrer-Policy spec change

gacelperfinian:
The default policy has been changed from `no-referrer-when-downgrade` to `strict-origin-when-cross-origin`, as stated in WHATWG, and Firefox had implemented it in version 87 as stated in [BMO](https://bugzilla.mozilla.org/show_bug.cgi?id=1589074) and of course [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#browser_compatibility).


Fyrd:
Thanks!

PR #6084, opened 2021-11-12
Custom scrollbars are no longer supported in iOS Safari since iOS 14

Ruyven:
Source: https://developer.apple.com/forums/thread/670065

See also Codepen:
https://codepen.io/ruyven/pen/jOLvbQy

The second column has a custom scroll bar, but doesn't use the custom scroll bar on iOS 14 and 15. (Originally made to demonstrate a bug in Safari where scrollable divs always have a dark scrollbar, regardless of background color.)


Fyrd:
Thanks!

PR #6089, opened 2021-11-19
sql-storage: Not supported in third party contexts in chrome 97

HolgerJeromin:
ref https://blog.chromium.org/2021/11/chrome-97-webtransport-new-array-static.html


Fyrd:
Thanks!

PR #6105, opened 2021-12-12
Fix css-logical-props data for Firefox 41–65

chris-morgan:
MDN compatibility data says the last parts came in Firefox 66, and https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/66 supports this.

(These parts weren’t in the CSS Logical Properties and Values Level 1 spec when Firefox shipped the first parts.)

While looking at this file, it seems to me that notes 3 and 4 should be merged.


Fyrd:
Yup, thanks for finding this. I'll go ahead and clean up the notes.

PR #6112, opened 2021-12-15
`background-clip: text` in Safari 14+

### LINKED ISSUE #6106: background-clip: text is supported unprefixed in Safari 14 and later
Related issue on MDN compat data: https://github.com/mdn/browser-compat-data/issues/13977

I tested in Safari 14.1 and Safari 15, and this value is supported unprefixed. Confirmed in Safari 13 that it was not supported.

Schweinepriester:
Fixes #6106.

https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip#browser_compatibility looks mixed, but https://wpt.fyi/results/css?label=master&label=experimental&aligned&q=background-clip seems fine, so…


Fyrd:
Thanks!

Issue #6110, opened 2021-12-15
Add Wide Color Gamut in 2D Canvas


Schweinepriester:
- Spec ([Thanks!](https://twitter.com/svgeesus/status/1471151671275409414)):
  - https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-color-space
  - https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace
- Safari: Shipped in 15.1 <https://webkit.org/blog/12058/wide-gamut-2d-graphics-using-html-canvas/>
- Chrome: Shipped in 94 <https://chromestatus.com/feature/5807007661555712>
- Firefox: ?
- Simple test could be stolen from <https://webkit.org/blog/12058/wide-gamut-2d-graphics-using-html-canvas/#:~:text=Feature%20detection,support%20is%20available.>, the function `canvasSupportsDisplayP3`

Issue #6107, opened 2021-12-14
Add Safari 15.2 and Safari on iOS 15.2


Schweinepriester:
<https://webkit.org/blog/12140/new-webkit-features-in-safari-15-2/>

While the new features don't seem to be that many, it's still different enough from 15.1 to warrant it's own entry, right?

~~We also could wait for~~

> ~~For more on what's in Safari 15.2, including bug fixes, see the [Safari 15.2 Release Notes (coming soon.)](https://developer.apple.com/documentation/safari-release-notes/)~~

~~and decide then.~~

<https://developer.apple.com/documentation/safari-release-notes/safari-15_2-release-notes> is here, definitely it's own version, among others:

> Added support for the File System Access API for creating and iterating files in an origin-specific storage space.
> Added wide gamut support for canvas.
> Added support for COOP and COEP HTTP headers.
> Re-enabled `SharedArrayBuffer` for sites that use COOP and COEP securely to create an isolated environment.

---

Side note:
No changelog for 15.1 though oO :D
https://developer.apple.com/documentation/safari-release-notes/

Fyrd:
Done!

PR #6115, opened 2021-12-16
Safari 15.2

Schweinepriester:
Following #6107 here are the changes B-)

<https://developer.apple.com/documentation/safari-release-notes/safari-15_2-release-notes>

### "Added support for the File System Access API for creating and iterating files in an origin-specific storage space."

The situation appears to be confusing and frankly kind of a mess
- https://twitter.com/tomayac/status/1471255239894392836 (the thread in general)
- https://github.com/WICG/file-system-access/issues/342#issuecomment-995699770
- https://github.com/WICG/file-handling/issues/70

I think we either could represent Safari as having no support or partial (not outside the sandbox/origin it appears), but since we had partial before I would think of partial here as hint for developers to check out whats possible…

### Re-enabled SharedArrayBuffer for sites that use COOP and COEP securely to create an isolated environment.

✔️ 

### Added wide gamut support for canvas.

=> #6110

### Added support for COOP and COEP HTTP headers.

Via MDN:
- https://caniuse.com/mdn-http_headers_cross-origin-embedder-policy
- https://caniuse.com/mdn-http_headers_cross-origin-opener-policy

~~There doesn't appear to be an issue/PR just yet, maybe I'll do it…~~ Has been adjusted in https://github.com/mdn/browser-compat-data/pull/14006, and is in main (e.g. [COEP](https://github.com/mdn/browser-compat-data/blob/main/http/headers/cross-origin-embedder-policy.json#L33)), but not live yet it seems: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy#browser_compatibility


Fyrd:
Looks good, thanks for the research!

PR #6118, opened 2021-12-17
`overscroll-behavior`: Align Safari

Schweinepriester:
I was wondering why it looked like this, only 14.1 having the flag:

image

Turns out, you added it (only) for 14.1:

https://github.com/Fyrd/caniuse/commit/499c067dc51b8141840847cc52ef365d92c05a73#diff-a91379d552c2c3ae6a5770edfd4b4381471c66cebf485998d0cb594ad28b0d45

Since it wasn't on TP, it didn't make it's way everywhere 0:-)

Checked Safari 15.1 and Safari on iOS 15.1, the flag "CSS Overscroll Behavior" is still there, therefore it most likely was available since and still is in TP (nothing in the [release notes](https://developer.apple.com/safari/technology-preview/release-notes/)).

With the flag enabled <https://tests.caniuse.com/css-overscroll-behavior> is green:

image


Fyrd:
Good catch, thanks!

PR #6123, opened 2021-12-21
Update specification links for TTF/OTF

claudiofreitas:
- Fix TTF specification broken link
- Includes OTF specification link


Fyrd:
Thanks!

Issue #5969, opened 2021-07-13
Add entry for WAAPI


tkiat:
caniuse recognizes the search term "Web Animations API" but not WAAPI. It seems the acronym is commonly used enough that I suggest adding it.

Fyrd:
Done: https://caniuse.com/?search=waapi

PR #6127, opened 2021-12-23
Add WAAPI as keyword to Web Animations API

Schweinepriester:
There was an issue before: https://github.com/Fyrd/caniuse/issues/5969

But currently https://caniuse.com/?search=waapi leads to nothing:

image


Fyrd:
Weird, thanks!

PR #6130, opened 2021-12-28
Cascade Layers: Shipping in Firefox 97 and Chrome 99

Schweinepriester:
Triggered by <https://www.bram.us/2021/12/27/css-in-2022/#the-hotlist--cascade-layers>

> Cascade layers will be supported in Chromium 99 and Firefox 97. They're also supported in Safari TP 133 *(behind a flag)*, and I expect them to ship in the first quarter of 2022.

I checked:

- https://bugs.chromium.org/p/chromium/issues/detail?id=1095765#c55
   > Ship CSS Cascade Layers to M99
- https://bugzilla.mozilla.org/show_bug.cgi?id=1699215#c6
   > Target Milestone: --- → 97 Branch

While <https://tests.caniuse.com/css-cascade-layers> is red in all cases (maybe the test is not totally right?) the codepen https://codepen.io/miriamsuzanne/pen/poweapY is green:

image

as well as

image


Fyrd:
Thanks! Have updated the test, is green now.

PR #6135, opened 2022-01-05
Full support for `text-emphasis` in Chrome 99

Schweinepriester:
https://bugs.chromium.org/p/chromium/issues/detail?id=666433#c38

> **Status:** Fixed (was: Available)\
> **Labels:** Target-99

Looking at <https://tests.caniuse.com/text-emphasis> in Chrome 96 already suggests it knows `left` prefixed as opposed to note 1:

image

~~Going for draft however for now since my Chrome Canary 99 keeps crashing and https://wpt.fyi/results/css?label=master&label=experimental&aligned&q=text-emphasis isn't at 99 yet - I would expect 14 / 148 to go way up or rather the test [here](https://wpt.fyi/results/css/css-text-decor?label=master&label=experimental&aligned&q=text-emphasis) being way more green (since they link to the bug above).~~

Looking good:

image

The tests from WPT seem good in 99 as well, at least a few I checked.


Fyrd:
Thanks!

PR #6144, opened 2022-01-14
input-inputmode: Adding result from own iPhone tests

HolgerJeromin:


jensimmons:
Did you create a ticket for this at [https://bugs.webkit.org](bugs.webkit.org)? That would be helpful.

HolgerJeromin:
> Did you create a ticket for this at [bugs.webkit.org](https://bugs.webkit.org/)? That would be helpful.

No, because as far as I understood Safari is kind of [spec](https://html.spec.whatwg.org/multipage/interaction.html#attr-inputmode-keyword-decimal) conform. So I opened:
https://github.com/whatwg/html/issues/7497

jensimmons:
Since Safari conforms to the current spec, a note about a bug is not warranted. 

@HolgerJeromin You are welcome to open an issue at bugs.webkit.org anyway. Or rather [feedbackassistant.apple.com](https://http://feedbackassistant.apple.com) is better, since this is an OS-level thing.

HolgerJeromin:
> Since Safari conforms to the current spec, a note about a bug is not warranted.

IMO a note on caniuse is useful if there is behavior which is different in some browsers to warn developers. 

I do not expect any useful reaction on a safari/webkit issue tracker. 
IMO first the spec should fully define what we expect.

jensimmons:
This issue is asking for a note to be added about Safari support, making it seem like it has a bug because it follows the spec, and the OP believes the spec should be different than it is? Can we please close this issue.

HolgerJeromin:
> This issue is asking for a note to be added about Safari support, making it seem like it has a bug because it follows the spec, and the OP believes the spec should be different than it is? 

IMO this condense to the question: 
What warrants a "note" in the caniuse data? Only bugs or also other hints regarding the feature?

The repo has this documentation:
https://github.com/Fyrd/caniuse/blob/main/CONTRIBUTING.md#supported-changes
> notes — Notes on feature support, often to explain what partial support refers to

But the maintainer of this repo may do what ever they like. 
I wanted to warn other devs that the feature could be useless for them (as it is for us).

PR #6149, opened 2022-01-22
Mark Opera 78+ as fully supporting position:sticky

xparadoxical:
Works for me on Opera 82.0.4227.50, so I checked other Opera versions and looks like the Chrome bug was fixed in Opera 78.


Fyrd:
Confirmed, thanks!

PR #6152, opened 2022-01-23
Remove redirecting link

estelle:
Removed word-wrap link as it redirects to the overflow-wrap page, which is one of the remaining links.
    {
      "url":"https://developer.mozilla.org/En/CSS/Word-wrap",
      "title":"MDN Web Docs - CSS word-wrap"
    },


Fyrd:
Thanks!

PR #6158, opened 2022-01-31
chore: update Firefox footnotes for COLRv1 format

yisibl:


features-json/colr-v1.json
```diff
@@ -62,7 +62,8 @@
       "94":"n",
       "95":"n",
       "96":"n",
-      "97":"n"
+      "97":"n",
+      "98":"n"
```
HM100, line 66: Needs to be removed in order for the checks to be passed. New browser versions can't be added via PRs

HM100 (changes requested):

Fyrd:
Thanks! Have added Edge 98.

yisibl:
Thanks to all.

PR #6160, opened 2022-02-01
Added Safari 15.4 beta feature support

jdatapple:
Adds Safari 15.4 beta support for:

- BroadcastChannel
- CSS appearance property
- CSS Cascade Layers
- CSS Containment
- CSS display: contents
- CSS :focus-visible
- CSS :has()
- Dialog element
- CSS ::backdrop
- New viewport unit variants


features-json/fullscreen.json
```diff
@@ -306,7 +306,7 @@
       "15":"a x #2 #3",
       "15.1":"a x #2 #3",
       "15.2-15.3":"a x #2 #3",
-      "15.4":"a x #2 #3",
+      "15.4":"a x #3",
       "TP":"a x #2 #3"
```
Schweinepriester, line 310: Can Note 2 regarding no `::backdrop` also be removed for TP? :)

features-json/fullscreen.json
```diff
@@ -413,7 +413,7 @@
       "14.5-14.8":"a #5",
       "15.0-15.1":"n",
```
Schweinepriester, line 414: Not directly related: I noticed Full Screen API was `n` for Safari on iOS 15.x before ([initial change](https://github.com/Fyrd/caniuse/commit/725d2b36644ac1dc593021f9b00b72ecc5328c06#diff-15d3af05bdeb0fe2ab82eac9b0103fa19eeb5eff2cf49ba2b5b9ba7c0b07f001) by @Fyrd) - I compared 14 and 15 on iPad using browserstack, seemed the same for me. I can send another PR for that after this one :)

Schweinepriester (approved):

features-json/fullscreen.json
jensimmons: I would expect that the note should still be on the site because earlier versions of Safari still have that note applied. That's the pattern I've seen across Can I Use.

jensimmons: Yes, note #2 should not apply to Safari Technology Preview.

features-json/fullscreen.json
jdatapple: I'm conferring with engineers to verify the correct data here.

jdatapple: I verified that 15.0-15.4 should retain the same partial support as the 14.5-14.8 entry.

Fyrd:
Thanks!

Antonio-Laguna:
Out of curiosity, given that features can be pulled last minute from a Beta, how is that 15.4 is being reflected on caniuse / mdn data?  Perhaps @jdatapple has more insight given his work but I was just updating CSSDB and started to see it flowing into features and got a bit confused.

jdatapple:
As things change in subsequent beta seeds and at public release, I will be updating CanIUse and MDN data to match.

Antonio-Laguna:
@jdatapple thanks for the answer. It's still a bit confusing to me as to why an unreleased version is showing on Can I Use or MDN. Beta is not generally available. I could be wrong but I'm not seeing this behaviour replicated by Chrome / Firefox.

We've just recently introduced `vendors_implementations` count and I feel that it might not be realistic to keep that as it is unless I find some mechanism to check whether the version is public or not (obviously not on you).

LifeIsStrange:
This comes a decade too late.. Webkit is abandonware in comparison to chromium and whenever I get the opportunity I deliberately choose to make my webapps redirect webkit users to an installation of a chromium derivative.
Still, that's a bit of progress for the poor people stuck on those Iprisons.

jensimmons:
> It's still a bit confusing to me as to why an unreleased version is showing on Can I Use or MDN. Beta is not generally available. I could be wrong but I'm not seeing this behaviour replicated by Chrome / Firefox.

Chrome and Firefox _absolutely_ show data for pre-released versions. That's the beauty of the data visualizations on Can I Use. Developers can quickly see which browser versions are coming in the future, which versions are the current versions, and what was in the past. 

Screen Shot 2022-02-04 at 3 32 04 PM

In this screenshot Firefox 97-98, Chrome 99-101, Safari 15.4–STP all refer to future releases. The row that starts on the left with IE11 is the current browser row. Everything above that is the past.

jensimmons:
@LifeIsStrange I think you’ve misunderstood what web standards means.

Antonio-Laguna:
Thanks, Jen for clearing my confusion! I'll check this and see what can I do to convey this better on CSSDB

jensimmons:
Sometimes people get confused because the version numbering for Chrome/Edge and Firefox has become _so_ similar in recent years. 

IE, Edge, Opera, Firefox, and Chrome used to have very different numbering from each other. Now they are almost identical. And Safari's numbering did not change to match the others. 

But the process of how a browser is released is very similar. Chrome/Edge, Firefox and Safari all have preview browsers, all have betas, and all have releases. 

It's helpful to web developers to be able to see what's coming in the future. And it's helpful to both the people writing the web standards and the engineers implementing new technology to have these preview browsers for public testing.

Antonio-Laguna:
@jensimmons that's exactly what happened to me! They release with a different cycle too and given I'm an iOS user myself I'm constantly aware of what's and what's not released.

Thanks for clarifying. This evolved into correcting the implementation number on CSSDB. If anyone is interested it happened here: https://github.com/csstools/cssdb/pull/82 and was just released.

Thanks again for your help!

PR #6165, opened 2022-02-03
CSS @apply rule should be removed

jensimmons:
1. This was never an official W3C CSSWG standard. And I don't believe that an idea sketched out in a personal git repo would count for inclusion on Can I Use today.
2. Tab (the author of the spec) abandoned the idea. https://www.xanthir.com/b4o00
3. Chrome removed the experiment from their browser. https://chromium.googlesource.com/chromium/src/+/5874fca7324e4523a4bdecc8999bdadfdb6c4eff

For more: https://chromestatus.com/feature/5753701012602880

This will never be a CSS standard. It will never be implemented in all browsers.


Schweinepriester:
Never say never… 🙃

But for real: Interesting and I agree with the direction! Though if it was up to me, I would keep a search result ([example](https://caniuse.com/sr_flif)) around for historic purposes. Afaik the search result data isn't open source though, so…

Fyrd:
Thanks! I agree, and will replace with a search result for folks curious about the status.

I don't regret having including it though, it was starting to be implemented across (Chromium) browsers and [the request issue](https://github.com/Fyrd/caniuse/issues/2480) indicated a significant amount of developer interest in it. That's enough for any unofficial feature to be included so folks can keep track of support levels.

PR #6166, opened 2022-02-04
Chrome: Small/large/dynamic/logical viewport units 🎉

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=1093055#c13>:

> Support small/large/dynamic/logical viewport units
>
> This CL adds support for new units:
>
> * svw, svh (etc): Resolves against the viewport as it would be with browser controls fully expanded.
> * lvw, lvh (etc): Resolves against the viewport as it would be with browser controls fully minimized.
> * dvw, dvh (etc): Resolves against the viewport as it presently appears given the current state of browser controls.
> * vi, vb: Resolves against vw/vh depending on the writing-mode of the target element.

🎉 

WIP to wait for the next version of canary to test using https://tests.caniuse.com/viewport-unit-variants and to check if it will be with version 100.

Current state with Version 100.0.4870.0:

image

EDIT: Still the same with Version 100.0.4877.0…

EDIT2: Still the same with Version 100.0.4896.0 - I'll wait for 101, if it doesn't appear then I'll close the PR…


Schweinepriester:
Still the same in canary Version 101.0.4923.1, so closing now.

But I'm sure it will come later this year as part of [Interop 2022](https://wpt.fyi/interop-2022) 0:-)

PR #6171, opened 2022-02-08
Update Safari 15.4 to partial support for lazy loading

jdatapple:
Lazy image loading is supported by default in Safari 15.4 and Safari Technology Preview. Lazy iframe loading is available via an off-by-default Experimental Features option.


Fyrd:
Thanks!

PR #6172, opened 2022-02-10
Update Picture element support version of android

lynkas:
source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#browser_compatibility


features-json/picture.json
```diff
@@ -429,6 +429,7 @@
       "4.2-4.3":"n",
       "4.4":"n",
       "4.4.3-4.4.4":"n",
+      "38":"y",
```
HM100, line 433: Adding new browser versions is not one of the things you can do using PRs

PR #6177, opened 2022-02-11
Add text-align-last prefixed support data for Safari

jdatapple:
Add support for prefixed `-webkit-text-align-last` since Safari 8 on macOS and Safari on iOS 8.

See [[CSS3] Add rendering support for -webkit-text-align-last](https://github.com/WebKit/WebKit/commit/eadea592557c7b9f6ce7cf649d9570667d9742d5)


jdatapple:
I verified with engineering. This is, in fact, behind a flag that is not enabled.

PR #6178, opened 2022-02-11
Fix support data for browsers missing flex-basis

jdatapple:
Support for the `flex-basis` property is an important part of the Intrinsic & Extrinsic Sizing in the CSS Box Sizing spec. 

There's currently some inconsistency in the data where some browsers missing `flex-basis` show "partial support", while others are showing full support. This PR fixes the inconsistent data to mark browsers missing `flex-basis` support to consistently show "partial support". Additionally, this adds STP support for `flex-basis` (added in [STP 140](https://webkit.org/blog/12255/release-notes-for-safari-technology-preview-140/)).


jensimmons:
Ping?

Fyrd:
Still needs more revisiting as mentioned in my previous comment, but I just confirmed support in Safari 16+ and have updated there appropriately.

jensimmons:
Thank you!

PR #6202, opened 2022-03-11
[meter element] Interoperability issues

da2x:
* Added links to segment boundary interoperability bug.
* Purged dead links.


Fyrd:
Thanks!

PR #6204, opened 2022-03-13
Chrome 101: CSS `:has()` behind flag

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=669058#c47>:

> Enable both snapshot and live :has() pseudo class support behind
the experiemental web platform feature flag.

Though

> There are some functional limitations.
> - :has() doesn't work properly with pseudo elements in the
  subselector.
> - :has() doesn't work properly with some logical combinations
  (:is(), :where() and :has()) in the subselector.

which seems to show on <https://wpt.fyi/results/css/selectors?label=master&label=experimental&aligned&q=has>:

image

But since it's behind a flag away, literally saying experimental, I think it's not yet anything to denote on caniuse.

For the record, <https://tests.caniuse.com/css-has>:

image

### DIFF
--- a/features-json/css-has.json
+++ b/features-json/css-has.json
@@ -265,8 +265,8 @@
       "98":"n",
       "99":"n",
       "100":"n",
-      "101":"n",
-      "102":"n"
+      "101":"n d #1",
+      "102":"n d #1"
     },
     "safari":{
       "3.1":"n",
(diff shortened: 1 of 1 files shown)


LifeIsStrange:
Finally, it's coming !!

Fyrd:
Nice, thanks!

PR #6206, opened 2022-03-14
Correct display:contents status for Safari

jensimmons:
The note that Safari 15.4 fixes the a11y bug with display: contents is premature.


Fyrd:
Thanks!

PR #6212, opened 2022-03-21
Added more details about Safari support

PitWenkin:
Added a second note about Safari, which more clearly explains what is and what is not supported. It's the same link as the one in the MDN browser compatibility  table.


Fyrd:
Thanks!

PR #6218, opened 2022-03-27
fix property list

estelle:
tab size is relevant for 3 values of white space, but only two were listed. 
added `white-space: break-spaces` to the list.

example: https://codepen.io/estelle/pen/WNdpPPo


Fyrd:
Thanks!

PR #6226, opened 2022-04-07
Safari 15.4 shipped full support for mask

jensimmons:
From the official release notes:
>Added the mask property, along with the long-hand forms mask-image, mask-size, mask-repeat-x, mask-repeat-y, mask-origin, and aliased -webkit-mask.

https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes


Fyrd:
Thanks!

Issue #6121, opened 2021-12-20
Add the `<model>` element


Schweinepriester:
<https://github.com/WebKit/explainers/tree/main/model>

Since it's being added to Safari…
<https://webkit.org/blog/12156/release-notes-for-safari-technology-preview-137/>:
> ### Experimental Model Element
>
> -   Added support for mouse-based manipulation of `<model>` on macOS ([r285986](https://trac.webkit.org/changeset/285986/webkit/))
> -   Added audio support ([r286065](https://trac.webkit.org/changeset/286065/webkit/))
> -   Added support for controlling looping animations ([r286066](https://trac.webkit.org/changeset/286066/webkit/))
> -   Added support for getting and setting the camera ([r286019](https://trac.webkit.org/changeset/286019/webkit/))
> -   Added support for pausing and resuming animations ([r286048](https://trac.webkit.org/changeset/286048/webkit/))
> -   Added support for seeking animations ([r286068](https://trac.webkit.org/changeset/286068/webkit/))
> 
> ### Accessibility
> 
> -   Added accessibility attributes for `<model>` ([r286406](https://trac.webkit.org/changeset/286406/webkit/))

Malvoz:
The new Github repo for this element proposal is https://github.com/immersive-web/model-element.
And draft spec: https://immersive-web.github.io/model-element/

jensimmons:
For clarity, `<model>` is currently in Safari Technology Preview, behind a flag.

jensimmons:
I'd like to very much endorse the addition of a page for the `<model>` element. 

Similar to the `img`, `audio`, and `video` elements, the `model` element provides a way to embed 3D models into a web page. 

This code example demonstrates how a `.usdz` file can be shown on a webpage, with video files as fallbacks. The `.usdz` file is a format for 3D geometric data.

```
<model style="width: 400px; height: 300px">
    <source src="assets/example.usdz" type="model/vnd.usdz+zip">
    <picture>
        <source src="animated-version.mp4" type="video/mp4">
        <source src="animated-version.webp" type="image/webp">
        <img src="animated-version.gif"/>
    </picture>
</model>
```

jensimmons:
@Schweinepriester I'd appreciate it if you could take a look at my attempt to create a PR, and fix it. https://github.com/Fyrd/caniuse/pull/6228

Schweinepriester:
https://developer.apple.com/documentation/safari-release-notes/safari-17-release-notes

[...]

PR #6228, opened 2022-04-07
Adding <model> element to Can I Use

jensimmons:
As suggested in #6121, I'm making an attempt to create a page for `<model>`.

There's no `HTML` category? Only `HTML5`? This is not HTML5, it's much later than version 5. No `media` category? That's where I would have put this, along with HTML. I'm guessing this might throw an error, but I'll try anyway.


features-json/model
```diff
@@ -0,0 +1,480 @@
+{
```
Malvoz, line 1: This should be a `.json` file.

Malvoz:
Isn't `<model>` highly experimental still? Unresolved accessibility issues last I checked, no spec, though there's the explainer. 

Should there be a note explicitly stating that it is experimental, and that authors can expect changes to it?

jensimmons:
Ah, right. `.json`. Thanks.

And yes, perhaps it is too early to create a page for `<model>`. The standards process is underway, but perhaps this page should be added once things are more mature.

PR #6230, opened 2022-04-08
Safari TP 143: Permissions API

Schweinepriester:
https://webkit.org/blog/12563/release-notes-for-safari-technology-preview-143/

> Enabled the Permissions API by default ([r291116](https://trac.webkit.org/changeset/291116/webkit/), [r290301](https://trac.webkit.org/changeset/290301/webkit/))

<https://tests.caniuse.com/permissions-api>:

image

Interacted with <https://googlechrome.github.io/samples/permissions/index.html> as well, seemed to behave the same as Chrome, except for midi, but that's of course expected, given the [(non) support](https://caniuse.com/midi).


Fyrd:
Thanks!

PR #6238, opened 2022-04-12
Firefox on Android 99: Network Information API disabled

Schweinepriester:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/99

> The [Network Information API](https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API) was previously enabled on Android (only), but is now disabled by default on all platforms. This API is on the path for removal because it exposes a significant amount of user information that might be used for fingerprinting. ([bug 1720353](https://bugzilla.mozilla.org/show_bug.cgi?id=1720353)).


Fyrd:
Thanks!

PR #6240, opened 2022-04-15
Update rrggbbaa spec

lens0021:


saschanaz:
Do we really need to have this when there is the corresponding MDN item? https://caniuse.com/mdn-css_types_color_alpha_hexadecimal_notation

features-json/css-rrggbbaa.json
```diff
@@ -1,8 +1,8 @@
 {
   "title":"#rrggbbaa hex color notation",
   "description":"The CSS Color Module Level 4 defines new 4 & 8 character hex notation for color to include the opacity level.",
-  "spec":"https://drafts.csswg.org/css-color/#hex-notation",
-  "status":"unoff",
+  "spec":"https://www.w3.org/TR/css-color-4/#hex-color",
```
saschanaz, line 4: Anyway, I think you can keep it as drafts.csswg.org.

lens0021:
> Do we really need to have this when there is the corresponding MDN item? https://caniuse.com/mdn-css_types_color_alpha_hexadecimal_notation

Apologies, I was confused when I was looking for features, and I'm still confused. CONTRIBUTING.md does not seem to describe mdn and I don't know the difference between drafts.csswg.org and W3C working draft. Do you think I should close this PR?

saschanaz:
I mean, should we consider this caniuse-specific item as duplicated? I see no point to have two items for the same feature.

Fyrd:
I agree with keeping it as the drafts spec but updating it to `wd`. Thanks!

As for the duplicate status, I have marked the MDN entry as a duplicate of this one which means next time the MDN data is rebuilt it will no longer appear (since the caniuse entry can include more info)

Fyrd:
Updated manually so closing this PR.

PR #6260, opened 2022-05-04
Update Firefox support for Streams

Schweinepriester:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/100#apis

> [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream), [`WritableStreamDefaultWriter`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStreamDefaultWriter), [`WritableStreamDefaultController`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStreamDefaultController), and [`ReadableStream.pipeTo()`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeTo) are now supported ([bug 1759597](https://bugzilla.mozilla.org/show_bug.cgi?id=1759597)).

So note 5 is gone with Firefox 100, but looking at <https://tests.caniuse.com/streams>

image

I noticed BYOB is missing and MDN confirms it was never there:
- https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamBYOBReader#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamBYOBRequest#browser_compatibility

=> Note 4 for all versions.

While `pipeTo` is available it's [marked as preview](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeTo#browser_compatibility) and `pipeThrough` [isn't available by default](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeThrough#browser_compatibility) so I left note 6 alone for now.


Fyrd:
Thanks!

PR #6277, opened 2022-05-24
Safari Technology Preview has 100% support of WebP

jensimmons:
Safari 14.0 implemented WebP completely and correctly. The only reason causing Safari 14–15 to be labeled "partial support" is the requirement for macOS 11 Big Sur or later. 

Safari Technology Preview now also requires macOS Big Sur or later. So it's impossible to run Safari Technology Preview on any version of macOS and not automatically have 100% full support of WebP.


Schweinepriester:
In theory I agree with this change and would have had it on my radar, but I would have opened the PR only once the next major version of Safari (16?) is public and confirmed to not run on macOS 10.15 Catalina.
Looking at the history (https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility) this seems very plausible, esp. with Safari TP already only being available starting with Big Sur, but since macOS 10.15 Catalina is the last version of 10.x I thought maybe there will be a deviation.

I assume you can't comment on the probable next version and also I'm basically only writing this to let you know why I didn't (attempt to) change this earlier :)

LifeIsStrange:
@jensimmons Time for JPEG XL now, chomium and firefox already have merged support of libjxl

Fyrd:
I think @Schweinepriester makes a good point but since it's technically true for TP I'll allow it. :)

PR #5415, opened 2020-05-05
Safari fixed input events in 12.1 (radio/checkbox) and 13.1 (file)

mkurz:
## Starting with Safari 12.1...
...the radio inputs and checkbox inputs fire “click”, “input”, and “change” events in order when clicked:
* https://bugs.webkit.org/show_bug.cgi?id=190223
* https://trac.webkit.org/changeset/236779/webkit

Read https://webkit.org/blog/8419/release-notes-for-safari-technology-preview-67/
> "Fixed radio inputs and checkbox inputs to fire “click”, “input”, and “change” events in order when clicked"

Preview 67 was mentioned in the 12.1 blog post here (see bottom): https://webkit.org/blog/8718/new-webkit-features-in-safari-12-1/

## Starting with Safari 13.1...
...finally the file input fires an input event before the change event:
* https://bugs.webkit.org/show_bug.cgi?id=204292
* https://trac.webkit.org/changeset/252768/webkit

Read https://webkit.org/blog/9672/release-notes-for-safari-technology-preview-97/
> "Changed the file input to fire an input event before the change event"

Preview 97 was mentioned in the 13.1 blog post here (see bottom): https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/

PR #6282, opened 2022-05-26
Input event: Match Safari on iOS to Safari

Schweinepriester:
Basically a follow-up to #5415.

Checked https://jsbin.com/jojoji/edit?html,output (mentioned in one of the bugs, https://bugs.webkit.org/show_bug.cgi?id=149398) on iOS 15.4, seems to work fine:

086BD79E-494B-41CA-A131-E2E274F9C659

Because the version numbers don't match exactly or rather are ranges there's some inaccuracy, but at least in the under promising way - on desktop 12.1 is improved with note 6 already, while 13.1 is fully fixed.

---

Additionally I think to most of the other mobile browsers note 4 probably isn't applying anymore, but haven't checked yet.


Fyrd:
Looks good, thanks!

PR #6284, opened 2022-06-01
Complex :nth-child selectors are Working Draft

jensimmons:
Complex :nth-child  and :nth-last-child selectors are part of Selectors 4, which is in Working Draft. This is not an unofficial specification. 
https://www.w3.org/TR/selectors-4/#the-nth-child-pseudo


Fyrd:
Thanks!

PR #6288, opened 2022-06-07
Safari 16: Add note about Push only being available in macOS Ventura

daytonlowell:
And changed iOS to `n` as it's not coming until 2023.

https://webkit.org/blog/12945/meet-web-push/


features-json/push-api.json
```diff
@@ -298,8 +298,8 @@
       "15.2-15.3":"n #3",
       "15.4":"n #3",
       "15.5":"n #3",
-      "16.0":"y",
-      "TP":"n d #5"
+      "16.0":"y #6",
+      "TP":"n d #5 #6"
```
Schweinepriester, line 302:
```suggestion
      "TP":"y #6"
```

Not sure if the corresponding version of TP is already out, but I believe we can be very sure it will match Safari 16 capabilities, so I would have changed that already 0:-)

Schweinepriester (commented):
You beat me to it, now you have to endure my comments 🙃

But for real, thank you! :)

daytonlowell:
Good points!

Schweinepriester (approved):

Fyrd:
Thanks!

PR #6291, opened 2022-06-07
Shared Web Workers for Safari TP as well

Schweinepriester:
Given you rightfully added support for Safari 16 I went and checked TP:
Looks like we missed it:
https://webkit.org/blog/12522/release-notes-for-safari-technology-preview-142/
> Added support for sharing Shared Workers (including across WebProcesses) ([r289247](https://trac.webkit.org/changeset/289247/webkit/))


Fyrd:
Thanks!

PR #6294, opened 2022-06-08
Update WebP to show Safari 16 as fully supported

jensimmons:
WebP is fully supported on macOS Big Sur and later. Safari 16 can only run on Big Sur and later. So therefore, all installs of Safari 16 have full support of WebP.


Schweinepriester (approved):

AndyHowarthSbuffles (changes requested):
`[jensimmons:patch-15]() 
Duplicate of #`

AndyHowarthSbuffles:
> [jensimmons:patch-15]() Duplicate of #

Fyrd:
Thanks!

PR #6295, opened 2022-06-08
Small adjustment to note 3 regarding Safari

jensimmons:
A small, clarifying adjustment to the note about WebP in Safari 14.0–15.6.


Schweinepriester:
LGTM, but shouldn't 15.6 be released first and added to caniuse? ]:->
No but seriously, good to go :)  

For Fyrd:
Looking at the [release notes](https://developer.apple.com/documentation/safari-release-notes/safari-15_6-release-notes) I think 15.5 and 15.6 can be one entry 0:-)

PR #6302, opened 2022-06-08
Safari 16 on macOS fully supports WebM

jensimmons:
Similar to WebP, Safari 16 requires macOS Big Sur or later, so it's impossible for it to run on a version of macOS that doesn't have WebM support. I've also update this note so it's very clear to web developers when it works or not.

### DIFF
--- a/features-json/webm.json
+++ b/features-json/webm.json
@@ -298,8 +298,8 @@
       "15.2-15.3":"a #7",
       "15.4":"a #7",
       "15.5":"a #7",
-      "16.0":"a #7",
-      "TP":"a #7"
+      "16.0":"y",
+      "TP":"y"
     },
     "opera":{
       "9":"n",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
But isn't this case special? Because per the note it requires a mid-cycle version namely macOS **11.3**, i.e. somebody running macOS 11.0 - 11.2 could get Safari 16 without support for WebM?

Or is this version range, 11.0 - 11.2, not able to get Safari 16? Then it would be fine.

jensimmons:
I'll find out.

jensimmons:
I just confirmed that upon launch Safari 16 will require macOS 11.6 (Big Sur), macOS 12.6 (Monterey), or macOS 13.x (Ventura).

Fyrd:
Cool, sounds like the changes are good to go as-is then. Thanks both!

Issue #6300, opened 2022-06-08
"HTML5 form features" is a weird page, out of date


jensimmons:
Hm. https://caniuse.com/forms

I'm not sure what to recommend to do about this page. It claims Firefox, Safari, Baidu and KaiOS have only partial support for "HTML5 forms", but there's no information about what is incomplete. There's nothing actionable about this. 

I don't believe either Safari or Firefox have incomplete form implementations. But also have no idea when whatever was missing when this page was created shipped. 

Should this page exist anymore? It feels old and abandoned. But also, it still counts towards the overall browser score on the homepage of caniuse.com. And it makes multiple browsers look ancient and out of touch.

Fyrd:
Yeah, this one's a holdover of a long time ago when [Opera announced a whole suite of new form controls](https://dev.opera.com/articles/new-form-features-in-html5/). Because they were the first browser to release them all in one go they become a single feature on caniuse. 

Since then I'd been updating it based on whether or not all of those sub-features were supported (see the Sub-features tab for the list). I think because the "month" and "week" input types aren't supported in Firefox & Safari the parent features are also only considered partially supported.

You're right though that such a page wouldn't be added today and I agree that it would be fine to remove it.

Fyrd:
The feature no longer appears, so closing this ticket.

PR #6312, opened 2022-06-09
Removing the HTML5 form features page

jensimmons:
As discussed in #6300. Because it doesn't match the current way pages are done. Too broad. Specific form types/features are covered in other existing pages.


Fyrd:
Instead of removing it I've just set "shown" to "false" for now. Also needed to remove the reference to this feature in all its sub-features. So closing this ticket as it's effectively done. :)

Issue #6298, opened 2022-06-08
A new page for Grid track animation should be added


jensimmons:
There's a page for CSS Grid (level 2) and a page for Subgrid (Grid level 2). But something that browsers did not universally ship with Grid 1 in 2017 is the ability to animate grid track (`grid-template-columns` and `grid-template-rows`). 

Firefox added this ability sometime in early 2019. (It landed in Firefox Nightly in Jan 2019). Safari is adding this ability in Safari 16, and already added it to STP. 

There is not anyplace to look up support for animating Grid tracks. MDN lists support for the properties, but only has a small note that these are CSS properties that should be animatable. Can I Use is the perfect place to store this data, especially now that broader support will bring more attention to the idea of making content move by using grids + animation.

PR #6314, opened 2022-06-09
Adding new page for CSS Grid Animation support

jensimmons:
Attempt No 2. 
As discussed in #6298.
Supported in Firefox since FF66. https://hacks.mozilla.org/2019/03/firefox-66-the-sound-of-silence/
Supported in Safari 16 and STP. https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes
Not supported in Chromium browsers (I tested using CodePen demo at https://css-irl.info/animating-css-grid/)


jensimmons:
I'm not sure what's wrong. I don't understand the error message. Probably something to do with the very end of the file. But everything regarding the content about Grids + animation, and about browser support is correct. 

Can someone who knows the Can I Use build process better fix this file?

Schweinepriester:
I'm not totally sure, but maybe it's the spacing, as in it maybe wants 2 spaces per level? Currently it seems to be tabs...

features-json/css-grid-animation.json
```diff
@@ -0,0 +1,491 @@
+{
+	"title":"CSS Grid animation",
+	"description":"As defined in CSS Grid level 1, grid-template-columns and grid-template-rows are properties that can be animated.",
+	"spec":"https://www.w3.org/TR/css-grid-1/",
+	"status":"cr",
+	"links":[
+		{
```
Fyrd, line 7: I think it's the empty "link" that it's got a problem with, just need to change this to an empty array. :)

jensimmons: Thanks. Fixed. Let's see if that passes.

jensimmons:
Grr.

Issue #6297, opened 2022-06-08
Container Queries should be broken out into three pages


jensimmons:
Right now, there's one page for "Container Queries". Most web developers will think of this as what's more specifically called "size queries". The ability to write CSS and have it applied conditionally based on the size of a container. 

There are also two other features coming: Container Queries Units, and Style Queries. 

These won't be covered by MDN pages (or at least Style Queries won't) because the CSS properties involved are the same. It's the overall feature, the overall concept and use case that's different.

So I think two more pages should be added to Can I Use, so there will be three in total:
- Container Queries
- Container Query Units
- Container Queries: Style Queries

Current browser support for CQ units is much the same as Container Queries right now (in Safari 16 + STP, in progress in Chromium behind a flag). 

Style Queries in not in Safari 16. I'm not sure if it is in progress for Chromium behind a flag, or if they haven't started yet.

Fyrd:
All three are now available, so closing this ticket.
 - https://caniuse.com/css-container-query-units
 - https://caniuse.com/css-container-queries
 - https://caniuse.com/css-container-queries-style
 
Thanks for your work on these @jensimmons !

PR #6315, opened 2022-06-09
Clarifying Container Queries page to be about Size

jensimmons:
Narrowing the scope of this page to be about Size Queries in the Container Queries spec. Updating description to reflect the maturity of the spec. As discussed in #6297.


Fyrd:
Thanks!

PR #6316, opened 2022-06-09
Create a page for Container Query Units

jensimmons:
As discussed in #6297, here's the page for Container Query Units.


css-container-query-units.json
```diff
@@ -0,0 +1,490 @@
...
+      "84":"n",
+      "85":"n",
+      "86":"n",
+      "87":"n",
+      "88":"n",
+      "89":"n",
+      "90":"n",
+      "91":"n",
+      "92":"n",
+      "93":"n d #1",
```
andruud, line 252: Note that Chrome 93-100 (inclusive) implements them as `q*`. 101 and onwards updated to `cq*`.

jensimmons:
Like #6318, Edge probably doesn't have a flag.

css-container-query-units.json
jensimmons: Thanks! I made the change.

Fyrd:
Thanks!

PR #6318, opened 2022-06-10
fix: edge has no flag for container queries

Th3S4mur41:
Looking for CSS container queries under [edge://flags](edge://flags) in all current versions (from 102 to 104 canary) results in "No matching experiments".

image

The flag seems to be currently limited to Chrome


Fyrd:
Confirmed, thanks!

PR #6317, opened 2022-06-10
Creating page for Container Queries — Style Queries

jensimmons:
As discussed in #6297.
No browsers support style queries yet. But there are about to be a lot of tutorials and such for Size Queries in Container Queries... and I expect web developers will start to hear a lot about Style queries in the process.

PR #6321, opened 2022-06-10
Creating new page for CSS Grid Animations

jensimmons:
Starting over. Let see if this works. 
As discussed in #6298.


Schweinepriester:
I think now I know what the issue is, one sec :)

jensimmons:
I have no idea why this is failing yet again. Do I have too many keywords?

features-json/css-grid-animation.json
```diff
@@ -0,0 +1,490 @@
...
+		}
+	},
+	"notes":"",
+	"notes_by_num":{
+
+	},
+	"usage_perc_y":0,
+	"usage_perc_a":0,
+	"ucprefix":false,
+	"parent":"parentfeatureid",
```
Schweinepriester, line 483:
```suggestion
	"parent":"",
```

Searching for the error I ended up at
https://github.com/Fyrd/caniuse/blob/c9309d571180baf8eac87833fa7c6b6754d3129a/validator/validate-jsons.js#L238-L242

While it's strange that the other PRs (at least #6317) didn't fail with that, it seems like the only place it could come from!?
Soooo maybe lets try my suggestion, emptying that? :D

jensimmons: Removed on the same branch.

Schweinepriester:
Hooray, seems to have worked 🎉

PR #6322, opened 2022-06-10
Mark WebP as "other"

jensimmons:
WebP should be marked as "other", like JPEG XL https://caniuse.com/jpegxl and AVIF https://caniuse.com/avif, not "unofficial".


Fyrd:
Bit tricky to determine this, but yeah with all browser vendors on board that seems reasonable enough to me :)

PR #6323, opened 2022-06-11
Changes to Push API regarding Safari

Schweinepriester:
Source: https://twitter.com/dfabu/status/1534223355225575424 (see above as well)

I've been struggling with myself if that warrants partial support, but for now I came to the conclusion yes, because in the past we gave partial even for only support starting with a specific macOS version and here's more.


Fyrd:
Agreed, thanks!

PR #6324, opened 2022-06-12
Fix for link fail I introduced in #6323 

Schweinepriester:
I had hoped this would work out as a link as it works in some markdown, [but nope](https://caniuse.com/push-api):

image

Sorry 0:-)


Fyrd:
Ha yeah I meant to manually fix that.

PR #6325, opened 2022-06-12
Variable fonts - Update notes for Firefox

### LINKED ISSUE #6292: Firefox doesn't support variable fonts on Windows 7 (https://caniuse.com/variable-fonts)
The notes for Firefox do indicate that MacOS 10.13+ is required for variable font support, but do not include caveat notes for Windows 7 which is not supported.

References:
https://bugzilla.mozilla.org/show_bug.cgi?id=1742970
https://bugzilla.mozilla.org/show_bug.cgi?id=1403299
https://bugzilla.mozilla.org/show_bug.cgi?id=1302685

Schweinepriester:
Fixes #6292.

Biggest struggle regarding the "or" between macOS and Windows - "/" didn't look good, but also I don't want to logically exclude Linux support ("[Linux OSes need the latest Linux Freetype version](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Fonts/Variable_Fonts_Guide)"), but oh well.

https://en.wikipedia.org/wiki/Windows_10#Updates_and_support

### DIFF
--- a/features-json/variable-fonts.json
+++ b/features-json/variable-fonts.json
@@ -493,12 +493,12 @@
   "notes":"",
   "notes_by_num":{
     "1":"Works with Experimental Web Platform features enabled",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #6327, opened 2022-06-13
fix typo

kmohrf:
I’m not a native speaker but still pretty sure the "no" is missing :)


Fyrd:
D'oh, nice catch, thanks. :)

PR #6330, opened 2022-06-16
Chrome 105: CSS Container Query Units

Schweinepriester:
https://bugs.chromium.org/p/chromium/issues/detail?id=1145970#c189

> Enable CSSContainerQueries for "stable"
>
> Enables:
> 
> - size container queries
> - inline-size containment
> - container length units
>
> Enable early on M105 to try to catch more fuzzer issues.

Diff: https://chromium.googlesource.com/chromium/src/+/ed08a64b71e627016a044f2d4899dcf07c7d0509%5E%21/#F0

Intent to Ship: https://groups.google.com/a/chromium.org/g/blink-dev/c/gwzxnTJDLJ8/m/iRNWbO2YBAAJ

There are no tests on <https://tests.caniuse.com/css-container-query-units> and I haven't found them on https://wpt.fyi/ 😅 

But once again, given the popularity I'm sure we/I will update if something changes :)


Fyrd:
Thanks! I've added some (lazy) tests that check if the units work in general, seem to confirm support in Safari TP and Chrome Canary.

PR #6008, opened 2021-09-01
Safari TP 131: `overflow: clip` 🎉 

Schweinepriester:
https://webkit.org/blog/11962/release-notes-for-safari-technology-preview-131/

> Implemented the `clip` value for `overflow` ([r280509](https://trac.webkit.org/changeset/280509/webkit/))

Haven't [test](https://tests.caniuse.com/css-overflow)ed it though.

PR #6334, opened 2022-06-20
Safari 16: `overflow: clip;` as well

Schweinepriester:
- <https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#browser_compatibility>
- https://github.com/mdn/browser-compat-data/pull/16586 ([the file change](https://github.com/mdn/browser-compat-data/pull/16586/files#diff-7b66b7fdab48139e06027de9230102721d78c31b3130a015ac87cae03d264a25))

also: whoops, I guess I was too eager with #6008 as https://github.com/mdn/browser-compat-data/pull/14815 proves me wrong - but meanwhile `"TP":"y"` is correct 0:-)

additionally looking at https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#browser_compatibility and https://tests.caniuse.com/css-overflow in Safari the picture isn't as clear regarding `overflow-x` and `-y`, so maybe there's room for improvement…


Fyrd:
Thanks!

PR #6341, opened 2022-06-22
Web archive link for ie.microsoft.com

forivall:
There might be more pages from ie.microsoft.com that need to be switched.

### DIFF
--- a/features-json/font-feature.json
+++ b/features-json/font-feature.json
@@ -5,7 +5,7 @@
   "status":"cr",
   "links":[
     {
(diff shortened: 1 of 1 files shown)


Schweinepriester (approved):

Schweinepriester:
Oh wow, didn't know this exists, noice!

> There might be more pages from ie.microsoft.com that need to be switched.

I found three (two, counting your current change): https://github.com/Fyrd/caniuse/search?q=ie.microsoft.com

Fyrd:
Went ahead and did all three manually, thanks!

PR #6348, opened 2022-06-26
AVIF coming to Safari & Safari on iOS 16

Schweinepriester:
- https://twitter.com/jaffathecake/status/1540697894683942912
- https://github.com/WebKit/WebKit/pull/1717

Given it's early and not yet [test](https://tests.caniuse.com/avif)able, draft for now.

@jensimmons are you able to comment on this? :)
If you are, do you know or are able to find out about support for AVIF sequences?
See note 2, Firefox doesn't have them yet: https://caniuse.com/avif
Test here: https://tests.caniuse.com/avif

### DIFF
--- a/features-json/avif.json
+++ b/features-json/avif.json
@@ -308,8 +308,8 @@
       "15.2-15.3":"n",
       "15.4":"n",
       "15.5":"n",
-      "16.0":"n",
-      "TP":"n"
+      "16.0":"a #3",
+      "TP":"a #3"
     },
     "opera":{
       "9":"n",
@@ -422,7 +422,7 @@
       "15.2-15.3":"n",
       "15.4":"n",
       "15.5":"n",
-      "16.0":"n"
+      "16.0":"y"
     },
     "op_mini":{
       "all":"n"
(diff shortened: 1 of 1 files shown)


jensimmons:
Support for AVIF is not in Safari 16 beta 2, so it is too early to make a change on Can I Use.

jensimmons:
AVIF support is in macOS Ventura and iOS 16, and we did just land support for AVIF in WebKit — so support in Safari is on its way!

I'll find out about support for AVIF sequences and let you know.

Schweinepriester:
Interesting development, potentially increasing pressure:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1686338#c28
- https://github.com/AOMediaCodec/av1-avif/pull/86

> Since animated AVIF files [now](https://github.com/AOMediaCodec/av1-avif/pull/86) use the same extension and MIME type as non-animated AVIF images, this bug constitutes an interoperability / web compatibility issue.

See the example there.

@Fyrd: Depending on how that plays out, maybe we should give partial for note 2 (for not supporting AVIF sequences)?

jensimmons:
WebKit's implementation does support image sequence and animations.

Fyrd:
Thanks!

PR #5893, opened 2021-05-07
Safari TP 124 improved support for `text-indent`

Schweinepriester:
- [ ] Figure out which state is actually correct

---

Starting this PR to get the ball rolling:

<https://webkit.org/blog/11672/release-notes-for-safari-technology-preview-124/>

> Enabled `hanging` and `each-line` keywords for the `text-indent` CSS property ([r275199](https://trac.webkit.org/changeset/275199/webkit/))

Looking at <https://tests.caniuse.com/css-text-indent>, TP 124 certainly improved on TP 123.

TP 123:

Bildschirmfoto 2021-05-08 um 00 20 14

TP 124:

Bildschirmfoto 2021-05-08 um 01 37 45

However it's not entirely green and also not the same the same as Chrome.

Chrome Canary 92.0.4500.0:

Bildschirmfoto 2021-05-08 um 01 42 12

Given that they talk about tests and improving spec compliance in the [changeset](https://trac.webkit.org/changeset/275199/webkit/) I'm wondering if <https://tests.caniuse.com/css-text-indent> is still correct or if support is truly still partial in TP 124.

If it's partial in TP 124, I'd change the PR to `a #3`, with note 3 stating the buggy behavior of `each-line`, ideally linking to a WebKit bug.

Neither <https://mdn-bcd-collector.appspot.com/tests/css/properties/text-indent> nor <https://wpt.fyi/results/css/css-text/text-indent> seem to cover `each-line`.

Ah, there are more tests at ([wpt.fyi for those](https://wpt.fyi/results/css/css-text/parsing?label=master&label=experimental&aligned&q=text-indent)):
- https://wpt.live/css/css-text/parsing/text-indent-computed.html
- https://wpt.live/css/css-text/parsing/text-indent-invalid.html
- https://wpt.live/css/css-text/parsing/text-indent-valid.html

buuuut they are the same:

TP 124:

Bildschirmfoto 2021-05-08 um 02 07 36

Chrome Canary 92.0.4500.0 (still with `#enable-experimental-web-platform-features` enabled):

Bildschirmfoto 2021-05-08 um 02 08 44

Looking at the [spec for `each-line`](https://drafts.csswg.org/css-text-3/#valdef-text-indent-each-line)
> each-line
> 
> Indentation affects the first line of each block container and each line after a [forced line break](https://drafts.csswg.org/css-text-3/#forced-line-break) (but not lines after a [soft wrap break](https://drafts.csswg.org/css-text-3/#soft-wrap-break)).

[...]

PR #6351, opened 2022-06-29
Safari 16 & TP 148: Full support of `text-indent` 🎉

Schweinepriester:
https://webkit.org/blog/12992/release-notes-for-safari-technology-preview-148/

> Added support for `each-line` keyword in `text-indent` ([250978@main](https://commits.webkit.org/250978@main))

But digging around there's https://bugs.webkit.org/show_bug.cgi?id=112755 resolved fixed:

> Enable "hanging" and "each-line" keywords for the text-indent CSS property

[MDN ](https://developer.mozilla.org/en-US/docs/Web/CSS/text-indent#browser_compatibility) has had that for almost a year via https://github.com/mdn/browser-compat-data/pull/10635.

The confusion is real, but anyway:
https://tests.caniuse.com/css-text-indent

image

Wuhu! :tada:

---

I checked and Safari 15.5 still has that intermediate state from before, hence I would leave <= 15.x as before:

image

=> Once this PR is resolved, https://github.com/Fyrd/caniuse/pull/5893 can be closed.
(Two of my oldest PRs closed then 🥳)


Schweinepriester:
I would strongly assume that this will ship with Safari & Safari on iOS 16 as well - @jensimmons, can you "encourage" that assumption? 0:-)
Should I change 16 already or should I wait for (the appropriate/next) beta?

Schweinepriester:
In case you were waiting for a response: The PR is now fully ready, @Fyrd :)

PR #6354, opened 2022-07-01
Web MIDI: Add WebKit support bug & update webkit_id

Schweinepriester:
- https://bugs.webkit.org/show_bug.cgi?id=107250
- https://webkit.org/status/#specification-web-midi

Even though they are "Not Considering" they haven't closed the bug yet.

They only thing I could find as to why they are not considering was https://webkit.org/tracking-prevention/.


Fyrd:
Thanks!

PR #6357, opened 2022-07-01
MathML: Update notes

Schweinepriester:
- Moved the note regarding Opera to a note with number to stick it to the old engine versions as the current general wording is soon about to become incorrect
- The "temporarily" for Chrome might be a bit cheeky, basically being 10 years, but this was a fast way to make it accurate again 0:-)


Fyrd:
Looks good to me, thanks!

PR #6363, opened 2022-07-07
Size Container Queries starting to ship in Chrome 105 🎉

Schweinepriester:
https://chromestatus.com/feature/6525308435955712

The source for note 3 is there as well:

> Note for M105: Table fragmentation in LayoutNG will not be shipped in M105, but expected to ship in M106. The consequence is that if an author combines table layout and size container queries inside a multicolumn layout, container queries will not work in that context in M105.

There doesn't seem to be a test on https://tests.caniuse.com/css-container-queries yet - I used 

```js
"container" in document.documentElement.style;
```

stolen from https://github.com/GoogleChromeLabs/container-query-polyfill:

image

Still `false` in 104 (currently Beta).


Fyrd:
Thanks! And added a test based on the "container" check. :)

PR #6374, opened 2022-07-11
Safari 16: Permissions API

Schweinepriester:
Same as https://github.com/Fyrd/caniuse/pull/6230; understandably given the support, only geolocation worked:


Fyrd:
Thanks!

PR #6386, opened 2022-07-20
Update flex gap support for Opera

foolip:
This was enabled in Chrome 84 and the corresponding Opera version is 70, not 73.

This will fix some inconsistency between entries:
https://caniuse.com/?search=flex%20gap


Fyrd:
Confirmed, thanks!

PR #6388, opened 2022-07-20
Updates to asm.js: Basically deprecated, links and minor changes

Schweinepriester:
- the deprecation note is a rrremix of both wikipedia and MDN
- copied the webkit id from [the chrome entry](https://chromestatus.com/feature/5053365658583040) - it's not there anymore, but I'd said its better this way than currently showing "WebKit status: Supported" (because of the wasm entry) which was never true for asm.js
- changing only the first chrome in note 1 is intentional


Fyrd:
Good changes, thanks!

PR #6395, opened 2022-07-23
Update audio.json

stacksize:
removing textopia.org link as it redirects to a blog that's promoting casino websites

### DIFF
--- a/features-json/audio.json
+++ b/features-json/audio.json
@@ -21,10 +21,6 @@
       "title":"Detailed article on support"
     },
     {
-      "url":"http://textopia.org/androidsoundformats.html",
-      "title":"File format test page"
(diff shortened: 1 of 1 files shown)


Schweinepriester:
Searched and found in https://web.archive.org/web/20170801000000*/http://textopia.org/androidsoundformats.html the seemingly latest version which was available:
https://web.archive.org/web/20170921185514/http://textopia.org/androidsoundformats.html

The samples seem to work accordingly, so theoretically we could keep the link around by replacing it with the archive version - though not totally sure if that's better than removing…

Fyrd:
Seems fairly useful, so have manually updated the link to the archive.org one. Thanks both!

PR #6398, opened 2022-07-24
Fix opera support for overflow: clip

artur-demidov:
Based on [caniuse overflow: clip](https://caniuse.com/mdn-css_properties_overflow_clip).
Opera is chromium browser, support should be similar to chrome/edge

P.S. 
I didn't test


Fyrd:
Tested and confirmed, thanks!

PR #4995, opened 2019-07-09
Added note about support in iPadOS 13

daytonlowell:
Source: https://developer.apple.com/documentation/safari_release_notes/safari_13_beta_3_release_notes

PR #6401, opened 2022-07-27
Updates to mediasource

Schweinepriester:
- unused notes 3 and 4 seem to have returned; ref #5470
- maybe just a fail on my side, but I spent way too long thinking about why note 2 is there if support starts with version 13 anyway - because the emphasis is on iPadOS, as in it's iPadOS only. hence here my attempt to make it subtly more clear; ref #4995


Fyrd:
Makes sense. Thanks!

PR #6390, opened 2022-07-21
Updates to U2F API & WebAuthn

Schweinepriester:
Fixes #6217.

Notes:

- links I haven't added which might be helpful to some:
  - https://developer.chrome.com/blog/deps-rems-95/#deprecate-u2f-api-cryptotoken
  - https://chromeenterprise.google/intl/en/policies/#U2fSecurityKeyApiEnabled
  - https://developer.chrome.com/origintrials/#/view_trial/-6366963973195038719
- I went with Chrome 106 because
  - they were talking about it here: https://groups.google.com/a/chromium.org/g/blink-dev/c/xHC3AtU_65A/m/27jV8fVsAgAJ
  - didn't seem they finalised the discussion
  - but https://developer.chrome.com/origintrials/#/view_trial/-6366963973195038719 includes 105, so I guess it was accepted
- potential todo: align Edge & Opera
  - Edge seems supportive here: https://groups.google.com/a/chromium.org/g/blink-dev/c/xHC3AtU_65A/m/Ai-evvu5AgAJ
  - but no other sign, so as opposed to chrome even less indication and since we don't seem to have a [test](https://tests.caniuse.com/u2f), I didn't touch Edge & Opera for now
- general question: safari? didn't [find an issue](https://bugs.webkit.org/buglist.cgi?bug_status=__all__&content=u2f&no_redirect=1&order=Importance&query_format=specific)…

Issue #6217, opened 2022-03-26
U2F API deprecated in Chrome v. 98


ph00lt0:
Please update the website a like.

Would be good to add distinction between the standards by adding FIDO2/WebAuthm API to the website separately. Currently only U2F API is mentioned which has stopped working.

https://www.yubico.com/blog/google-chrome-u2f-api-decommission-what-the-change-means-for-your-users-and-how-to-prepare/

ph00lt0:
@Schweinepriester correct! 

I didn't find that page earlier, it should also show up under [webauthm](https://caniuse.com/?search=webauthm&static=1) probably.

Schweinepriester:
Well, it does already show up for https://caniuse.com/?search=webauthn - but I got to ask, since you keep mentioning it with m at the end: Isn't it just webauthn, with n?

ph00lt0:
@Schweinepriester oh excuse for the consistent typo, my bad.

Schweinepriester:
Didn't manage to create a PR just yet, but some links for now:
- https://developer.chrome.com/origintrials/#/view_trial/-6366963973195038719
- https://chromestatus.com/feature/5759004926017536
- https://chromeenterprise.google/intl/en/policies/#U2fSecurityKeyApiEnabled

PR #6403, opened 2022-07-29
U2F => `unoff` & tag with typo again

Schweinepriester:
Following up what [you said here](https://github.com/Fyrd/caniuse/pull/6390#issuecomment-1193371304)

> [@Schweinepriester](https://github.com/Schweinepriester) Think we should change the status to `unoff`? The spec seems to have been a ["review draft"](https://fidoalliance.org/specs/fido-u2f-v1.0-nfc-bt-amendment-20150514/fido-u2f-javascript-api.html) so probably should already have been…

and I totally agree!

Additionally I'm trying to add the wrong `webauthm` tag again - you fixed the typo in #6390, but I intentionally wrote it with m, because that was one issue in #6217 :D
But if you understandably don't want tags-with-spelling-error (to increase findability), feel free to remove again :)

### DIFF
--- a/features-json/u2f.json
+++ b/features-json/u2f.json
@@ -2,7 +2,7 @@
   "title":"FIDO U2F API",
   "description":"JavaScript API to interact with Universal Second Factor (U2F) devices. This allows users to log into sites more securely using two-factor authentication with a USB dongle.",
   "spec":"https://fidoalliance.org/specs/fido-u2f-v1.0-nfc-bt-amendment-20150514/fido-u2f-javascript-api.html",
-  "status":"other",
+  "status":"unoff",
   "links":[
     {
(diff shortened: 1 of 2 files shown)


Schweinepriester:
Slightly related:
I had thought before if I should suggest an additional `status`, like `dep` or something, to mark a spec as deprecated right at the top and not only down using `notes` - what do you think?

Fyrd:
Oooh ha no you're right typos are fine in keywords, just didn't think about that at the time. :) 

Re: status, maybe, I feel it makes things easier to lump everything that browsers shouldn't be expected to support under a single one (for browser scores for example).  Also it's not too clear to me when it would count as deprecated? I imagine a number of specs would start out as "unofficial" then over time if there's not enough interest they'd count as "deprecated" but it may not always be clear when it's gone from one to the other...

PR #6407, opened 2022-08-02
Updates to Web Share API: Edge on macOS, support bugs, …

Schweinepriester:
Using https://web-share.glitch.me/ I tested Edge on macOS after [this comment](https://bugs.chromium.org/p/chromium/issues/detail?id=1162971#c19) and indeed it's there since at least 95:

image

(93 didn't work, 94 didn't want to give me the correct version on Browserstack.)

Technically we are most likely missing Linux support across the board, but since that wasn't mentioned before…

Links for convenience:
- Demos
  - https://web-share.glitch.me/
  - https://mdn.github.io/dom-examples/web-share/
  - https://w3c.github.io/web-share/demos/share.html added only this one, being the most 'official' and flexible one
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API
  - not sure if their support is totally correct, because it doesn't seem to work at all in Chrome on macOS for example
- Chromium
  - https://bugs.chromium.org/p/chromium/issues/detail?id=1144920
- Firefox
  - https://bugzilla.mozilla.org/show_bug.cgi?id=1653481
  - There was recently-ish movement in https://bugzilla.mozilla.org/show_bug.cgi?id=1772948 B-)


Fyrd:
Thanks!

PR #6408, opened 2022-08-03
Firefox datalist looks like it was fixed in 102

kohenkatz:
[The bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1474137) linked in note # 6 says that it was fixed in Firefox 102. I don't have 102 installed right now, but I tested in 104 and it does work there.

Here's a [working link](https://raw.githack.com/mitchhentges/firefox-dynamic-datalist-bug/master/index.html) to the test page cited in the report since the link in the bug report seems to be dead.


Fyrd:
Thanks!

PR #6413, opened 2022-08-08
Firefox 105: OffscreenCanvas 🎉

Schweinepriester:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1779009#c2
- https://hg.mozilla.org/integration/autoland/rev/ca0fc9488e5d
  > milestone	105.0a1

[Test](https://tests.caniuse.com/offscreencanvas)ing is kinda futile as it was enabled by default before in at least nightly and developer, so really only makes sense once 105 is released, so…
I'll continue to keep an eye on this o7


Fyrd:
Thanks!

PR #6423, opened 2022-08-15
Fix missing word in phrasing

vanillajonathan:


Fyrd:
D'oh, good catch, thanks!

PR #6427, opened 2022-08-20
Update css-has.json - Firefox supports behind flag

Thenakedgun:
[MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility) states:

> From version 103: this feature is behind the `layout.css.has-selector.enabled` preferences (needs to be set to `true`).

Tested on Windows with Firefox 103.0.2 and verified this is correct.

### DIFF
--- a/features-json/css-has.json
+++ b/features-json/css-has.json
@@ -175,9 +175,9 @@
       "100":"n",
       "101":"n",
       "102":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Confirmed, thanks!

PR #6429, opened 2022-08-21
Changed description and spec as HTTP/3 is now standardized

JanTrichter:
Changed description and spec link as HTTP/3 is now standardized.
The spec link now points to [rfc9114](https://datatracker.ietf.org/doc/rfc9114/) and the description has been updated to reflect the current state of the protocol as **Proposed Standard**.


Fyrd:
Nice, thanks!

PR #6434, opened 2022-08-29
Fix Chrome MathML Support Status

TheTripleV:
Chrome will not support MathML without a feature flag until version 108.

Source: https://chromestatus.com/feature/5240822173794304


Fyrd:
Thanks!

Issue #6047, opened 2021-10-07
Web bundles


ghost:
https://web.dev/tags/web-bundles/

jcayzac:
Also https://github.com/WICG/webpackage#web-bundles

PR #6437, opened 2022-08-30
feat: add subresource-bundling

jeffkaufman:
Chrome 104 enabled `<script type=webbundle>`.  See https://chromestatus.com/feature/5710618575241216


Fyrd:
@jeffkaufman No missing steps, though if you could include a test of any kind in which to detect this feature that would be helpful!

jeffkaufman:
@Fyrd is the https://www.jefftk.com/wbn-demo/ link the kind of test you mean?

Fyrd:
Yup, that'll do. Thanks!

Fyrd:
btw, is this the same feature as https://github.com/Fyrd/caniuse/issues/6047 ?

hayatoito:
#6047 seems a bit ambiguous. That doesn't mention any particular feature. It's just about Web Bundle format itself? 

This PR is about "Subresource Loading with Web Bundles", which is the actual feature which uses Web Bundle format to load subresources from there in Web.

jeffkaufman:
I agree that #6047 is more general and not quite the same as this. For example, I would interpret #6407 as including being able to navigate a top level browsing context to a web bundle.

jeffkaufman:
Thanks for the merge!

PR #6441, opened 2022-09-07
Add Firefox Android support for HTML media capture

Jap2-0:
Manually verified on the latest version; MDN suggests that support has existed since version 79 (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture).


Jap2-0:
Also: the current version of Firefox for Android on caniuse is listed as 101, but appears to actually be 104.

Fyrd:
Confirmed, thanks! Will look into updating the version number as well.

PR #6443, opened 2022-09-08
Update Firefox for Android features that should be at parity with desktop

Jap2-0:
I found a few features that had different statuses on Firefox desktop and Android without an obvious reason; I manually verified whether or not there were actual differences and found a few where (presumably) only one version was updated when both should have been.


Fyrd:
Oh wow I was updating the version and had the same thought to check on that! So I actually just made most of these updates. I'll see if I can resolve the conflicts.

Fyrd:
With the version number update it was easier to update the changes manually so have done that instead. Thanks though!

PR #6445, opened 2022-09-11
Tiny indirect typo improvement to HTML Media Capture

Schweinepriester:
Arrived here via https://austingil.com/html-capture-attribute/ 0:-)

Support on iOS seems to have improved, the `capture` attribute might even be supported, but I'm not fully convinced everything works 100 % after checking the [tests](https://tests.caniuse.com/html-media-capture), so I didn't touch it (yet?).


Fyrd:
Thanks!

PR #6452, opened 2022-09-13
Updates to AVIF following the release of Safari 16, with correct data

jensimmons:
- Moved information about AVIF support shipping in Safari 16 on macOS Ventura from 16.0 to 16.1.
- Added information about support for AVIF animated image sequences in Safari on iOS coming in 16.1.
- Updated Note 2 about animated image sequences to be more factual, and less opinionated.
- Updated Note 3 to have a needed "."
- Removed Note 4 "Does not support images with noise synthesis" because it is not accurate. Safari 16 does support film grain. The test that failed was likely created with an older AVIF creating library that had not yet been updated to the current specification.


Fyrd:
Thanks!

features-json/avif.json
```diff
@@ -509,9 +509,8 @@
   "notes":"",
   "notes_by_num":{
     "1":"Can be enabled in Firefox via the `image.avif.enabled` pref in `about:config`.",
-    "2":"Only supports still images. AVIF sequences are not supported!",
-    "3":"Only available on macOS 13 Ventura or later",
-    "4":"Does not support images with noise synthesis."
+    "2":"Supports still images. Animated image sequences are not supported.",
```
mathiasbynens, line 512: This note is currently only used for Firefox. Should it be set for Safari 16.0 as well?

Schweinepriester: @mathiasbynens but it is on Safari for iOS 16.0 and since Safari (for macOS) 16.0 doesn't support AVIF at all, it doesn't need to be there?

image

mathiasbynens: Ah, I see, my bad!

PR #6455, opened 2022-09-14
Improve description of `:has()` + add a resource

jensimmons:
Currently the description for `:has()` says:
>Only select elements containing specified content. For example, `a:has(>img)` selects all `<a>` elements that contain an `` child.

First, `a:has(>img)` doesn't select all `<a>` elements that contain an ``. It will only select those that have an `` as a direct child. Because it's `>img` instead of `img`. 

The easiest way to fix this without adding complexity is to simply remove the `>` and make it:  
>For example, `a:has(img)` selects all `<a>` elements that contain an `` child.

Second, I think the first sentence could be a bit more clear. I propose:
> Select elements containing specific content. For example, `a:has(img)` selects all `<a>` elements that contain an `` child.

In addition, I added a link to resources to include an [in-depth tutorial](https://webkit.org/blog/13096/css-has-pseudo-class/) on how to use :has() that walks developers through specific examples.

### DIFF
--- a/features-json/css-has.json
+++ b/features-json/css-has.json
@@ -1,12 +1,16 @@
 {
   "title":":has() CSS relational pseudo-class",
-  "description":"Only select elements containing specified content. For example, `a:has(>img)` selects all `<a>` elements that contain an `<img>` child.",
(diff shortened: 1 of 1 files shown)


Fyrd:
Good updates, thanks!

PR #6458, opened 2022-09-15
Clarifications regarding accessibility and `display: contents` 

jensimmons:
Here's a pull request with clearer details about what is and is not accessible when `display: contents` is applied to elements in Safari 16.0, Safari 16.1, and Safari Technology Preview.

### DIFF
--- a/features-json/css-display-contents.json
+++ b/features-json/css-display-contents.json
@@ -314,9 +314,9 @@
       "15.4":"a #2",
       "15.5":"a #2",
       "15.6":"a #2",
-      "16.0":"a #2",
-      "16.1":"a #2",
-      "TP":"a #2"
+      "16.0":"y #4 #5",
+      "16.1":"y #5",
+      "TP":"y #5"
     },
     "opera":{
       "9":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks! I was considering making this same change actually. :)

aardrian:
I strongly disagree with this PR changing the support status to "Supports" given all that I outlined in [PR #6451](https://github.com/Fyrd/caniuse/pull/6451#issuecomment-1247343838). And not just because

I have been tracking the impact of these bugs on users (not authors) for years and most recently testing with the Safari TPs as well. I am not relying on other teams telling me what does or does not work. I am actually testing (and doing so with users).

These years-old bugs are genuine barriers to disabled users.

Safari has now claimed _three times_ that these bugs are fixed. It has a vested interest in minimizing the perception of the impact of its bugs, and I feel this PR is serving that purpose instead of reflecting the genuine barriers.

These should be marked as "Partial support" because they genuinely _do not work_ in for core elements.

I don't want to get into a PR tit-for-tat, where I go and file a PR to undo a PR with which I disagree. I want this entry to reflect reality.

Issue #6450, opened 2022-09-12
Add passkeys


jensimmons:
Passkeys provide a new way to log in. Instead of having a username and password, you create a username and your device creates a cryptographic key pair — keeping the private key secret while sharing the public key with the website. Anytime you want to login again, your device communicates with the website, and authenticates that it's you. Using a passkey is far more secure than using a password, and it's far easier to use.

The set of technologies that make passkeys possible are defined in a set of open standards from the FIDO Alliance and the W3C. Learn more at: https://fidoalliance.org/multi-device-fido-credentials

WebAuthn is part of the story — and Can I Use already has a page for [Web Authentication API](https://caniuse.com/webauthn). But support for Web Authn alone does not signal whether or not a device or browser has support for passkeys. Web developers are going to want to know which browsers have support for passkeys as they start offering it as an option.

It would be great for Can I Use to add a page for passkeys. Since it's not a single API — it's an overall feature made possible by multiple APIs, only some of which are web-specific — this makes sense to be on Can I Use (and not in data provided by MDN BCD). Passkeys is a common noun, not a product brand. It's a word that can be used by anyone, much like the word "password". 

Currently, support for passkeys is available in Safari 16.0 for iOS, and will be available in Safari 16.1 for macOS. 

Both Google and Microsoft have announced their intentions to support passkeys. [Google](https://developers.google.com/identity/fido) says: "On Android, we aim to have passkey support available to developers towards the end of 2022." [Microsoft](https://techcommunity.microsoft.com/t5/microsoft-entra-azure-ad-blog/expansion-of-fido-standard-and-new-updates-for-microsoft/ba-p/3290633) said: "We enthusiastically encourage website owners and app developers to join Microsoft, Apple, Google, and the FIDO Alliance to support passkeys and help realize our vision of a truly passwordless world."

jensimmons:
Is adding a page for Passkeys something you are willing to do? 

Chrome 108 now has support. It shipped Dec 8, 2022. 
https://blog.chromium.org/2022/12/introducing-passkeys-in-chrome.html

Fyrd:
Probably, yes. Just want to be careful that we can test it appropriately so the data is accurate. The test at https://www.passkeys.io _may_ be sufficient, will need to take a closer look.

Synchro:
[...]

Issue #5649, opened 2020-10-13
Tracking various WebAuthn features


lgarron:
The [WebAuthn spec editor's draft](https://w3c.github.io/webauthn/) is verify much a living document, which has added or removed features every few months. There is an old "[level 1](https://www.w3.org/TR/webauthn-1/)" spec and an in-progress "[level 2](https://www.w3.org/TR/webauthn-2/)" spec, but the way browsers have introduced features does not correspond to these levels. Therefore, [https://caniuse.com/?search=webauthn](https://caniuse.com/?search=webauthn) is missing a lot of useful details for website authors. For example:

| Feature | Introduced to spec | Supported in browsers | Not supported |
|-|-|-|-|
| Basic security key support | long ago | [see caniuse.com](https://caniuse.com/?search=webauthn) | [Chrome on iOS](https://bugs.chromium.org/p/chromium/issues/detail?id=1101804#c4) (cannot use WebKit implementation) |
| [`uvm`](https://w3c.github.io/webauthn/#sctn-uvm-extension) | [Sep. 1, 2016](https://github.com/w3c/webauthn/pull/169) | ??? |  |
| [`requireResidentKey`](https://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-requireresidentkey) (boolean) authenticator selectionª | [Sep. 13, 2017](https://github.com/w3c/webauthn/pull/555) | [Chrome 70 on macOS](https://bugs.chromium.org/p/chromium/issues/detail?id=678128) (mid-2017)<br>[Chrome 73 on Windows](https://www.chromestatus.com/feature/4781249621131264)<br>[Firefox 66 on Windows](https://blog.mozilla.org/security/2019/03/19/passwordless-web-authentication-support-via-windows-hello/) (early 2019)<br>Safari on iOS 14 ([Sep. 16, 2020](https://en.wikipedia.org/wiki/IOS_14))<br>Safari 14 on macOS 11ᵇ (late 2020) | [Chrome on iOS](https://bugs.chromium.org/p/chromium/issues/detail?id=1101804#c4) (cannot use WebKit implementation)<br>[Chrome on Android](https://github.com/w3c/webauthn/issues/1457#issuecomment-692121194) (hopefully in 2021)<br>[Firefox on macOS](https://bugzilla.mozilla.org/show_bug.cgi?id=1536482) (stalled) | |
[...]

PR #6459, opened 2022-09-15
Add passkeys to Can I Use

jensimmons:
Creating the JSON file to add passkeys to Can I Use, to fulfill https://github.com/Fyrd/caniuse/issues/6450


jensimmons:
Once again, I have an error in the new file I'm trying to create. I often don't know how to handle the formatting of notes/bugs/resources when those fields are empty. 

Thoughts @Schweinepriester ?

Fyrd:
@jensimmons If you look at the validator result ("Details" link) you'll see the error `"Error: [passkeys] Failed hasCategories validation on "categories". Got this: Web API,security"`. I believe the categories should be `["JS API", "Security"]`.

I think you'll also want to leave the `parent` value as empty string and set any necessary `keywords` to something other than the sample default.

Fyrd:
It would also be great to have a way to test support for this to easily tell when support is added in other browsers. I looked around a little bit but wasn't able to find anything simple yet.

jensimmons:
I just asked our engineers if there's a demo from FIDO Alliance that could serve as a great test, but it seems like not. 

They did clarify that supporting passkeys includes:
1. User-verifying platform authenticator
2. Support for “conditional mediation” in CredMan
3. Platform it’s running on supports cross-device sign-in via the hybrid transport
4. Credentials are backed up and/or synced

ptman:
https://www.passkeys.io/ - has a demo, but aren't passkeys just using the WebAuthn API? So if you support WebAuthn at a specific level, you support passkeys? https://github.com/Fyrd/caniuse/issues/5649

ptman (approved):

jensimmons:
Passkeys are now supported in Chrome 108+

https://blog.chromium.org/2022/12/introducing-passkeys-in-chrome.html

jensimmons:
> aren't passkeys just using the WebAuthn API? So if you support WebAuthn at a specific level, you support passkeys?

No. WebAuthn is part of the stack -- the part a web developer codes.But support for passkeys is much more complex than that. It goes deeper into the underlying operating system, often involving security and various kinds of biometric credential management. 

I wish Can I Use would be willing to add a page for this. Web developers should be implementing passkey support on their websites and web apps, but there is not good data available on which browsers on which operating systems have support. Not having the resources developers need holds back the web from replacing passwords with a far more secure way to log in.

Fyrd:
Have merged & updated this PR, though still in the process of testing support and updating information about the feature. In particular I think notes may be necessary on different behavior for different OSes? Further help would certainly be appreciated.

Fyrd:
Wound up linking to https://passkeys.dev/device-support/ in the notes, seems pretty comprehensive.

PR #6462, opened 2022-09-18
Safari 16.1: Scroll-To-Text Fragment & Push API

Schweinepriester:
Now that the dust has settled a bit…

<https://developer.apple.com/documentation/safari-release-notes/safari-16_1-release-notes>:

> Added Web Push Notifications support on macOS Ventura.

> Added support for scroll to text fragment.

### Web Push Notifications

I'm currently not able to test that, but seeing it's not mentioned in the [16.0 release notes](https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes) I'm strongly assuming it's been moved to 16.1, akin to AVIF.

### Scroll to text fragment

* Technically the "Scroll To Text Fragment" experimental flag strangely isn't enabled by default in TP, but as it is in 16.1 I'm sure it will be soon
* Didn't test iOS 16.1, but I'm trusting the changelog as it doesn't mention a restriction; will test once released

Using the test via https://tests.caniuse.com/url-scroll-to-text-fragment:

https://user-images.githubusercontent.com/2644614/190901955-07a2b9ef-3c85-4623-8ebb-f772e2359980.mov

Also tried something more practical, using <https://www.urbandictionary.com/define.php?term=noice#:~:text=When%20something%27s%20nice%20af.>:

Bildschirmfoto 2022-09-18 um 13 27 01

🎉


Schweinepriester:
@jensimmons would you be able to review this PR, if everything is correct? :)

jensimmons:
Scroll to text fragment is correct. It could be merged, oh, except it's in the same PR as Push.

features-json/push-api.json
```diff
@@ -308,7 +308,7 @@
       "15.4":"n #3",
       "15.5":"n #3",
       "15.6":"n #3",
-      "16.0":"a #6",
```
jensimmons, line 311: This is correct. Safari 16 for macOS Ventura will be out in October, numbered 16.1.

features-json/push-api.json
```diff
@@ -308,7 +308,7 @@
...
+      "16.0":"n #3",
       "16.1":"a #6",
```
jensimmons, line 312: Here, I would change 16.1 to `a #3 #6`, because Safari will continue to support Safari Push Notifications for the foreseeable future. We are not dropping support. And I wouldn't want web developers to get worried, if the `#3` were to disappear.

jensimmons (changes requested):
Overall looks good. I did leave one note for line 312 for Push.

features-json/push-api.json
jensimmons: Ah, I should have marked line 312 for this comment... but you get the idea.

Schweinepriester: I do, thank you! :)

features-json/push-api.json
Schweinepriester: Thank you, have adjusted including note 3, hope it's good this way!

Schweinepriester:
Current feedback addressed, ready to be merged now, @Fyrd :)

Fyrd:
Thanks both!

PR #6464, opened 2022-09-22
Update HEVC support page

martijnberger:
https://bitmovin.com/google-adds-hevc-support-chrome/
https://bugs.chromium.org/p/chromium/issues/detail?id=460703

Validated on Chrome (105) on windows. macOS and Android.

### DIFF
--- a/features-json/hevc.json
+++ b/features-json/hevc.json
@@ -174,9 +174,9 @@
       "102":"n",
       "103":"n",
       "104":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Confirmed, thanks!

Fyrd:
Just noticed the values in this PR were changed for Firefox versions instead of the Chrome ones. Have fixed.

PR #6468, opened 2022-09-23
Chromium and Firefox both fail a11y on buttons — Option 2

jensimmons:
Adrian's testing shows both Chromium and Firefox have accessibility problems when `display: contents` is applied to `buttons`.

This PR keeps all browsers with most of the accessibility problems fixed as having full support — while marking each bright green box with a clear number that corresponds with the issue that web developers should be aware of. It sets Safari to supporting as well, to match what other browsers are getting.


Fyrd:
Closing as this has been resolved in a separate change.

PR #6470, opened 2022-09-25
COLRv1: Firefox 105+ flag & add link

Schweinepriester:
- https://material.io/blog/color-fonts-are-here
- literally matching the MDN entry: https://caniuse.com/?search=colrv1 => https://caniuse.com/mdn-css_at-rules_font-face_opentype_colrv1
- the bug https://bugzilla.mozilla.org/show_bug.cgi?id=1740530 plus changeset https://hg.mozilla.org/integration/autoland/rev/a5e90bc3b2e1
- looks like it might be coming by default in 107, but as there are no changes there yet I will update in the future: https://bugzilla.mozilla.org/show_bug.cgi?id=1791558
- finally, tested using https://tests.caniuse.com/colr-v1 => https://codepen.io/drott_chrome/pen/ExvaYxm
   - firefox [doesn't seem to support CBDT/CBLC](https://caniuse.com/mdn-css_at-rules_font-face_opentype_cbdt_cblc), so
      >  Emoji on left should be vector versions of emoji on right
      
      isn't true, but left matches Chrome

image

### DIFF
--- a/features-json/colr-v1.json
+++ b/features-json/colr-v1.json
@@ -523,7 +523,7 @@
   "notes":"",
   "notes_by_num":{
     "1":"Can be enabled with the \"COLR v1 Fonts\" feature flag under `chrome://flags`",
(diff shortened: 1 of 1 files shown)


Fyrd:
Nice, thanks!

PR #6479, opened 2022-10-01
Fix: HEVC is officially supported since M107 not M105

StaZhu:
1. HEVC support will be enabled by default since M107 not M105(https://chromiumdash.appspot.com/commit/f46f3399d7ff1119f1f1b9026d31110d7e5b9d8a), the current page is incorrect. Some user who are using chrome below version 107 but greater than version 104 maybe able to use the feature by passing `--enable-features=PlatformHEVCDecoderSupport` switch argument or by randomly hited "finch experment", but in general, not everyone can use it now.
2. Update the correct os and version detail that chrome current supports and also update the link to its support detail.
3. Update `Microsoft Edge` need to install `HEVC video extension` and only on Windows can use the feature.
4. Android support should also be started from M107.

### DIFF
--- a/features-json/hevc.json
+++ b/features-json/hevc.json
@@ -279,8 +279,8 @@
       "102":"n",
       "103":"n",
       "104":"n",
-      "105":"a #5",
-      "106":"a #5",
+      "105":"n",
+      "106":"n",
       "107":"a #5",
       "108":"a #5",
       "109":"a #5"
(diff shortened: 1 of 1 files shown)


yisibl:
Hi, can you investigate the situation in Samsung Internet and Opera?

StaZhu:
> Hi, can you investigate the situation in Samsung Internet and Opera?

I dont have a android phone nearby, also I dont think they support it for now since M107 just cut its branch half month ago, As for other browsers which also based on chromium, they should have this feature when they upgrade their broswer kernal to M107.

Fyrd:
Thanks!

PR #6484, opened 2022-10-04
Change Firefox support for WebAuthn to partial

iandunn:
It only supports FIDO2 PINs on Windows, and doesn't work on Mac/Linux/Android.

* https://bugzilla.mozilla.org/show_bug.cgi?id=1530370
* https://bugzilla.mozilla.org/show_bug.cgi?id=1530373
* https://github.com/mozilla/authenticator-rs/issues/33

Attempting to register or authenticate with the webauthn.io demo will fail with “user verification” required, but succeed with it disabled.


Fyrd:
Should the note perhaps be updated to be more descriptive?

iandunn:
👍🏻 

I'll also update it to note that FF [doesn't recognize TouchID as a provider](https://bugzilla.mozilla.org/show_bug.cgi?id=1536482)

iandunn:
I added a a bit about OS support to the description in 759d32509bc499cac83c9e98281aeaade8e53522. Let me know if there was anything else you had in mind.

leo60228:
Firefox for Android supports biometric authentication, that note should only apply to desktop (macOS).

Fyrd:
Thanks, and have updated per @leo60228's comment.

PR #6491, opened 2022-10-10
A minute update on SameSite and Safari

kiding:
Since Safari 14.1(.2) was still available on macOS Mojave, Note `#5` should still apply. The support was dropped along with the release of Safari 15.

References
- [Wikipedia - Safari version history](https://en.wikipedia.org/wiki/Safari_version_history#Safari_14)
- [WebKit - New WebKit Features in Safari 15](https://webkit.org/blog/11989/new-webkit-features-in-safari-15/)


Fyrd:
Thanks!

PR #6493, opened 2022-10-13
Update "Opus" to "Opus audio format"

Rupurudu:
This update makes opus title similiar to other audio codecs like aac, flac, and vorbis.


Fyrd:
Thanks!

PR #6500, opened 2022-10-20
Chrome 109: CSS Nesting behind flag

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=1095675#c13>:

> Chromium now has basic CSS Nesting support behind the experimental flag.

And indeed, <https://tests.caniuse.com/css-nesting> with the flag turned on:

image

I thought about adding "basic" or "experimental" in the note as the spec is still shifting I believe, but I think you mentioned `d` implies experimental before anyway?

### DIFF
--- a/features-json/css-nesting.json
+++ b/features-json/css-nesting.json
@@ -289,7 +289,7 @@
       "106":"n",
       "107":"n",
       "108":"n",
-      "109":"n"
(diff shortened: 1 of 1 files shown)


Fyrd:
Exciting! Thanks!

Issue #6505, opened 2022-10-24
iOS & iPadOS 16.1 don't animate the AVIF


Schweinepriester:
<https://tests.caniuse.com/avif>:

https://user-images.githubusercontent.com/2644614/197596582-6877560b-e0ba-4356-8c5a-44542e698ab9.MP4

Checked iPadOS 16.1 as well.

This time as an issue first, perhaps it was encoded using an old encoder as well?
Are there other examples showing it works?

CC @jensimmons 

EDIT:

The files in the Firefox bug at https://bugzilla.mozilla.org/show_bug.cgi?id=1686338 also don't work, having been added to the bug a year ago.
sequence-without-pitm.avif works in Chrome 106.0.5249.119 on Windows 10, sequence-with-pitm.avif flickers, but not sure if that is proper and sequence-with-pitm-avif-major.avif doesn't move.

jensimmons:
Please do create new images for your tests. The specs for AVIF changed. Image encoders have been updated — including recently when bugs in them were found.

Schweinepriester:
Fixed by https://github.com/Fyrd/caniuse/pull/6557.

PR #6508, opened 2022-10-26
ios 16 and 16.1 still doesn't support AVIF images taht are animated or that have noise synthesis

BuyMyMojo:


Schweinepriester:
Did you check on Safari on macOS as well? Is it only an issue on mobile?

Also related #6505 - are we not able to find an animated AVIF that works in Safari on iOS/iPadOS? 🤔

BuyMyMojo:
> Did you check on Safari on macOS as well? Is it only an issue on mobile?

I do not have a MacOS device I can easily check on, I'll see what I can do.

> Also related https://github.com/Fyrd/caniuse/issues/6505 - are we not able to find an animated AVIF that works in Safari on iOS/iPadOS? 🤔

I have tried making them in a few programs with the most basic settings to give them the best shot of at least being partially supported but none of them have been able ot load for me.

jensimmons:
Could you please add a description and update the title for this PR, so it's clear what this is about?

BuyMyMojo:
> Could you please add a description and update the title for this PR, so it's clear what this is about?

updated

Fyrd:
@jensimmons Do you dispute these findings at all? If there's any sample files the WebKit/Safari team can share that would also be very helpful.

Schweinepriester:
Same as #6505 I believe, this PR is now already done by https://github.com/Fyrd/caniuse/pull/6557.

Fyrd:
Closing as per last comment

PR #6513, opened 2022-10-31
Chrome 110: Drop JPEG XL

Schweinepriester:
https://bugs.chromium.org/p/chromium/issues/detail?id=1178058#c84

> Thank you everyone for your comments and feedback regarding JPEG XL. We will be removing the JPEG XL code and flag from Chromium for the following reasons:
>
> - Experimental flags and code should not remain indefinitely
> - There is not enough interest from the entire ecosystem to continue experimenting with JPEG XL
> - The new image format does not bring sufficient incremental benefits over existing formats to warrant enabling it by default 
> - By removing the flag and the code in M110, it reduces the maintenance burden and allows us to focus on improving existing formats in Chrome

---

See the replies, people are not happy, including me, but as of right now the change stands.

I will follow this and happily change it back if they change their mind.


jonsneyers:
It is what it is. Let's just hope they come to their senses on this topic sooner rather than later.

Fyrd:
Thanks for the update!

Schweinepriester:
@antermin Thank you, PR created: https://github.com/Fyrd/caniuse/pull/6620

PR #6515, opened 2022-11-03
APNG is now part of the official PNG specification

svgeesus:


features-json/apng.json
```diff
@@ -1,8 +1,7 @@
 {
   "title":"Animated PNG (APNG)",
   "description":"Like animated GIFs, but allowing 24-bit colors and alpha transparency",
-  "spec":"https://wiki.mozilla.org/APNG_Specification",
-  "status":"unoff",
+  "spec":"https://www.w3.org/TR/png",
```
Schweinepriester, line 4: Hi there Chris 👋 

Great that it's part of the spec now! :)

I think the CI expects the `status` to be present, so [from the list](https://github.com/Fyrd/caniuse/blob/main/CONTRIBUTING.md#supported-changes) I think `wd` would be most applicable?

Like so:

```suggestion
  "spec":"https://www.w3.org/TR/png",
  "status":"wd",
```

Fyrd:
Thanks!

PR #6519, opened 2022-11-05
Update Safari usage for logical shorthand

bakura10:
Hi,

I realized during testing that iOS 14.5+ and Safari 14.1 does not support margin-inline/padding-inline/margin-block/padding-block shorthand when the value is set as a CSS property (fixed in Safari 15).

This can be reproduced it here: https://codepen.io/bakura10/pen/wvXzzVR

image


Fyrd:
Thanks!

PR #6525, opened 2022-11-11
Update footnote 5

DanielvandeVelde:
The bug on footnote 5 `background-attachment` has been closed: https://bugs.webkit.org/show_bug.cgi?id=219324, therefore all items with that footnote can be closed and are now acknowledged as supporting `background-attachment`


DanielvandeVelde:
Updated the PR to make sure pipelines doesn't fail a second time.

rebtail (changes requested):
Change

Fyrd:
@DanielvandeVelde I just [tried this](https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment) on iOS 16.1 and the bug still seems to occur. Have you tested it yourself with other conclusions?

Fyrd:
Closing based on my own testing.

PR #6534, opened 2022-11-23
Safari: not honoring <base> in dynamic imports in inlined modules

jrencz:
https://bugs.webkit.org/show_bug.cgi?id=201692

Description is an extraction from comments in the bug ([only dynamic imports](https://bugs.webkit.org/show_bug.cgi?id=201692#c1), [inlined modules](https://bugs.webkit.org/show_bug.cgi?id=201692#c3)).
Today I bumped into that problem myself


Fyrd:
Thanks!

PR #6498, opened 2022-10-18
Firefox 108: Container Query Units

Schweinepriester:
https://bugzilla.mozilla.org/show_bug.cgi?id=1744231#c10

> Status: ASSIGNED → RESOLVED
Closed: 2 hours ago
[status-firefox108](https://bugzilla.mozilla.org/buglist.cgi?f1=cf_status_firefox108&o1=isnotempty): --- → [fixed](https://bugzilla.mozilla.org/buglist.cgi?f1=cf_status_firefox108&o1=equals&v1=fixed)
Resolution: --- → FIXED
Target Milestone: --- → 108 Branch

Pending the nightly build and FF 108 on caniuse 🙃

Issue #42, opened 2012-10-23
CSS3 "appearance" property


laukstein:
CSS3 `appearance` property  http://www.w3.org/TR/2004/CR-css3-ui-20040511/#appearance0
http://www.w3schools.com/cssref/css3_pr_appearance.asp
https://developer.mozilla.org/en-US/docs/CSS/-moz-appearance

Actually works prefixed just on Webkit.
Firefox related bug https://bugzilla.mozilla.org/show_bug.cgi?id=649849

I am not sure, but the property might be currently hold or removed from standards http://dev.w3.org/csswg/css3-ui/

PR #6535, opened 2022-11-23
Firefox 109 enabled support for CSS Container Queries and Units

shospodarets:
- [Here is the related issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1801123) closed by Firefox Dev team
- [Here is the codepen demo](https://codepen.io/shospodarets/pen/zYaWLLp) tested in Firefox Nightly 109
- Here is the screenshot from the working demo in Firefox: Here is the screenshot from the working demo in Firefox


Schweinepriester:
While I'm invested in this as well and very much looking forward to it (https://github.com/Fyrd/caniuse/pull/6498 ;)), I'm afraid the current change in [bug 1801123](https://bugzilla.mozilla.org/show_bug.cgi?id=1801123) is for nightly only, see https://hg.mozilla.org/mozilla-central/rev/335d18658070:

```diff
-  value: false
+  value: @IS_NIGHTLY_BUILD@
```

→ I'm afraid as of now your changes are not correct.

We could do something like
`"n #42"`
and the note then stating something like
"Can be enabled by setting `layout.css.container-queries.enabled` to `true` in about:config. Enabled by default in Nightly only."

shospodarets:
I just checked, and the intention is to enable it during 109 version going towards the stable channel.
I propose to go a simple way- let’s wait until 109 goes to Dev and towards stable (so “nightly” removed), and merge if all looks right,
or I will decline otherwise.

Schweinepriester:
Very fair!

To improve my process: How did you check?

shospodarets:
> Very fair!
> 
> To improve my process: How did you check?

By asking the team working on FF 😉

Fyrd:
Thanks!

shospodarets:
@Schweinepriester you appeared to be right, the FF team didn't change the "IF NIGHTLY" previous time due to a few found issues, so they moved the support enablement into version 110 🤷‍♂️🙂.
PR [is here](https://github.com/Fyrd/caniuse/pull/6567)

PR #6536, opened 2022-11-24
Font-weight problem on macOS 13 with Firefox 107

adamlaki:
If you use Firefox 107 on macOS 13 with `local()` to load the local font if present, it will break the font-weight property, and all of the font-weight will be 400. 

I couldn't find any related information online. It would be good if someone could validate the statement above.

### DIFF
--- a/features-json/fontface.json
+++ b/features-json/fontface.json
@@ -19,6 +19,9 @@
     },
     {
       "description":"Android 2.2 - 3.0 are reported not to support `local()` value"
+    },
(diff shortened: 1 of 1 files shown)


adamlaki:
Sorry, it seems that I made a mistake. We can omit the `local()`. When we add it, FF loads the fonts, but we should pack them by weights (400 -> Poppins, 700 Poppins Bold, and so on).

So no bug here, just misuse.

PR #6540, opened 2022-12-03
Update HEVC support status

StaZhu:
1. Edge Mac support HEVC since version 107.
2. Chrome Android support HEVC since version 107.


Fyrd:
Confirmed, thanks!

PR #6542, opened 2022-12-08
Chrome is shipping font-variant-alternates

Schweinepriester:
Intent to Ship
- https://twitter.com/intenttoship/status/1600505302410993667
- https://groups.google.com/a/chromium.org/g/blink-dev/c/IDXErld2g04/m/tJRD0Wu7CQAJ

> Sample links
> https://roettsch.es/montecarlo.html

https://chromestatus.com/feature/5070560260784128

With [this commit](https://bugs.chromium.org/p/chromium/issues/detail?id=716567) Chrome 111 now practically matches Firefox on https://roettsch.es/montecarlo.html:

Firefox:

image

Chrome:

image

stroke-width or something like that seems to be a bit different, but I guess that's just different rendering and as such fine.

Chrome Dev `Version 110.0.5481.24 (Offizieller Build) dev (64-Bit)` shows without support (with the flag disabled).

---

**Open questions/remarks**
1. Not sure what to do aboute note 1 on Chrome? 
Just remove it starting with 111?
Couldn't quickly figure out where it came from.
2. https://tests.caniuse.com/font-variant-alternates is red on both Firefox & Chrome, perhaps that test needs to be updated?
WPT appears to be https://wpt.fyi/results/css?label=master&label=experimental&aligned=&view=subtest&q=font-variant-alternates, esp. under `css-fonts/`.


Fyrd:
Yeah I think we can remove note 1 whenever there's support, I believe it's just there to let people know that they can get equivalent support using `font-feature-settings` where support is missing.

Also the test seems to work for me, possibly only on MacOS? May need to look into that if it's failing on Windows.

Schweinepriester:
Note 1 removed, PR ready to go!

Huh, now I'm really curious, because for me on macOS 12.6.2 it's also red for both Firefox & Chrome 111!?

image

Alternative/addition to changing the test: simply adding a link to https://roettsch.es/montecarlo.html there 0:-)

Fyrd:
That is weird indeed!
Screen Shot 2023-01-14 at 5 12 36 PM
Screen Shot 2023-01-14 at 5 14 05 PM

I'm on Monterey 12.6.0, that's the only difference but it seems unlikely...

Have added a link to your test case. :)

PR #6554, opened 2022-12-30
Updates to viewport units

Schweinepriester:
> sv* units, lv* units, dv* units and the logical vi/vb units.

stolen from https://chromestatus.com/feature/5170718078140416.

> `sv*` units (`svb`, `svh`, `svi`, `svmax`, `svmin`, `svw`), `lv*` units (`lvb`, `lvh`, `lvi`, `lvmax`, `lvmin`, `lvw`), `dv*` units (`dvb`, `dvh`, `dvi`, `dvmax`, `dvmin`, `dvw`) and the logical `vi`/`vb` units

is a bit unwieldy, but I thought it would be good to have them all for search results? Or move the bulk to tags?

MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/length

image

Random side note: https://trac.webkit.org/changeset/286458/webkit/

---

It would probably be nice to update https://tests.caniuse.com/viewport-unit-variants as well to cover more/all units :)

### DIFF
--- a/features-json/viewport-unit-variants.json
+++ b/features-json/viewport-unit-variants.json
@@ -1,6 +1,6 @@
 {
-  "title":"Large, Small, and Dynamic viewport units",
-  "description":"Viewport units similar to `vw` and `vh` that are based on shown or hidden browser UI states to address shortcomings of the original units. Currently defined as the `lvh`, `lvw`, `svh`, `svw`, `dvh`, `dvw` units.",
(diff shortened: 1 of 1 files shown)


Fyrd:
Hmm, tough one. On the one hand I do feel moving the full list to keywords would be better but that might not be obvious enough to all visitors. So let's try it like this for now. Thanks!

And yeah, I'll update the tests too.

PR #6563, opened 2023-01-10
Safari Push Notifications shipped in Safari 7, not 9.1

jensimmons:
Apple shipped Safari Push Notifications in Safari 7 for OS X Mavericks in October 2013.  https://web.archive.org/web/20210419000205/https://developer.apple.com/videos/play/wwdc2013/614/ https://www.macrumors.com/2013/10/21/websites-begin-supporting-safari-push-notifications-ahead-of-mavericks-launch/


jensimmons:
Currently the Can I Use page for "Push API" shows Safari 16.1 for macOS adding support for Web Push via Push API (along with Notifications API). That is correct. 

Screenshot 2023-01-11 at 4 31 19 PM

It also shows Safari 9.1 adding the original support for Push (in 2016), in note 3.

> Safari 9.1 - 16.0 supported a custom implementation which remains available in later versions, see [Safari Push Notifications](https://developer.apple.com/notifications/safari-push-notifications/) and [WWDC video](https://web.archive.org/web/20210419000205/https://developer.apple.com/videos/play/wwdc2013/614/)

It was while doing some historical research, I discovered that this is inaccurate. It was Safari 7.0 that first added support in 2013.

It might seem weird to be updating something that happened in 2013, but in writing about this topic, I realized this info is incorrect, and since Can I Use is deeply respected, I'd hate for folks who are writing about the history of push notifications and Safari to seem to be wrong if they site 2013 instead of 2016. 

Oh, I should update the note in the PR, not just the version numbers. Let me send a new PR.

Fyrd:
Thanks!

PR #6568, opened 2023-01-13
datalist is supported in firefox as of 109

boazsender:
The referenced bugs for datalists on inputs type [range](https://bugzilla.mozilla.org/show_bug.cgi?id=841942) and [color](https://bugzilla.mozilla.org/show_bug.cgi?id=960984) have been fixed and landed in 109, and 110 respectively. 109 will release next Tuesday, and 110 will release February 14th.


Fyrd:
Nice, thanks!

PR #6571, opened 2023-01-13
Add tags to File System Access API

### LINKED ISSUE #6570: Match OPFS search with filesystem api
Hi, this is a feature suggestion. Sometimes when looking for certain features on caniuse.com it will also match the feature set that it's part of. I found this lacking with a search for "OPFS" or "Origin Private File System". This should probably give results for the "[File System Access API](https://caniuse.com/native-filesystem-api)", as the chart includes information for the OPFS part

Schweinepriester:
Fixes #6570.


Fyrd:
Thanks!

PR #6575, opened 2023-01-16
Update css-nth-child-of.json

sesse:
Chromium-based browsers will support this syntax from M111:

https://chromium-review.googlesource.com/c/chromium/src/+/4166964


Fyrd:
Thanks! I removed the upcoming Edge versions for now, they're only tracked until the current version.

PR #6579, opened 2023-01-20
Chrome 112: CSS Nesting

Schweinepriester:
- https://mastodon.social/@intenttoship@botsin.space/109720903039657818
- https://groups.google.com/a/chromium.org/g/blink-dev/c/eFCrkiLynfU/m/JLsh3zQuAAAJ
- https://chromestatus.com/feature/5800613594529792

> Estimated milestones
> DevTrial on desktop 109
> DevTrial on Android 109
> Shipping 112

Chrome Canary is still at 111 right now, so https://tests.caniuse.com/css-nesting isn't testable, but I don't think it's blocking this PR as I'm sure we'll adjust if needed.
(Esp. since the syntax is still in flux.)

### DIFF
--- a/features-json/css-nesting.json
+++ b/features-json/css-nesting.json
@@ -298,7 +298,7 @@
       "109":"n d #1",
       "110":"n d #1",
       "111":"n d #1",
-      "112":"n d #1"
+      "112":"y"
     },
     "safari":{
       "3.1":"n",
@@ -536,7 +536,7 @@
   "parent":"",
   "keywords":"&,@nest,nested css",
   "ie_id":"",
-  "chrome_id":"",
+  "chrome_id":"5800613594529792",
   "firefox_id":"",
   "webkit_id":"",
   "shown":true


Fyrd:
Exciting, thanks!

romainmenke:
Can we revert this?

There is absolutely no reason to mark this as supported.

- no consensus on the specification
- there critical specification issues that still need to be resolved

_This is adding pressure on the process that isn't helping :)_

jensimmons:
"No consensus" is not accurate. "Critical issues" is not accurate. 
But even if that were true, why would Can I Use hold back information about browser implementation status? 

We should provide up to date information to web developers. There's precedence for providing such data about what browsers are doing, even before a specification is finalized. Just look at CSS Grid. The Can I Use page tracked implementations through five years of specification progress, 2012-2017. All before the feature shipped. 

Why would we not tell developers the status of browser implementations? There's no reason to hide objective facts.

romainmenke:
If vendors agree that they are shipping nesting no matter what, then yeah sure, add it to Can I Use :)

I understand that everyone wants to start communicating this feature.

I don't think it helps anyone to not wait a little bit until there is clarity on the last remaining issues. Just yesterday and the day before that a bunch of issues were discussed and a lot of progress was made.

At least knowing that a formal objection won't be filed would be good, but maybe that is just me :)

jensimmons:
No one has said "that they are shipping nesting no matter what". And there might still be a formal objection. But neither of those things affects whether or not it's appropriate for Can I Use to show that browsers are working on experimental/preview implementations. 

Again, look at CSS Grid. Can I Use reported that browsers were implementing for several _years_ before they shipped. The spec was absolutely still in flux during that time. There were many remaining issues... an unending list. That is the process for creating something new. And part of that process includes shipping into browsers like Chrome Dev, Firefox Nightly, and Safari Technology Preview — precisely so that web developers can be part of the process before the standard is locked. 

We want web developers to know Nesting is in progress, and to know where to find implementations. That way they can test it out. They can try it in multiple browsers and make sure it works the same way in each one. They can chime in with ideas or thoughts about the spec.... We want to include them, not exclude them by keeping the experimental implementations a secret. That's the whole reason browsers switched from prefixes to flags & preview browsers — so web developers can try things out and provide feedback. Long before something ships. Long before they are able to use the technology on production websites for customers. That's why Can I Use shows boxes in the future. And it's good for everyone for those boxes to be accurate. 

Green boxes vs red boxes on Can I Use has zero impact on the decision making process browser makers go through in deciding whether or not to ship. The green box is not a promise of when something will ship. Features get delayed all the time.

romainmenke:
> We want web developers to know Nesting is in progress, and to know where to find implementations. That way they can test it out. They can try it in multiple browsers and make sure it works the same way in each one. They can chime in with ideas or thoughts about the spec.... We want to include them, not exclude them by keeping the experimental implementations a secret. That's the whole reason browsers switched from prefixes to flags & preview browsers — so web developers can try things out and provide feedback. Long before something ships. Long before they are able to use the technology on production websites for customers. That's why Can I Use shows boxes in the future. And it's good for everyone for those boxes to be accurate.

I completely agree with this!
But that is not what is being communicated on Can I Use right now.
This does not capture any of that nuance. Maybe this is just me :)

Screenshot 2023-01-27 at 17 16 36

Issue #6583, opened 2023-01-24
Revert "nesting is supported"


romainmenke:
Can we revert this?

There is no reason to mark this as supported.

- no consensus on the specification
- there are critical specification issues that still need to be resolved

_This is adding pressure on the process that isn't helping :)_

_Originally posted by @romainmenke in https://github.com/Fyrd/caniuse/issues/6579#issuecomment-1401958113_

Schweinepriester:
I'm not opposed to reverting per se, but are you able to provide links for those two points?

Because in the [I2S thread for Blink](https://groups.google.com/a/chromium.org/g/blink-dev/c/eFCrkiLynfU/m/JLsh3zQuAAAJ) it sounds quite a bit more optimistic?

> It looks like the CSSWG resolved to adopt "Option 3",
with representatives from all 3 engines voting in favor - and the WebKit
blog survey also resulted in "Option 3" as the top choice.

> I think that we should ship this.

> Taking a look at the open spec issues (https://github.com/w3c/csswg-drafts/labels/css-nesting-1) some are explicitly ideas for future changes, but are there any where shipping amounts to a decision that isn't easily changed? I'm thinking especially of the CSSOM issues.

> Enabled feature for preview (TP?) here:
> 
> https://github.com/WebKit/WebKit/commit/ceafa556d06a475ae4b9362abeb64a42128ce9ee

romainmenke:
This might be the best reference at this time : https://github.com/w3c/csswg-drafts/issues/8249#issuecomment-1399386227

Fyrd:
Alright, has been reverted until it's more certain to be included.

Schweinepriester:
@romainmenke please keep us updated :)

romainmenke:
Thank you all! 🙇

PR #6591, opened 2023-01-26
CSS Nesting

Schweinepriester:
Previous: https://github.com/Fyrd/caniuse/pull/6579, https://github.com/Fyrd/caniuse/issues/6583

Now [Safari TP 162 is here](https://webkit.org/blog/13703/release-notes-for-safari-technology-preview-162/) as well:

> Enabled CSS Nesting ([258564@main](https://commits.webkit.org/258564@main), [258560@main](https://commits.webkit.org/258560@main))

---

Creating this draft PR to keep track with the intention of letting it sit here until the time is right.

CC @romainmenke


jensimmons:
What are you waiting for until "the time is right"? 

Why not publish now?

jensimmons:
Ah, I read the comment about not providing browser implementation data until after a specification is finished. 

I don't think that makes sense. It doesn't match what Can I Use normally does. Please do update the Nesting page with the data for Safari Technology Preview 162.

Fyrd:
Yeah I was under the impression that the discussions meant it might not be enabled in Chrome after all, so that was the reason for caution there. If that's not in dispute then no, I don't see any reason not to mark it as supported if indeed it's there.

PR #6596, opened 2023-02-01
Add a memory leak bug in Chrome for the lazy loading attribute

CyberAP:
`loading=lazy` has [a memory leak in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1213045) that still hasn't been fixed.


Fyrd:
Thanks!

PR #6608, opened 2023-02-10
Firefox 92+ ignores unsafe referrer policies

benediktwerner:
Source: https://bugzilla.mozilla.org/show_bug.cgi?id=1720294

Also updated the Safari note to mention the other two header values that would allow a longer referrer but are ignored due to ITP.


Fyrd:
Thanks!

PR #6610, opened 2023-02-11
Rename: CSS Paint API -> CSS Painting API

andyearnshaw:
CSS Painting API is the official name used by the [W3C Candidate Recommendation Draft][1]. Using the official name in a search on the caniuse.com website yields no results.

[1]: https://www.w3.org/TR/css-paint-api-1/


Fyrd:
Thanks!

PR #6627, opened 2023-02-25
Updates to image-set

atjn:
Changes:
- Safari has just landed support for the `type()` notation 🥳
  - https://github.com/WebKit/WebKit/pull/10623/files
- Safari doesn't support omitting the resolution
  - https://wpt.fyi/results/css/css-images/image-set/image-set-no-res-rendering.html
  - https://wpt.fyi/results/css/css-images/image-set/image-set-no-res-rendering-2.html
  - https://wpt.fyi/results/css/css-images/image-set/image-set-parsing.html
- Firefox doesn't support `calc()` for calculating the resolution
  - https://wpt.fyi/results/css/css-images/image-set/image-set-parsing.html

I would say that both issues are too minor to mark the browser support as `partial`. The Safari issue is completely solved by setting `1x` on each entry, and the Firefox issue seems very niche to me. It is however not something I feel strongly about, feel free to change it if you disagree.


Fyrd:
Looks good to me, thanks!

PR #6629, opened 2023-03-01
Update credential-management.json

hawc:
Removed broken link to Codelabs.

### DIFF
--- a/features-json/credential-management.json
+++ b/features-json/credential-management.json
@@ -13,10 +13,6 @@
       "title":"MDN Web Docs - Credential Management API"
     },
     {
-      "url":"https://g.co/codelabs/cmapi",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
I would propose (can't make a code suggestion on deleted lines apparently) to replace the the entry with https://web.archive.org/web/20200717113658/https://codelabs.developers.google.com/codelabs/credential-management-api/, which appears to work, and rename it to something like "Old Codelab tutorial".

But maybe Fyrd agrees with removing, then I have at least persisted it here 0:-)

Fyrd:
I think in this case the instructions there could get out of date pretty easily (if they aren't already) so removing is probably best in this case. Thanks!

PR #6632, opened 2023-03-04
Remove es7 from decorators keywords

senocular:
Early on as the decorators proposal first got momentum, it was (wrongfully) assumed that decorators would make it into ES7. That is not the case and searches for "es7" on caniuse today will bring up decorators even though they're not part of ES7.


Schweinepriester (approved):

Fyrd:
ohai Trevor, fancy seeing you here :)

And good catch, thanks!

PR #6548, opened 2022-12-14
Chrome is shipping WebGPU

Schweinepriester:
Intent to Ship
- https://twitter.com/intenttoship/status/1603075618291187713
- https://groups.google.com/a/chromium.org/g/blink-dev/c/VomzPhvJCxI/m/SUhU9Z0vAgAJ
- https://chromestatus.com/feature/6213121689518080

> We are targeting shipping in M113 though there's still some work to do, so this is an aspirational milestone and could slip one or two milestones.

Test: https://tests.caniuse.com/webgpu

~~UPDATE 2023-02-28:~~

~~Canary is 113 now, but it doesn't appear to be supported…~~

UPDATE 2023-03-16:

Appears to be enabled by default now:

image

Though https://gpuweb.github.io/cts/standalone/ is still the same compared to stable (382).

Marking as ready to go, I'm sure we'll adjust if needed o7

PR #6638, opened 2023-03-08
WebGPU is scheduled to arrive in Chrome 113...

haraldreingruber:
... according to https://chromestatus.com/roadmap


Schweinepriester:
I'm excited as well, so this is a duplicate of https://github.com/Fyrd/caniuse/pull/6548/, sorry!

Though I haven't found a way to ~proof~ test it's actually in 113 - maybe it has been shifted to 114 as they mention in the [intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/VomzPhvJCxI/m/SUhU9Z0vAgAJ)…?

There's https://github.com/gpuweb/cts but that currently appears to match stable Chrome.

haraldreingruber:
> Though I haven't found a way to ~proof~ test it's actually in 113 - maybe it has been shifted to 114 as they mention in the [intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/VomzPhvJCxI/m/SUhU9Z0vAgAJ)…?

I've tested https://webgpu.github.io/webgpu-samples/samples/shadowMapping on Chrome Dev which is currently v113, but I guess there is no guarantee that features won't be postponed before the make it into stable.

Schweinepriester:
but to me that looks the same in stable chrome (110/111), i.e. looks like it's polyfilled…?
(in case it matters: I'm on windows 10)

haraldreingruber:
True. Just saw it as well. Sorry for the confusion.

PR #6637, opened 2023-03-08
Safari TP 165: full `image-set` & `modulepreload`

Schweinepriester:
<https://webkit.org/blog/13932/release-notes-for-safari-technology-preview-165/>:

> -   Added support for `x` units in `calc()` function ([260678@main](https://commits.webkit.org/260678@main))
> -   Added support to `image-set()` for resolution and type as optional arguments ([260796@main](https://commits.webkit.org/260796@main))
> …
> -   Added support for `link` `rel=modulepreload` ([260761@main](https://commits.webkit.org/260761@main), [260709@main](https://commits.webkit.org/260709@main), [260659@main](https://commits.webkit.org/260659@main))

~~<https://tests.caniuse.com/css-image-set>~~

<https://wpt.fyi/results/css/css-images/image-set?label=experimental&label=master&aligned=&q=image-set>

image

<https://tests.caniuse.com/link-rel-modulepreload>:

image

and also wpt [presumably here](https://wpt.fyi/results/preload?label=experimental&label=master&aligned=&view=subtest&q=modulepreload).

~~Draft because I wasn't able to test yet.~~

🎉

PR #6539, opened 2022-12-01
Chrome is unprefixing -webkit-image-set

Schweinepriester:
https://bugs.chromium.org/p/chromium/issues/detail?id=630597#c57

> Unprefix -webkit-image-set

Potentially even more via https://chromestatus.com/feature/5432024223449088:

> However, our implementation has not been per the spec, as only URL values were accepted for the image, and only resolutions with 'x' as a unit were accepted. This changes our implementation to match the spec, accepting all <image> types or a <string> for the image, and all <resolution> types for the resolution. After these changes, image-set will be unprefixed.

WPT: https://wpt.fyi/results/?label=master&label=experimental&aligned=&q=image-set&view=subtest
As of 2022-12-01 Chrome is at 28 / 83.

EDIT:

Nope, doesn't seem like it, WPT still at 28 / 83.

But it's [110 already](https://bugs.chromium.org/p/chromium/issues/detail?id=630597#c57) (using [this](https://chrome.google.com/webstore/detail/crbug-release-indicator/ppbnemgliamfnlhhnclbibcpbnfdclnn)):

> Status: Fixed (was: Started)
> CANARY 110.0.5452.0

<https://tests.caniuse.com/css-image-set>:

image

PR #6641, opened 2023-03-10
Chrome is shipping `image-set` as well

Schweinepriester:
<https://bugs.chromium.org/p/chromium/issues/detail?id=630597#c63>

> Enable CSSImageSet
> 
> This patch enables the CSSImageSet flag by default. This adds the\
unprefixed "image-set" type so authors no longer need to use\
"-webkit-image-set". It also updates the implementation to the current\
spec with new resolution units (dppx, dpi, dpcm), image type support\
(e.g., type("image/avif")), raw urls without "url()", and gradient\
image options.

https://groups.google.com/a/chromium.org/g/blink-dev/c/JXcRP4MJc9I/m/qbkgbJvNBAAJ

> the last patches have landed in 113.

~~It doesn't appear to be in Canary yet without the experimental flag, surely in the next one, but~~ [WPT looks good already](https://wpt.fyi/results/css/css-images/image-set?label=experimental&label=master&aligned=&view=subtest):

image

(compared to yesterday for sure: https://github.com/Fyrd/caniuse/pull/6637)

New build arrived (113.0.5643.0)!
<https://tests.caniuse.com/css-image-set>:

image

As for 111 and 112, seems they backed out since https://github.com/Fyrd/caniuse/pull/6539 ([test](https://tests.caniuse.com/css-image-set)):

image


Fyrd:
Nice, thanks!

PR #6652, opened 2023-03-16
Firefox is shipping `:nth… of S`

Schweinepriester:
https://groups.google.com/a/mozilla.org/g/dev-platform/c/eQeeFKYJ038/m/EfFYIA87AAAJ

> As of today, I intend to turn :nth-child(An+B of <selector list>) and :nth-last-child(An+B of <selector list>) on by default on all platforms. It has been developed behind the layout.css.nth-child-of.enabled preference.

https://bugzilla.mozilla.org/show_bug.cgi?id=1808229

> Enable support for nth-child(An+B of selector list) and nth-last-child(An+B of selector list) by default

~~Draft because not clear which version it will be and was true in nightly before, so…~~

~~I went preliminary with 112 but might very well be 113.~~

113 it is:
https://hg.mozilla.org/integration/autoland/rev/3ba1f8c46718

> milestone	113.0a1

Test: <https://tests.caniuse.com/css-nth-child-of>

Is green now, but it was `@IS_NIGHTLY_BUILD@` before, so we'll have to trust for now.


Fyrd:
Thanks!

PR #6657, opened 2023-03-22
Updated HEVC support status to make it more accurate

StaZhu:
1. Actually macOS and Android also bundled a software decoder, on these platforms, as long as the system version meets the requirements, all devices will support it.
2. Linux and ChromeOS's support is very limited, since Chromium currently only support VAAPI for HEVC.


Fyrd:
Thanks!

PR #6663, opened 2023-03-28
Correcting definition of A2HS

jensimmons:
Removing this part of the definition of A2HS:
>To enable this behaviour, a website must serve a valid Web App Manifest and load it's assets through a [Service Worker](https://caniuse.com/#feat=serviceworkers).

That is the definition in some browsers, but not others. A simpler definition is better — especially to reflect the variation between implementations. Safari has never required a Service Worker for Add to Home Screen.


Fyrd:
Updated per my last comment

PR #5548, opened 2020-07-23
Safari 14: HTTP/3 is not enabled by default

kiding:
Reference: https://github.com/WebKit/webkit/commit/85a1ac7f69985c84709247d467085282ea5d0bb8

<details>
    <summary>Screenshots</summary>

</details>

PR #5484, opened 2020-06-22
Safari & iOS Safari 14: WebP, HTTP/3, BigInt, image-orientation and Picture-in-Picture 🎉

Schweinepriester:
Once Safari & iOS Safari 14 are available (#5483), this PR should be ready to go.

~Currently WIP~ Done i think.

PR #6664, opened 2023-03-28
HTTP/3 support shipped in Safari 14.0

jensimmons:
Support for HTTP/3 shipped in Safari 14.0. 
https://developer.apple.com/documentation/safari-release-notes/safari-14-release-notes

I don't know where the information about it being experimental only or for certain macOS versions only came from, but it's incorrect.


Schweinepriester:
I took the time to dig up the history:
- https://github.com/Fyrd/caniuse/pull/5484 → On by default
- https://github.com/Fyrd/caniuse/pull/5548: → Off by default
- https://github.com/Fyrd/caniuse/pull/5535 → Only macOS 11 Big Sur and later
- https://github.com/Fyrd/caniuse/pull/5836 → Back to off by default

CC @kiding @tunetheweb

Esp. because of [this comment](https://github.com/Fyrd/caniuse/pull/5535#issuecomment-693751687)

> Safari 14 has just been released for macOS 10.15 Catalina and 10.14 Mojave, and can confirm the two features are indeed os-dependent.

and Fyrd even confirming I'm very skeptical if on by default for all macOS versions is true.

Starting with at the latest Safari 16 it's probably true, [because](https://en.wikipedia.org/wiki/Safari_version_history#Version_compatibility) no older macOS version than 11 is supported.

Not sure what was tested in https://github.com/Fyrd/caniuse/pull/5836 causing the revert.

But speaking of testing - would you consider any of the tests linked on https://tests.caniuse.com/http3 to be a valid test?
(WPT doesn't seem to have tests for HTTP/3 or does it?)

Once we have a proper test we can naturally double check.

(If I had to guess what eventually turns out to be correct, it would be `a d #4` with the note stating macOS 11 Big Sur and later for 14 - 15 and `y` for 16+.)

tunetheweb:
I've just tested this on my iPhone (iOS 16.3.1) and it is not using HTTP/3:

image

It's also showing as disabled by default In Settings->Safari->Advanced->Experimental Features->HTTP/3 (enabling it and restarting Safari does show HTTP/3 is used, but when I reset these settings back to Default it is again not used).

On my Mac (16.3 on Ventura 13.2.1) it is showing as using HTTP/3:

image

However in the Develop->Experimental features it's showing as unticked by default - which is very confusing!

image

I don't see that menu option in Safari Technical Preview so maybe this option is not needed any more on desktop?

I'm updating both to 16.4 to see it it's any different but that could take a bit of time...

tunetheweb:
Confirmed that on Mobile 16.4 (iOS 16.4) that HTTP/3 no longer exists in the Develop->Experimental Features but it does **NOT** appear to be enabled when looking at an H3 site through Web Inspector, nor does https://cloudflare-quic.com/ say HTTP/3 is being used.

Confirmed that on Desktop 16.4 (Ventura 13.3) that HTTP/3 no longer exists in the Develop->Experimental Features but is enabled. No idea what version it was enabled from. I did test it when I raised #5836 and it did not appear to be working then despite that being raised well after Safari 14 was released.

radum:
Ventura 13.2 Safari Version 16.3 (18614.4.6.1.5) it is not enabled by default and I have an option under Experimental features, but it is disabled by default.

If I enable it works just fine.

jensimmons:
HTTP/3 is enabled for ~50% users at any given time, and not for 100% of users. One user testing and finding it off by default doesn't represent all users. It depends. Which is what is causing the confusion. 

The presence or absence of the feature flag doesn't signal anything about whether/not HTTP/3 is enabled. 

This does require macOS 11 Big Sur.

Let me create a new PR with correct information.

jensimmons:
I'm still gathering information about when the feature flag was removed from Safari.

features-json/http3.json
```diff
@@ -444,19 +444,19 @@
...
+      "15.2-15.3":"a #4",
+      "15.4":"a #4",
+      "15.5":"a #4",
+      "15.6":"a #4",
+      "16.0":"a #4",
+      "16.1":"a #4",
+      "16.2":"a #4",
+      "16.3":"a #4",
+      "16.4":"a #4",
+      "16.5":"a #4"
```
Schweinepriester, line 459: For `ios_saf` most likely note 5 was intended, right? :)
(Because iOS != macOS 0:-))

Like so:

```suggestion
      "14.0-14.4":"a #5",
      "14.5-14.8":"a #5",
      "15.0-15.1":"a #5",
      "15.2-15.3":"a #5",
      "15.4":"a #5",
      "15.5":"a #5",
      "15.6":"a #5",
      "16.0":"a #5",
      "16.1":"a #5",
      "16.2":"a #5",
      "16.3":"a #5",
      "16.4":"a #5",
      "16.5":"a #5"
```

features-json/http3.json
```diff
@@ -534,7 +534,8 @@
     "1":"Can be enabled in Firefox via the `network.http.http3.enabled` pref in `about:config`.",
     "2":"Can be enabled in Chrome(Chromium) via the `--enable-quic` & `--quic-version=h3-23` command line arguments.",
     "3":"Can be enabled in Chrome(Chromium) via the `--enable-quic` & `--quic-version=h3-29` command line arguments.",
-    "4":"Can be enabled in Safari on macOS 11 Big Sur and later in the \"Experimental Features\" developer menu."
+    "4":"Requires macOS 11 Big Sur or later."
```
Schweinepriester, line 537: A comma is missing to make the build happy, like so

```suggestion
    "4":"Requires macOS 11 Big Sur or later.",
```

Schweinepriester:
@jensimmons Thank you for looking into this! :)

For the future: How will we eventually know that Safari has turned HTTP/3 full on, i.e. 100 % of users get HTTP/3?
Is there any bug we can follow? Or will it be mentioned in the changelog, or…?
Please make it in some way or another transparent 🙏

jensimmons:
This PR should be correct now.

It does the following:
1. Corrects support for Safari 14+ on both macOS and iOS from none to partial. With a new note (Note 5) stating it's only enabled for a portion of users.  
2. Simplifies the note (Note 4) about the requirement for macOS 11 Big Sur or later, and removes that note from Safari 16+, since Safari 16 can only exist on macOS Big Sur (therefore _all_ Safari 16 users are on a version of macOS that has support). 
3. Updates the feature flag designation to show that the flag existed on both macOS and iOS from Safari 14–16.3. The flag was removed in Safari 16.4+.

jensimmons:
I don't have any information to share at the moment about when HTTP/3 will be turned on for all users. Or when the feature flag might return. Investigations are still underway regarding performance — we want to make sure this feature in fact does improve the experience of users. I expect when either change is made, we will list the change in Safari release notes.

Fyrd:
Many thanks!

x2q:
With Safari version 17 now released it seems like HTTP3 is now enabled by default.

jensimmons:
I don't believe anything changed in our HTTP3 support. We did not announce a change in our release notes or article. I am looking into it.

Schweinepriester:
@jensimmons As the topic came up again (https://github.com/Fyrd/caniuse/issues/7114), are there any news regarding HTTP3 support (by default), were you able to find something? :)

jensimmons:
Currently, HTTP3 is on for a very large portion of users, but not all. It depends. 
I don't have new news yet. Perhaps soon.

PR #6478, opened 2022-09-30
Safari TP: Partial Wake Lock API + MDN link

Schweinepriester:
- https://github.com/WebKit/WebKit/commit/46a272e5b78f15c439e2b28973c3211fdf36123b
- https://twitter.com/tomayac/status/1575941229673558016
- https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API

~Pending the actual TP.~

<https://tests.caniuse.com/wake-lock>:

image

However I went with partial as they didn't mentioned it in the changelog (given revision 255039 [TP 156](https://webkit.org/blog/13394/release-notes-for-safari-technology-preview-156/) would have been it; [TP 157](https://webkit.org/blog/13575/release-notes-for-safari-technology-preview-157/) for good measure) and also <https://wpt.fyi/results/screen-wake-lock?label=master&label=experimental&aligned=&view=subtest&q=wake+lock>:

image

---

If I had to bet, I'd say more parts are coming with another TP and then they will mention it in the changelog 0:-)

PR #6672, opened 2023-04-06
Safari 16.4 does fully support Screen Wake Lock

jensimmons:
Screen Wake Lock was originally marked "partial" support in [this PR](https://github.com/Fyrd/caniuse/pull/6478#issue-1393029166). 

There are still [WPT test](https://wpt.fyi/results/screen-wake-lock?label=master&label=experimental&aligned&view=subtest&q=wake%20lock) failures, but those failures are not because our implementation is lacking. We have full support.

The tests are failing for reasons unrelated to Screen Wake Lock, due to a lack of a WebDriver set permission command, which the testing framework uses. WPT is not the main test suite for Safari. These tests are written by other browsers for other browsers, and are often structured in ways that do not work universally.


Schweinepriester (approved):

Fyrd:
Thanks!

PR #6673, opened 2023-04-06
Changes to descriptions of AVIF and JPEG XL

wantehchang:
Copy the design goals of JPEG XL from https://jpeg.org/jpegxl/.

Remove the comparison between AVIF and JPEG XL, which is not essential to the raison d'être of caniuse: answering the "can I use ...?" question.

### DIFF
--- a/features-json/avif.json
+++ b/features-json/avif.json
@@ -1,6 +1,6 @@
 {
   "title":"AVIF image format",
-  "description":"A modern image format based on the [AV1 video format](/av1). AVIF generally has better compression than [WebP](/webp), JPEG, PNG and GIF and is designed to supersede them. AVIF competes with [JPEG XL](/jpegxl) which has similar compression quality and is generally seen as more feature-rich than AVIF.",
+  "description":"A modern image format based on the [AV1 video format](/av1). AVIF generally has better compression than [WebP](/webp), JPEG, PNG and GIF and is designed to supersede them.",
   "spec":"https://aomediacodec.github.io/av1-avif/",
   "status":"other",
   "links":[


wantehchang:
@Fyrd @atjn Please review. Thanks.

This pull request partially reverts atjn's changes in commit e535b18a261. My apologies. But it would be good for caniuse to focus on support data.

atjn:
> Remove the comparison between AVIF and JPEG XL [...] it would be good for caniuse to focus on support data.

Can you explain why that would be good? AFAICT users appreciate that features are shortly described and mentions other similar features. As I see it, your change will only serve to remove information that could be useful for many users.

As a quick example, if I have just heard of a new image format called JPEG-XL and I go to look it up on Caniuse, the impression that I will now get, is that the best image format on the web is WebP, since it is fully supported and only superseded by JXL, which is not yet supported anywhere. There is no mention of AVIF anywhere.

> Copy the design goals of JPEG XL from https://jpeg.org/jpegxl/.

I don't have any hard objections to this change, but I don't see why it is needed. I intentionally shortened the description to only focus on the web aspects because I believed that brevity was of more value than mentioning pro photography  features which are less useful on web pages. I did however still mention that JPEG-XL has more features, which indirectly includes the photography features without making the sentence longer.

wantehchang:
Anton: Thank you for your comment. I didn't realize you intentionally omitted professional photography in the description of JPEG XL. I will remove the changes to jpegxl.json from this pull request.

My answer to your first question is in the commit message:

> Remove the comparison between AVIF and JPEG XL, which is not essential to the raison d'être of caniuse: answering the "can I use ...?" question.

Fyrd:
Thanks for suggesting. While I agree caniuse should not be too opinionated on competing technologies, I think the statement offers value. There's plenty of other smaller ways in which caniuse tries to be helpful to developers beyond the "Can I use" question. 

If you disagree with the statement that it's "generally seen" to be more feature rich perhaps that can be removed or updated, but that does seem to be the case from what I've seen.

PR #6686, opened 2023-04-19
fix typos

int-y1:
These typos were found with codespell.


Fyrd:
Very nice, thanks for this!

PR #6688, opened 2023-04-19
Update that Safari 16.4 Web Push Notifications are delivered without …

Jamie4224:
As said in the title. Notifications aren't that useful if the user isn't actually notified. You don't get alerted when your phone is locked or you are not looking. When locked it will appear when you unlock the iPhone. It would be useful to know before implementing something that is basically useless.


Fyrd:
Probably depends on how important the notification is, I'm sure that varies depending on the usage. But certainly agree that's useful information to share!

Jamie4224:
> Probably depends on how important the notification is, I'm sure that varies depending on the usage. But certainly agree that's useful information to share!

I totally agree, I was in a bit of a rant mode. Glad this has been merged anyway, it is very useful to know in advance and might help someone decide if Safari iOS Web Push is beneficial to one's project.

Issue #6623, opened 2023-02-22
AV1 on Edge -> Partial?


indolering:
There is a disparity between how [HEVC support](https://caniuse.com/hevc) and [AV1 support](https://caniuse.com/av1) is calculated:

- HEVC is marked as partially supported, even when that support is contingent on hardware support AND the MS Store HEVC video extension.
  - The `extension && hardware` requirement might be a typo but I no longer have a Windows machine, so I can't test it.
- AV1 is marked as unsupported, even when it only requires the MS Store AV1 video extension (but not a hardware decoder).

It seems like the fix is to mark partial support when restricted by extensions or hardware.  This is a bit of a nitpick, but these numbers are being used to write [grossly misleading headlines](https://streaminglearningcenter.com/codecs/hevc-passes-av1-on-caniuse.html).  So the more precise we can make the wording, the better 😉.

indolering:
Another wrinkle to consider: according to the Edge extension [h.265ify](https://github.com/jlnsrk/h265ify), Edge doesn't report h.265 support even with the MS Store HEVC codec pack installed (so no playback by default).   It's not clear if this is also the case with the AV1 MS Store extension.  And I have no idea how hardware support impacts all this either ... is that where the `extension && hardware` requirement comes in?

But I don't have a windows machine and so I can't test anything ... which is the problem caniuse tries to help solve ☹.

indolering:
Apparently Edge + AV1 codec pack "just works" and [I suspect](https://github.com/jlnsrk/h265ify/issues/1) that h265ify is designed to work around website testing the incorrect HEVC MIME strings.  It may be that HEVC is only supported in one API but not another, but I'm still awaiting feedback.

Fyrd:
I spent some time looking into it and I believe the reason for the discrepancy is that HEVC is supported for Edge on MacOS, while AV1 is not. So _some_ users on Edge will have support by default for HEVC (hence partial) but not for AV1 (except on Windows + codec pack, hence disabled by default).

If you think any wording should be added to help clarification I'm happy to take another PR. Thanks!

PR #6689, opened 2023-04-25
Edge partial support for AV1

indolering:
Changed Edge from *disabled* to *partial* to match how HEVC support is calculated.  It might be better to mark HEVC as disabled to match how AV1 support is calculated, but AV1-> `partial` avoids drama. However, I'm not sure the preferred course of action because #6623 never got any comments.


Fyrd:
Closing as per my comment on #6623

PR #6706, opened 2023-05-11
Firefox 114: WebTransport

Schweinepriester:
https://bugzilla.mozilla.org/show_bug.cgi?id=1709355#c3

> Note added to 114 beta with this wording:
>
> [WebTransport](https://w3c.github.io/webtransport) is now enabled by default and will be going to release with 114. As the original [Explainer](https://github.com/w3c/webtransport/blob/main/explainer.md) notes, it enables multiple use-cases that are hard or impossible to handle without it, especially for Gaming and live streaming. It covers cases that are problematic for alternative mechanisms, such as WebSockets.
> 
> Built on top of HTTP3 (HTTP2 support will be coming later). The current implementation in Firefox is passing 505 out of 565 Web-Platform Tests.

<https://tests.caniuse.com/webtransport>:

image

<https://wpt.fyi/results/?label=master&label=experimental&aligned=&q=webtransport>:

image


Fyrd:
Thanks!

PR #6707, opened 2023-05-12
Update status of WebGPU for Safari

tobiasfabian:
WebGPU is no longer available as an experimental feature of Safari or Safari Technology Preview, at least not since version 15.

https://developer.apple.com/forums/thread/692979
https://github.com/gpuweb/gpuweb/discussions/3245

However, Apple is working on a (new) implementation of WebGPU.
https://github.com/WebKit/WebKit/pulls?q=label%3AWebGPU


Fyrd:
Thanks!

PR #6717, opened 2023-05-21
Remove redundant note from Firefox css-scrollbar support

lukewarlow:
Note 6 is redundant as the light and dark values were removed from the [`scrollbar-color`](https://drafts.csswg.org/css-scrollbars/#scrollbar-color) spec. And <length> values were removed from the [`scrollbar-width`](https://drafts.csswg.org/css-scrollbars/#scrollbar-width) spec.


Fyrd:
Thanks! If there's no more issues to mark support as partial then support can be changed to `y` for these browsers.

lukewarlow:
Ah yes I will change that to y instead. Firefox fully supports these properties (they're the only browser to implement them ATM)

Fyrd:
Thanks!

PR #6722, opened 2023-05-26
Firefox for Android: import maps

jmsche:
Hi,

It seems import maps is supported on Firefox for Android since version 108: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap#browser_compatibility

Does not seem I can change the version to v108, so I just updated the v113 information, but it would be nice if it could display that it works since v108?


Fyrd:
Yeah only tracking the latest version for Firefox for Android. Thanks!

PR #6724, opened 2023-05-27
fix grammar

int-y1:
1. Get all the text
2. Put text into Grammarly
3. Keep the most obvious suggestions


Fyrd:
Very nice, thanks!

riki714714:
Gracias

PR #6725, opened 2023-05-27
Update MathML support for Samsung Internet version 21.0

fred-wang:
Samsung Internet version 21.0 was released on May 19 and uses Chromium 110. MathML support was verified with [2].

[1] https://github.com/mdn/browser-compat-data/pull/19964
[2] https://fred-wang.github.io/MathFonts/mozilla_mathml_test/


fred-wang:
`npm run validate` because `sample-data.json` does not know about the new version of samsung browser. Note sure what's the process to add a new browser version.

Fyrd:
Has been added/updated!

PR #6729, opened 2023-05-30
Update css-grid.json for Safari bugs

aardrian:
Added information about Safari/iOS issues with tables that use `display:grid`. Includes link to Safari bug https://bugs.webkit.org/show_bug.cgi?id=257458.

### DIFF
--- a/features-json/css-grid.json
+++ b/features-json/css-grid.json
@@ -479,19 +479,19 @@
       "13.2":"y",
       "13.3":"y",
       "13.4-13.7":"y",
-      "14.0-14.4":"y",
-      "14.5-14.8":"y",
-      "15.0-15.1":"y",
-      "15.2-15.3":"y",
-      "15.4":"y",
-      "15.5":"y",
-      "15.6":"y",
-      "16.0":"y",
-      "16.1":"y",
-      "16.2":"y",
-      "16.3":"y",
-      "16.4":"y",
-      "16.5":"y"
+      "14.0-14.4":"a #5",
+      "14.5-14.8":"a #5",
+      "15.0-15.1":"a #5",
+      "15.2-15.3":"a #5",
+      "15.4":"a #5",
+      "15.5":"a #5",
+      "15.6":"a #5",
+      "16.0":"a #5",
+      "16.1":"a #5",
+      "16.2":"a #5",
+      "16.3":"a #5",
+      "16.4":"a #5",
+      "16.5":"a #5"
     },
     "op_mini":{
       "all":"n"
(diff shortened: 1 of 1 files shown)


Fyrd:
Are there any examples in the wild of developers applying these properties being applied to `tr` elements? I agree with the seriousness of the issue but if it's not a common use case (as we'd previously seen for display: contents and different types of elements) I think it's sufficient to add a note and link to the respective bug links.

aardrian:
> Are there any examples in the wild of developers applying these properties to `tr` elements?

I cannot share the ones that have prompted me to keep reporting these bugs as they are client projects.

However, I have had many banking and financial industry clients (who _really_ would not want me painting a target on their backs) over the last few years that rely heavily on data tables for their financial bits. In many of those cases I have prevented them from going into the wild because I know about these bugs.

But there a non-zero number of responsive table tutorials that advocate for these display properties. Some of them at least link to bugs I have been filing. There is at least one WordPress plug-in that uses CSS display properties on tables. I would have to do some spelunking to find cases, but without knowing the threshold you need I don't know how much effort to expend. If you would be satisfied with WordPress plug-ins with install counts I could start there.

Fyrd:
Thanks, that's helpful to know. Happy to take your word for it, just wanted to be sure it wasn't mostly theoretical. 

I'm wondering if instead of spreading this information across the different types of `display` values we should instead create a separate support table named something like "Arbitrary display values for table elements" or "Custom table display styling" that could cover any issues using such techniques. I feel like people would be more likely to search for something like that than just the display properties themselves. What do you think?

aardrian:
My concern is that these issues are not restricted to `<table>`s. I have not tested all possible variations; I just happen to have been working with client projects that use tables and CSS display properties.

For example, this WebKit bug shows that `display: contents` breaks far more than tables: [239478 AX: Support display:contents for elements with role attributes equal to: directory, grid, treegrid, table, row, gridcell, cell, columnheader, tree, treeitem](https://bugs.webkit.org/show_bug.cgi?id=239478)

So I am wary of limiting this to HTML `<table>`s simply because I have not taken my copious free time to perform testing across all these other roles and more.

jensimmons:
I agree that laying out tables using Grid is not a core use case for Grid. We should add the note, since this is more important than an obscure bug, but we should not downgrade support from full to partial. Safari supports Grid. 

Also, these bugs are fixed in Safari 17.0.

aardrian:
I can confirm Safari's table bugs are fixed in Safari iPadOS 17 (prior versions do not support CSS `grid` on tables). I maintain it is enough of an issue when critical information is involved that a downgrade (for soon-to-be older versions) is warranted.

PR #6730, opened 2023-05-30
Update flexbox.json for Safari bugs

aardrian:
Added information about Safari/iOS issues with tables that use `display:flex`. Includes link to Safari bug https://bugs.webkit.org/show_bug.cgi?id=257458.

### DIFF
--- a/features-json/flexbox.json
+++ b/features-json/flexbox.json
@@ -517,19 +517,19 @@
       "13.2":"y",
       "13.3":"y",
       "13.4-13.7":"y",
-      "14.0-14.4":"y",
-      "14.5-14.8":"y",
-      "15.0-15.1":"y",
-      "15.2-15.3":"y",
-      "15.4":"y",
-      "15.5":"y",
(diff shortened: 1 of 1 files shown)


jensimmons:
It's a great idea to mark Flexbox with this bug. But I disagree with downgrading Flexbox as a whole as "partially supported". The bug only affects tables that have `display: flex` on the `<tr>`s. 

Layout out HTML tables using Flexbox is not a core usecase. 

Let's include the footnote, but leave the level of support.

aardrian:
> It's a great idea to mark Flexbox with this bug. But I disagree with downgrading Flexbox as a whole as "partially supported". The bug only affects tables that have display: flex on the <tr>s.

I feel "only" is doing a lot of work there considering I found this because of a banking client using it this way (and another client considering it until I flagged it).

Given my small set of clients, I can only assume (perhaps poorly) it was more pervasive in non-public data table uses (where lots of financial data appears).

PR #6731, opened 2023-05-30
Update inline-block.json for Safari bugs

aardrian:
Added information about Safari/iOS issues with tables that use `display: inline-block`. Includes link to Safari bug https://bugs.webkit.org/show_bug.cgi?id=257458.

### DIFF
--- a/features-json/inline-block.json
+++ b/features-json/inline-block.json
@@ -465,19 +465,19 @@
       "13.2":"y",
       "13.3":"y",
       "13.4-13.7":"y",
-      "14.0-14.4":"y",
-      "14.5-14.8":"y",
-      "15.0-15.1":"y",
-      "15.2-15.3":"y",
-      "15.4":"y",
-      "15.5":"y",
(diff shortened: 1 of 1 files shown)


features-json/inline-block.json
```diff
@@ -465,19 +465,19 @@
...
+      "15.0-15.1":"a #2",
+      "15.2-15.3":"a #2",
+      "15.4":"a #2",
+      "15.5":"a #2",
+      "15.6":"a #2",
+      "16.0":"a #2",
+      "16.1":"a #2",
+      "16.2":"a #2",
+      "16.3":"a #2",
+      "16.4":"a #2",
```
aws11142, line 479: ****GIWA

jensimmons:
The fix for this problem is shipping Safari 17.0. 

Thanks for your dedication to finding such bugs, @aardrian.

aardrian:
I confirmed tables are exposed correctly in Safari iPadOS 17.

Issue #6736, opened 2023-06-03
Add Safari 17


jensimmons:
This might be a nice time to add a spot in the system for data about a potential future Safari 17.

Schweinepriester:
Aren't you a day or two early? 🤓

Only kidding 0:-)

+1

Schweinepriester:
Safari 17 was added in https://github.com/Fyrd/caniuse/commit/7fa607a13b560c578147e06e943223e12c52fb9c - now where is that changelog? 0:-)

Schweinepriester:
there we go: https://developer.apple.com/documentation/safari-release-notes/safari-17-release-notes 🥳

jensimmons:
Even more information: https://webkit.org/blog/14205/news-from-wwdc23-webkit-features-in-safari-17-beta/

PR #6659, opened 2023-03-23
Safari TP 166: @counter-style

Schweinepriester:
<https://webkit.org/blog/13964/release-notes-for-safari-technology-preview-166/>

> Enabled `@counter-style` by default ([261182@main](https://commits.webkit.org/261182@main), [260912@main](https://commits.webkit.org/260912@main), [261135@main](https://commits.webkit.org/261135@main))

<https://tests.caniuse.com/css-at-counter-style> etc.:

image

Because of https://wpt.fyi/results/css/css-counter-styles?label=master&label=experimental&aligned=&view=subtest

image

I went with partial for now (perhaps check once they have 166).
Additionally I highly doubt note 1 was implemented, but have neither found a test case to verify nor a bug to link to.

PR #6738, opened 2023-06-05
Safari 17: `image-set`, modulepreload, HEIF, `@counter-style` and more 🎉

Schweinepriester:
<https://developer.apple.com/documentation/safari-release-notes/safari-17-release-notes>

Fixes https://github.com/Fyrd/caniuse/issues/6736.

I have no way of testing yet.

## `image-set`:

> Added support for optional resolution and type arguments in image-set(). (77598590)
> …
> Fixed: Improved image-set compatibility. (105097744)

caniuse already had TP as `y`, so probably no surprises here.

## modulepreload:

> Added support for `<link rel="modulepreload">`. (88670991)

caniuse already had TP as `y`, so probably no surprises here.

## HEIF

> Added support for HEIC/HEIF images. (99517108)

I went for `y` for now as the changelog says, but if history is anything to go by, it will probably end up being partial for some time.

## @counter-style

> Added support for `@counter-style`. (30318695)

Aligned to `a` as was set in https://github.com/Fyrd/caniuse/pull/6659, still unclear what full support includes/excludes.

## text-box-trim & text-box-edge

> -   Fixed: Renamed `leading-trim` to `text-box-trim` to match the latest specification. (108788823)
> -   Fixed `text-edge` to be renamed to `text-box-edge` to match the latest specification. (108811923)
> -   Fixed `text-box-trim` renaming the `normal` value to `none`. (108925698)

Was also already in TP 171: https://webkit.org/blog/14200/release-notes-for-safari-technology-preview-171/
So note 1 is obsolete now 🎉


Schweinepriester:
Based on https://webkit.org/blog/14205/news-from-wwdc23-webkit-features-in-safari-17-beta/, primarily

> While the default web app experience on Mac is great, as a web app developer you can use web technologies like [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers) to make it even better. By providing a [web app manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest), you can customize the presentation of your web app, including the display mode, name, theme color, and start URL. Learn about the details by watching [*What's new in Web Apps*](https://developer.apple.com/videos/play/wwdc2023/10120/) at WWDC23.

I have amended the PR to have partial support for <https://caniuse.com/web-app-manifest>.

jensimmons:
Wait why is Add to Dock on Mac considered "partial support"??

> Based on https://webkit.org/blog/14205/news-from-wwdc23-webkit-features-in-safari-17-beta/, primarily
> 
> > While the default web app experience on Mac is great, as a web app developer you can use web technologies like [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers) to make it even better. By providing a [web app manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest), you can customize the presentation of your web app, including the display mode, name, theme color, and start URL. Learn about the details by watching [What's new in Web Apps](https://developer.apple.com/videos/play/wwdc2023/10120/) at WWDC23.
> 
> I have amended the PR to have partial support for https://caniuse.com/web-app-manifest.

There is nothing in the spec requiring the existence of a manifest file. We support manifest files on macOS. If a web developer includes one, it is used to enhance the experience of the web app. If there is no manifest file, user can still add any website to their Dock and open it as a web app. Safari simply provides defaults. Any web app developer can override those defaults in their custom manifest file any time they want. We did not remove support for manifest files from WebKit on macOS. 

These decisions follow the principles of the web — [support existing content](https://www.w3.org/TR/html-design-principles/#compatibility) and [priority of constituencies](https://www.w3.org/TR/html-design-principles/#priority-of-constituencies). The decision to support web apps for all websites is _not_ a lack of support for Web Manifest files. There is nothing in [the web standard](https://w3c.github.io/manifest/) that bans allowing users make web apps from websites that lack a manifest file. 

Therefore "partial support" is unwarranted.

Schweinepriester:
Saw https://github.com/Fyrd/caniuse/pull/6744 already, so lets continue the main discussion there, but to just explain my past self: 
I only picked partial because I couldn't (and can't) test and because I felt like https://caniuse.com/web-app-manifest is a bit unclear - Is it just A2HS (as the title implies), is it all of Web App Manifest (as the filename, URL and spec URL implies), or what is the scope? - therefore I defaulted to matching what was already there for iOS.

Let's make that clear, probably also with input from Fyrd and adjust the browser support accordingly in https://github.com/Fyrd/caniuse/pull/6744.

PR #6741, opened 2023-06-06
Update css-overflow-overlay for chromium 114

otomad:
In chromium 114 (chrome 114 and edge 114), `overflow: overlay` behaves the same as "auto", and no effect to overlay scrollbar again.

https://chromestatus.com/feature/5194091479957504


Fyrd:
Thanks!

PR #6743, opened 2023-06-06
Update css-text-box-trim.json

jensimmons:
Text-box-trim and -edge are in preview in Safari, not shipping. Meaning on by default in Safari Technology Preview, and available behind a feature flag in Safari.

And actually, it's been in Safari 16.4+ behind a flag.

Also... where is Safari 16.6 for iOS? It seems to be missing from the list, but it's in beta, currently, too. And has support for this behind a feature flag as well.


yisibl:
As far as I know, does `text-box-edge` only support partial property values? If so, please add the relevant description.

Fyrd:
Thanks! 16.6 was recently added, have updated the PR accordingly.

jensimmons:
> As far as I know, does text-box-edge only support partial property values? If so, please add the relevant description.

The spec continues to change in significant ways. There's no need to add notes about each change to the spec and each update made to Safari's implementation behind a flag to match the new version of the spec. That's far too much data. And none of it impacts users, since no browser will ship until the spec stops changing. This happens for every new technology, and the blow-by-blow of how specs evolve is not documented here.

yisibl:
@jensimmons OK, I agree. Thanks!

PR #6616, opened 2023-02-16
Safari 16.4: Push, `:dir()`, range syntax, … at lot of stuff! 🎉 

Schweinepriester:
<https://webkit.org/blog/13878/web-push-for-web-apps-on-ios-and-ipados/>
&
<https://developer.apple.com/documentation/safari-release-notes/safari-16_4-release-notes>:

> Added support for the `:dir()` pseudo-class.

Didn't test because it was in TP already.

---

> Added support for the range syntax from Media Queries level 4.

Didn't test because it was in TP already.

---

> Added support for Declarative Shadow DOM.

See https://github.com/Fyrd/caniuse/pull/6565 for discussion what to give Safari (and older Chrome).

---

> Added support for RegExp lookbehind assertions.

Didn't test because it was in TP already.

---

> Added support for Import Maps.

Didn't test because it was in TP already.

---

> Added support for lazy loading iframes.

Didn't test because it was in TP already.

---

> Added support for AVIF on macOS Monterey and macOS Big Sur.

Didn't test because it was in TP already.

---

> Added video-only support for Web Codecs.

Given it's mentioned like this, I have no reason to believe it's different from recently (https://github.com/Fyrd/caniuse/pull/6522).

---

> Added support for font-size-adjust.

Didn't test because it was in TP already.

---

> Added support for 2D-only OffscreenCanvas.

I believe https://caniuse.com/offscreencanvas is only concerned with 2D at the moment? TP was `y` already, so used that as well for now.
WPT interop 2023 also doesn't show a huge gap or anything [here](https://wpt.fyi/results/?label=experimental&label=master&product=chrome&product=firefox&product=safari&aligned=&view=subtest&q=label%3Ainterop-2023-offscreencanvas).

---

> Added support for the Screen Wake Lock API.

Went with partial, as at least TP [via WPT still is the same](https://wpt.fyi/results/screen-wake-lock?label=master&label=experimental&aligned=&view=subtest&q=wake+lock) as in https://github.com/Fyrd/caniuse/pull/6478.

---

> Added support for the unprefixed Fullscreen API on macOS and iPadOS.

~~Didn't test because it was in TP already.~~
Checked on iPadOS, the note is still applicable.

---

> Added support for third-party browsers to offer Add to Home Screen from the Share menu.

plus all over in https://webkit.org/blog/13878/web-push-for-web-apps-on-ios-and-ipados/, mentioning different manifest members
[...]

PR #6744, opened 2023-06-06
Update web-app-manifest.json

jensimmons:
Correcting data to clarify that Safari does in fact support web app manifest files.


features-json/web-app-manifest.json
```diff
@@ -1,7 +1,6 @@
 {
   "title":"Add to home screen (A2HS)",
-  "description":"The ability for a user to \"install\" a website and use it as if it was a natively installed app. To enable this behaviour, a website must serve a valid Web App Manifest.",
-  "spec":"https://www.w3.org/TR/appmanifest/",
+  "description":"The ability for a user to \“install\” a website and use it as if it was a natively installed app. ",
```
atjn, line 3:
```suggestion
  "description":"The ability for a user to \"install\" a website and use it as if it was a natively installed app. ",
  "spec":"https://www.w3.org/TR/appmanifest/",
```

atjn: That should make it pass

atjn:
I'm not sure I like that this never mentions the requirements to implement a service worker or a web app manifest. If I was developing on a mac, where Safari allows any website to be installed (still not sure how that works), I could easily miss the fact that the spec + other browsers don't allow that.

Maybe this is solved by changing the description from `The ability for a user to "install" a website` to `The ability for a user to "install" a PWA`? Maybe it should be even more specific than that?

I also think there should be a note for Safari to mark that it supports way more sites than other browsers do.

Schweinepriester:
> Maybe this is solved by changing the description from `The ability for a user to "install" a website` to `The ability for a user to "install" a PWA`? Maybe it should be even more specific than that?

Perhaps split that into two entries on caniuse, with `/web-app-manifest` becoming the latter?

<details>
  <summary>Maybe my initial rumbling was too much, but I'll include it below, because I think the scope should be clearer.</summary>
  
  
  I think it's vital to make sure first what the scope of https://caniuse.com/web-app-manifest actually is.
 
  * Just A2HS?
  * All of Web App Manifest?
  * A portion of Web App Manifest, perhaps setting a scope for A2HS?
  * Other options?
  
  Or maybe having all or some of those options as independent features on caniuse.
  
  And as usual deriving what to test for, being able to check that then.
  
  Maybe I'm remembering incorrectly but isn't the spec changing quite a bit still? Like aren't there new capabilities in this spec from time to time?
  
  <https://www.w3.org/TR/appmanifest/#sotd> goes in that direction
  
  > Implementors need to be aware that this specification is not stable. However, aspects of this specification are shipping in at least one browser (see links to implementation status at the top of this document). Implementors who are not taking part in the discussions will find the specification changing out from under them in incompatible ways.
  
  So all of Web App Manifest probably is not the best option, given we'd have to probably always demote once something new is added.
</details>

features-json/web-app-manifest.json
```diff
@@ -468,27 +467,27 @@
...
+      "14.5-14.8":"y #2",
+      "15.0-15.1":"y #2",
+      "15.2-15.3":"y #2",
+      "15.4":"y #2",
+      "15.5":"y #2",
+      "15.6":"y #2",
+      "16.0":"y #2",
+      "16.1":"y #2",
+      "16.2":"y #2",
+      "16.3":"y #2",
```
Schweinepriester, line 487: With the current state I disagree with Safari on iOS before 16.4 getting `y`, as note 2 was the - imho valid - reason for partial.

```suggestion
      "11.3-11.4":"a #2",
      "12.0-12.1":"a #2",
      "12.2-12.5":"a #2",
      "13.0-13.1":"a #2",
      "13.2":"a #2",
      "13.3":"a #2",
      "13.4-13.7":"a #2",
      "14.0-14.4":"a #2",
      "14.5-14.8":"a #2",
      "15.0-15.1":"a #2",
      "15.2-15.3":"a #2",
      "15.4":"a #2",
      "15.5":"a #2",
      "15.6":"a #2",
      "16.0":"a #2",
      "16.1":"a #2",
      "16.2":"a #2",
      "16.3":"a #2",
```

For reference:
https://github.com/Fyrd/caniuse/pull/6616
> > Added support for third-party browsers to offer Add to Home Screen from the Share menu.
>
> plus all over in https://webkit.org/blog/13878/web-push-for-web-apps-on-ios-and-ipados/, mentioning different manifest members
> → Preliminary for now I've simply removed the note from https://caniuse.com/web-app-manifest, but further research is needed!

Perhaps that research has arrived with this PR, changing 16.4 and onwards.

atjn: +1, this is a unique problem on iOS because third party browsers are blocked, but it essentially means that any feature not supported in webviews will demote iOS to partial support.

atjn:
To respond to your suggestion about splitting the table into two; the general idea behind the web app manifest was to make websites more app-like, and the natural extension of that was to make them installable (A2HS). That is why it started as the "web app manifest" table and I later renamed it to "A2HS". 
I think it is fine to keep the table the way it is, because caniuse is best for tracking this kind of generalised support for a large feature that covers many smaller features, while MDN will be a better place if you want to know the exact support for one specific field in the manifest. As you also mention, I think covering support for the web app manifest caniuse style, is almost as useful as trying to cover javascript in a single table.

Fyrd:
Thanks all! Agree with @Schweinepriester that the iOS partial support indeed refers to the feature not being available for 3rd party browsers, so keeping that for now.

Also agree the manifest is still worth mentioning, but I've rephrased it so it doesn't sound too much like it's required. I don't see any problem with allowing any website to be installable as web app - also added a note to mention this feature.

I also think keeping this particular support info related to basic web app installability makes more sense than to cover each new manifest feature. Could re-title it to include "(basic support)" if that feels necessary, similar to some other caniuse features.

Happy to hear any further feedback to these changes before I merge.

features-json/web-app-manifest.json
```diff
@@ -563,10 +563,10 @@
       "3.0-3.1":"n"
     }
   },
-  "notes":"",
+  "notes":"Safari 17+ also allows web sites without a manifest file to be installed as apps",
```
atjn, line 566:
```suggestion
  "notes":"Safari 17+ also allows web sites without a manifest and a service worker to be installed as apps.",
```

atjn (approved):
I have one suggestion, but with or without it, I think this looks good.

I suspect that other browsers will follow suit on lowering the installability criterias, and when that happens, I think we need to document the criteria more thoroughly.

jensimmons:
I'm a bit unclear where this is landing.

Is this page about "Add to Home Screen" or about Web App Manifest support?

If it's about Add to Home Screen, then we should add partial support for iOS all the way back to iOS1 — with a note that manifest files were not supported for Safari on iOS 1.0–11.2, but "Add to Home Screen" was. (Apple invented Add to Home Screen and shipped it with the first iPhone.)

If it's about general/basic support for Manifest files, then we should change the title of the page (currently "Add to home screen (A2HS)"). (No need to change the URL: https://caniuse.com/web-app-manifest or link to WD.)

Also, about the note 
>"Safari 17+ also allows web sites without a manifest and a service worker to be installed as apps."
It should be Safari on macOS... since this applies to macOS.

And we should add a note about iOS as well, since iOS has never required a Service worker. It's Chrome that requires a service worker for web apps.

atjn:
@jensimmons we are specifically tracking the ability to install a web app such that from the user-perspective, it works like an app installed from the app store. This will for example mean that the browser should ensure that the app's service worker is installed (such that it works offline), and when launching the web app, should allow it to launch in its own window without UI.

@jensimmons can you elaborate on how the add to home screen works for Safari without service workers? I don't have Apple hardware, so it is impossible for me to test.

My guess is that iOS Safari 1-11.2 A2HS was equivalent to "add link to home screen", which is supported in virtually all browsers. This is just a link that launches the browser, and therefore does not constitute A2HS support on caniuse.

As for iOS Safari 11.3+ and MacOS Safari 17+, I am guessing that you use the same approach when installing websites that don't provide a serviceworker? From the official demo, it seemed like you launch the app in its own chrome, and I guess that you just load the website normally, meaning it won't work without internet, but that is not your fault, the website should add a serviceworker for that. Or do you have some cool tech that caches the pages in a clever way?

Are my guesses correct?

atjn:
If my guesses are correct, I would suggest that we "by default" expect browsers to support A2HS for all websites, regardless of whether they have a manifest file or a service worker. Then we add a note for all browsers that have extra requirements, something like `Only supported for sites that link to a web app manifest and register a service worker` on all chromium-based browsers (would need to test that all browser require it).

jensimmons:
> we are specifically tracking the ability to install a web app such that from the user-perspective, it works like an app installed from the app store. This will for example mean that the browser should ensure that the app's service worker is installed (such that it works offline), and when launching the web app, should allow it to launch in its own window without UI.

This reminds me of the debate about "the mobile web" and how best to make a website that works on devices with small screens. Periodically there were strongly held opinions about how best to make a "mobile website". For a while, companies made two websites at two different domains, and used user-agent sniffing to redirect traffic. Later "responsive web design" became the trend. RWD had a clear technical definition, but that definition was _not_ encoded in a web standard. RWD = media queries + flexible media ( `img { max-width: 100%; }` ) + flexible layouts ( float based layouts with widths set in %s). 

For sure, media queries, `width` and `min-width`, %-based values for sizing, the ability to apply CSS to images to resize them, and the CSS `float` property are all key technologies that _are_ defined in web standards. And if you combine them in a certain way, you get RWD. Which is awesome. But Can I Use does not have a page for "RWD support". There's no web standard for RWD. RWD is a concept, a technique. It's not a technology. And now that float-based layouts are no longer the best practice (in my _opinion_, sites should use CSS Grid instead), RWD as original defined is showing its age. 

PWA is similar. There are many cool web technologies that can be combined together to create fantastic experiences. But there is no web standard that defines a particular combination as "the correct way". It's up to people who make websites to decide what kind of experience they want to provide, and to figure out which technologies to use to create that experience. 

It does not make any more sense for Can I Use to provide a page for PWA than it did for RWD. Not every project is trying to "work like an app installed from the app store". It depends on the project whether "launching in its own window without UI" is the best design. 

Can I Use is all about empowering web designers and developers with facts, so they can decide what to do. This is not the right place to evangelize a particular world view of how websites or web apps "are suppose to do it".  

It _is_ a place to determine if web browsers are "doing it right" — where doing it right is objectively defined in a web standard. Without a web standard, it's just opinions, ideas and trends. And this idea of "be like app stores" is not defined in any web standard. That phrase begs many questions — which App Store? Apps on which operating system? In which decade? As apps and app stores continue to evolve, what would that mean if Can I Use did try to follow "is this browser doing it right to support web apps that work like apps from the app stores"? It's like trying to hit a moving target. It's far better to have a page for Service workers, a different page for manifest files, another page for web push, yet another page for CSS Grid, another for the next thing web apps want to use... on and on. For all of the web technologies, separately, like they are intended to be used. As a mix-and-match as you'd like toolkit for web development. Not a paint-by-numbers formula. Leave that to the third-party frameworks.

jensimmons:
> Can you elaborate on how the add to home screen works for Safari without service workers? 

On iOS, if a web developer includes a manifest file with a `display` mode of `standalone` (or `fullscreen`), it opens as a web app when the user taps the icon from their Home Screen. The developer _can_ provide a Service Worker — and do things with it: support offline, implement web push, etc. But the lack of a Service Worker does not preclude the project from opening as a web app. Only a manifest file is required on iOS and iPadOS. (And not required at all on macOS. All sites that have been added to the users Dock will open as a web app. The manifest file allows the developer to alter the defaults for how the window looks/works, and more.) 

Without a manifest file (or with a manifest file that has a `display` mode of `browser` (or `minimal-ui`), when the user taps the icon from their Home Screen on iOS and iPadOS, the website opens in the users default browser. 

> This is just a link that launches the browser, and therefore does not constitute A2HS support on caniuse.

It does open in the default browser. But why does that "not constitute A2HS support"? It's literally called "Add to Home Screen", and has been for many — for longer than manifest files have existed.

If we want to call this Can I Use Page "Add to Home Screen", then it should be about Add to Home Screen. If we want it to be about manifest file support, then it should be called Manifest file (or Web App Manifest / Web Application Manifest). Honestly, it might make sense to have two separate pages. 

In no case, should we call this page "A2HS" or "Web App Manifest", but then only consider a browser to have "full support" if it's following the definition of "PWA". (See my comment above.)

> I don't have Apple hardware, so it is impossible for me to test.

For more on how web apps on iOS and macOS use the manifest file, watch (or read the transcript from) "What's new in web apps": https://developer.apple.com/videos/play/wwdc2023/10120/

jensimmons:
> …web app such that from the user-perspective, it works like an app installed from the app store. This will for example mean that the browser should ensure that the app's service worker is installed (such that it works offline), and when launching the web app, should allow it to launch in its own window…

In some ways, you are hitting on where Apple is coming from. This is about the user's perspective, not a particular tech stack. If they tap or click on the icon, and it opens in its own window like other apps, then it's a web app. Exactly.

But that doesn't mean it _has to_ work offline to be a web app. (Have you seen how many apps don't work offline? Tons!) It doesn't mean it has to be for productivity. It can be any flavor of website that a user adds to their Home Screen / Dock. A book. A photo album site. A photo editing site. A social network. Their own personal site. Their favorite restaurant. Or blog. Whatever. Anything can be a "web app". It can be built from WordPress, React, Vue, use Bootstrap, Tailwind, or none of the above. It doesn't matter. If it's been added to the Home Screen or Dock. And it opens in its own window, then it's a web app. Period. On macOS, it could be a single HTML page, with zero JavaScript, no Service Worker, no Manifest file, nothing but a single HTML file, and if the user adds it to their Dock, it's a web apps. Because that's the _users_ experience. 

All the more reason why "A2HS" and "Web manifest" perhaps should be two different pages on Can I Use.

atjn:
Thank you for elaborating on how Safari does this!

If I understand you correctly, your issue is that we are calling this `Add to home screen` because that name also covers other things, such as simply launching the website in a browser.
I can see why that might be a confusing name. We could change this to `Install to home screen` or `Install website to device`. I think that would make it clear.

We are specifically tracking _offline install_ and _fullscreen launch_ as a combined feature here because they allow for the app to appear as though it is natively installed. If you press a natively installed app icon, you would never end up in a browser. The intent is not to evangelize that this is the "correct" way for a website to work, we are just tracking which browsers would allow a website to work that way if it wants to.

The more general "add a browser link to the home screen" should be tracked in a separate table, but I am not sure the table would be very useful, as I think most developers know that this is supported in all browsers.

jensimmons:
> Your issue is that we are calling this Add to home screen because that name also covers other things, such as simply launching the website in a browser.

No, I don't think that's right. Add to Home Screen _is_ about web apps. It's just that 1) it predates the existence of web app manifest files, and 2) some implementations allow websites to become web apps without a manifest file. So these two things are separate and should be untangled.

I do not believe using "Install" is a good replacement for either Add to Home Screen or Web App Manifest, because it came much later, and is a UI unique to certain operating systems. It's not where this started, nor is reflective of the manifest specification. 

> We are specifically tracking offline install and fullscreen launch as a combined feature here because they allow for the app to appear as though it is natively installed.

_This_ is the problem. Quite literally the original editor of the Web App Manifest specification repeatedly states that these two things should _not_ be combined into any sort of "must be done together". They are not any more connected than media queries are to flexible images. The web allows for technology to be combined as developers desire. Such ideas are defined in the HTML Design Principles, baked into the fabric of the web itself: https://www.w3.org/TR/html-design-principles/

atjn:
> No, I don't think that's right. Add to Home Screen is about web apps

We are not tracking support for the concept of web apps in this table. As you said yourself, "web apps" is a broad term that cannot be defined as a single feature, so it would not make sense to track support for it.

> I do not believe using "Install" is a good replacement for either Add to Home Screen or Web App Manifest, because it came much later, and is a UI unique to certain operating systems. It's not where this started, nor is reflective of the manifest specification.

That is fine, because we are not tracking support for A2HS, and we are not tracking support for the Web App Manifest. We are only tracking support for that specific UI unique to certain operating systems.

> This is the problem. Quite literally the original editor of the Web App Manifest specification repeatedly states that these two things should not be combined into any sort of "must be done together".

We are not stipulating that these must be done together.
- If you are a developer and you want your website to launch in full screen, you can easily check for support in this table: https://caniuse.com/mdn-html_manifest_display
- If you want to have your website work offline, you can check for support in this table: https://caniuse.com/serviceworkers
- If you want your website to work and appear the same as a natively installed app, you can check out this table. That is the only thing this table is tracking.

features-json/web-app-manifest.json
```diff
@@ -1,6 +1,6 @@
 {
   "title":"Add to home screen (A2HS)",
-  "description":"The ability for a user to \"install\" a website and use it as if it was a natively installed app. To enable this behaviour, a website must serve a valid Web App Manifest.",
+  "description":"The ability for a user to \"install\" a website and use it as if it was a natively installed app. Metadata for the web app like name, links to icons, preferred URL, etc can be configured in the Web App Manifest.",
```
marcoscaceres, line 3: Just to clarify a little bit, the ability to "install" is just one capability provided by Web Manifest: web manifest can be used to, for instance, set the theme-color of a web page without installing the application first. The manifest can also be used in other contexts (e.g., it could hypothetically be used to populate a new tab page). 

```suggestion
  "description":"The ability for a user agent make use of web manifest metadata, which includes things like the theme-color, name, links to icons, start URL, etc. for a web application.",
```

marcoscaceres:
Chiming in as Editor of the Web Manifest spec (and previously as maintainer of Firefox's Web Manifest implementation, and now one of the core maintainers of WebKit's Manifest implementation). 

Tl;dr: please don't make this "Add to Homescreen (A2HS)". It's tying things to browser UI experience. Please use the general "Web Manifest" instead. 

> We are specifically tracking offline install and fullscreen launch as a combined feature here because they allow for the app to appear as though it is natively installed. 

This is codifying one particular experience (the Chrome one, it sounds like) and goes against the spirit and intent of the Web Manifest spec. Basically, what @jensimmons said above 👆.  Please don't do that!

> If you press a natively installed app icon, you would never end up in a browser. The intent is not to evangelize that this is the "correct" way for a website to work, we are just tracking which browsers would allow a website to work that way if it wants to.

This doesn't make any sense from a "can I use" perspective because that's a browser UI affordance: By this logic, why not then also have a "bookmarks" section to answer "can I bookmark pages"? Or a "History" section, to "can I look at my browsing history?" section. 

As @jensimmons pointed out, the ability to "Add to Homescreen" is unrelated to any web standard (going back to iOS 1.0!). It just happens that Web Manifest assists with that by providing a bit of metadata, icons, etc. But web manifest is completely optional to adding to home screen and getting an "app like" experience. 

> We are only tracking support for that specific UI unique to certain operating systems.

This is not good: you are basically tracking "is it Chrome"? Please don't do that. Worse, this is documenting arbitrary browser UI and has nothing to do with "can I use Web Feature X?"

Again, I respectfully ask that instead you focus on: "does the user agent support and make meaningful use of Web Manifest" (i.e., "can I use web manifest with browser X?").

So, to echo what @jensimmons said above: if this is about "Add to home screen (A2HS)" then @jensimmons is absolutely right that iOS 1.0 was the first version to support "A2HS". If this is about "Web Manifest" support, then @jensimmons is also right, in that Safari has supported Web Manifest for a very long time, even if it supports some things "partially". 

The requirement "Safari on iOS does not support A2HS in WebViews like Chrome and Firefox." seems completely arbitrary, because it's an implementation detail - and it's also incorrect: some may "WebViews" others use the browser itself (not a "WebView"). 

So, again, the "can I use?" question is really: does a user agent generally support the properties of the Web Manifest. To which is fair to say Safari mostly ("partially") does, with a few exceptions... as noted in the more specific manifest field requirements.  

Lastly, @atjn, respectfully, there is no requirement anywhere that there be an associated Service Worker with a Web App. That's a proprietary/arbitrary (and completely unnecessary, IMO) Chrome requirement. **It's not a requirement of any web standard**. As @jensimmons points out, the same can be said about the display modes.

atjn:
@marcoscaceres thanks for the feedback.

I think you misunderstood what I am trying to do. Please read my clarifications and tell me if/why you still disagree.

(WAM = Web App Manifest)
(I2D = Install to Device - IMO this table should be called I2D)

> Again, I respectfully ask that instead you focus on: "does the user agent support and make meaningful use of Web Manifest" (i.e., "can I use web manifest with browser X?").

> So, again, the "can I use?" question is really: does a user agent generally support the properties of the Web Manifest

We _can_ track general support for WAM if you think that is useful. In that case the I2D support table can be moved to a new table.
I just don't see why it would be beneficial to track WAM support. It is a constantly moving target, which means we would need 20+ support notes, making the table almost unreadable. You can check out [the support table on MDN](https://developer.mozilla.org/en-US/docs/Web/Manifest#browser_compatibility) to see how fragmented the support landscape is. Which of the WAM fields would you consider a requirement for "Meaningful use of WAM", and how would you communicate that distinction to the average caniuse reader? This task seems impossible to me, given how many unrelated use cases WAM can afford, and trying to do so would probably be evangelizing/codifying a specific use of WAM, which as you say, would be against the spirit and the intent of WAM.

I think it is much more useful that you can search for the specific WAM entries that you want to use, either by reading MDN, or by [searching for it on caniuse](https://caniuse.com/mdn-html_manifest_display). And if you are looking for a specific feature that uses parts of WAM (such as I2D), you can search specifically for it and get support details for only that use case.

> > We are only tracking support for that specific UI unique to certain operating systems.
> 
> This is not good: you are basically tracking "is it Chrome"? Please don't do that.

No we are not. Safari passes these requirements as well. Actually, I would add notes to Chrome that mention that it has additional requirements (site must register serviceworker, must have WAM), making it clear that Safari support is better (see https://github.com/Fyrd/caniuse/pull/6744#issuecomment-1592003343).

> > We are specifically tracking offline install and fullscreen launch as a combined feature here because they allow for the app to appear as though it is natively installed.
> 
> This is codifying one particular experience [...] and goes against the spirit and intent of the Web Manifest spec. Basically, what @jensimmons said above point_up_2. Please don't do that!

Correct, this is codifying an experience, but it also makes it very clear that it is codifying an experience. IMO the title should be `Install to device`, and the description should be preserved, making it very clear to the reader that it only tracks this narrowly defined install experience.

> > If you press a natively installed app icon, you would never end up in a browser. The intent is not to evangelize that this is the "correct" way for a website to work, we are just tracking which browsers would allow a website to work that way if it wants to.
> 
> This doesn't make any sense from a "can I use" perspective because that's a browser UI affordance: By this logic, why not then also have a "bookmarks" section to answer "can I bookmark pages"? Or a "History" section, to "can I look at my browsing history?" section.

This feature is massive for web apps because it allows them to be used in place of natively installed apps. This is also why both Chrome and Safari touted it as a major new feature on their *developer* conferences.  Support for this install experience can mean the difference between a developer choosing to build a website that redirects to a native app, or a developer choosing to build a PWA with an install flow. And understanding which browsers support this feature can take a lot of time. So if I am a developer, I want to be able to search for this feature, and find a neat overview of the support landscape that someone else made for me. Why should caniuse not provide that?

If a feature such as "bookmarks" or "history" could mean the difference between millions of developers writing a native app or a web app, I would support the inclusion of that feature into caniuse as well. Caniuse does sometimes track support for things that are not standardized or not strictly a "feature", but that developers still want to check support for. It is one of the things that make it unique from MDN; not better or worse, just different, and useful for different things.

> Worse, this is documenting arbitrary browser UI and has nothing to do with "can I use Web Feature X?"

To me, it is pretty well defined. What about this is arbitrary to you? It seems to be well defined enough that [Chrome has formal requirements](https://web.dev/install-criteria/) that define when the feature is allowed to be used. They also [add formal API's](https://developer.mozilla.org/en-US/docs/Web/API/BeforeInstallPromptEvent) to improve the install experience. These are mostly non-standard, but that does not mean we can't track support for them, and as long as there is no W3C standard, any other browser that implements similar tech will also be marked as supported, as exemplified by the fact that Safari is marked fully supported, and none of us have argued against that.

> As @jensimmons pointed out, the ability to "Add to Homescreen" is unrelated to any web standard (going back to iOS 1.0!). It just happens that Web Manifest assists with that by providing a bit of metadata, icons, etc. But web manifest is completely optional to adding to home screen and getting an "app like" experience.
>
> So, to echo what @jensimmons said above: if this is about "Add to home screen (A2HS)" then @jensimmons is absolutely right that iOS 1.0 was the first version to support "A2HS". If this is about "Web Manifest" support, then @jensimmons is also right, in that Safari has supported Web Manifest for a very long time, even if it supports some things "partially".
> 
> The requirement "Safari on iOS does not support A2HS in WebViews like Chrome and Firefox." seems completely arbitrary, because it's an implementation detail - and it's also incorrect: some may "WebViews" others use the browser itself (not a "WebView").

As explained in this reply and in https://github.com/Fyrd/caniuse/pull/6744#issuecomment-1592109051, we are not tracking support for A2HS, and we are not tracking general support for WAM.

> Lastly, @atjn, respectfully, there is no requirement anywhere that there be an associated Service Worker with a Web App. That's a proprietary/arbitrary (and completely unnecessary, IMO) Chrome requirement. **It's not a requirement of any web standard**. As @jensimmons points out, the same can be said about the display modes.

I agree, see https://github.com/Fyrd/caniuse/pull/6744#issuecomment-1592003343

marcoscaceres:
> (I2D = Install to Device - IMO this table should be called I2D)

I don't think this should be a thing, tbh.

> We can track general support for WAM if you think that is useful. 

I think that would be useful because it subsumes/includes "I2D". And the table could explain that by supporting the Web Manifest spec, a browser can support ability to install web applications, which can make them look and behave similarly to other applications installed on the OS. 

I acknowledge installability is a big deal (and the primary use case for web manifest). But again, as a developer myself, I'm primarily interested in "can I use" Web Manifest - and, specifically, what property members can I actually use? (some of which imply "installed", for example, the some display mode, orientation locking, shortcuts, etc.)

atjn:
> > We can track general support for WAM if you think that is useful.
> 
> I think that would be useful because it subsumes/includes "I2D". And the table could explain that by supporting the Web Manifest spec, a browser can support ability to install web applications, which can make them look and behave similarly to other applications installed on the OS.

If we were to include I2D support in the WAM support table, we would have to mark iOS Safari support partial because it does not allow third party browsers to install apps. This would not be fair, as Safari does support many parts of the WAM spec for third party browsers. On the other hand, if we mark Safari with full support for WAM, this makes the table unusable/harder to use for developers trying to figure out where I2D is supported.
This is just one out of many examples of why I think these should not be tracked together, and also why I think it is wrong to say that WAM subsumes I2D.
In order to not mislead the developers that read caniuse, we must have a clear distinction between when we are tracking the general WAM tech, and when we are tracking a specific install experience. Otherwise we would be codifying/evangelizing a specific feature, which you are all clearly against, and for good reason.

> I acknowledge installability is a big deal (and the primary use case for web manifest). But again, as a developer myself, I'm primarily interested in "can I use" Web Manifest - and, specifically, what property members can I actually use? (some of which imply "installed", for example, the some display mode, orientation locking, shortcuts, etc.)

I am not forcing you to ever look at the I2D support table, I am just humbly requesting that the developers who want to have it, can be allowed to have it. I think I have explained already why a developer would be very interested in knowing if I2D is supported. I would also like to mention that if any browser ever designs a different way of installing that is not I2D, then we would open a new support table and link the tables to each other in the description, much like we have [done with images](https://caniuse.com/avif). We would never freeze out a competitive idea, we would just let the developer decide what they want based on the support data and relevant information.

I would like to reiterate that caniuse already has support notes for every single WAM field, and as you mention yourself, they are very useful because they only track a single field each.
I have already noted that we _can_ track general support for WAM as well, but in that case you would have to explain to me where the threshold is for when a browser supports it, because as far as I can see, there exists no threshold that would result in information that is usable for a developer. If we don't want to codify/evangelize a specific use case for WAM, we would have to mark Chrome `fully supported`, while Safari and Firefox would just be `partial support` because they don't support every single field. And every time a new filed is added to the spec, all browsers revert to partial support again. Please explain to me how this information is useful.

atjn:
I think the only issue here could be that the table ID is `web-app-manifest`, and while it isn't shown anywhere in the UI, it could cause confusion.
@Fyrd would it be okay to rename the table ID? If not, and if @jensimmons and @marcoscaceres wants us to, I would support opening a new table for `install-to-device`, and reverting this table to tracking general WAM support, which IMO is still not useful information, it would just be necessary to keep around because we don't want to delete existing tables.

If we revert this table to general WAM support, I would suggest marking every browser that supports at least one field as `partial support` with a note that the web app spec is a very complex technology, and that the developer should search for a specific field/feature in order to gain more information. This would save us a lot of work and IMO impossible discussions about what constitutes "full support".

Fyrd:
Yeah, I think the best way forward would be a new support table for "install to device" (or some other name) and keeping this one for some definition of support for WAM:

Some ideas for that definition could be:
- Supports at least one field
- Supports a predefined set of fields
- Supports a certain snapshot of the specification

Assuming of course we feel any of those would seem helpful to developers. If it doesn't then I don't see much need in having it at all.

One other option is to not have a WAM support table but instead have just a search result that provides a little info and links to the individual field support tables. This is currently what I do on the site for ES modules, e.g. https://caniuse.com/sr_es13 Would that perhaps be more appropriate?

marcoscaceres:
> One other option is to not have a WAM support table but instead have just a search result that provides a little info and links to the individual field support tables. This is currently what I do on the site for ES modules, e.g. https://caniuse.com/sr_es13 Would that perhaps be more appropriate?

Oh! I like this more to be honest. I'd be supportive of that. @jensimmons, thoughts?

atjn:
I second @marcoscaceres, that sounds like the best solution :)

Fyrd:
Alright, for now I've hidden the web-app-manifest support feature from the site (may remove entirely later) and added a search result entry:
https://caniuse.com/sr_web-app-manifest

@atjn want to start a new PR for "I2D"? Then if there's further discussion needed on that subject we can continue it there.

Closing this PR now as I believe we've come to a resolution, though further discussion on the topic is certainly welcome. Thanks everyone for your contributions!

PR #6747, opened 2023-06-09
Updates to Referrer-Policy

atjn:
For most websites, the spec update that changed the default from `no-referrer-when-downgrade` to `strict-origin-when-cross-origin` is very important, because the new default means they can completely ignore the header. Therefore I think it makes sense to mark all browsers with the old default as `partial` support. 

I also updated the data to reflect that Safari and a bunch of smaller browsers have updated to the new default. It is hard to test this, so I am partially basing this on [data from MDN](https://caniuse.com/mdn-http_headers_referrer-policy_default_strict-origin-when-cross-origin) and information about the underlying engines being used.

I also made the note from #6608 visible, as it seemed like a mistake that it was not shown. I have not tested whether the note is correct.


Fyrd:
Looks good, thanks!

PR #6748, opened 2023-06-10
Disallow dangling notes

atjn:
I have noticed a few times that we sometimes have "hidden" notes that are not shown on caniuse because we forgot to add references to the note. It is hard to manually detect that this is happening because the tables can be very detailed.

I have added a check to the validator that disallows notes without a reference, and also disallows references to nonexistent notes. Now we can be sure that all notes we see in the `json` files are displayed on caniuse, and I think that is a big improvement. Let me know if you disagree :)

The check found a lot of of existing dangling notes. I have done my best to remove notes that seem like they were supposed to be removed, and to correctly link notes that seem like they were supposed to be shown.

The check fails because it needs the fix I made in #6747, otherwise it should be ready to be merged.


Fyrd:
This is excellent!! Thanks much for writing this, nice to keep things cleaned up. :)

PR #6756, opened 2023-06-15
Add partial support for AVIF in Edge Dev/Canary 

Th3S4mur41:
It doesn't work in Edge 116 (Dev/Canary) on Windows 10 21H2.
It works in Edge 116 (Dev/Canary) on Windows 11 Moment 3 (22621.1848). 

To be confirmed if Moment 3 is the key factor here?
It doesn't work in Edge 114 (Stable) on Windows 11 Moment 3 (22621.1848). 

Need some help with the validation here too, since Edge Canary is not part of the data model yet.

### DIFF
--- a/features-json/avif.json
+++ b/features-json/avif.json
@@ -79,9 +79,7 @@
       "111":"n",
       "112":"n",
       "113":"n",
-      "114":"n",
-      "115":"n",
-      "116":"a #6"
+      "114":"n"
     },
     "firefox":{
       "2":"n",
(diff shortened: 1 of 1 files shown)


2553BAG (approved):

Th3S4mur41:
@Fyrd could you help add the Missing Edge Beta/Canary versions to the data model?

Fyrd:
Support in Beta/Canary versions can be more in flux so any support data there is not as reliable as that for stable versions. As a result I'm only tracking stable Edge as most of the time support matches Chrome anyway. 

Would recommend instead to add the information under "Notes" instead, especially if there's uncertainly on the subject.

Th3S4mur41:
Thanks @Fyrd PR updated as suggested

Fyrd:
Thanks!

tats-u:
@Th3S4mur41 The note you added misleads us into believing that Edge 116 Stable added support for AVIF. You never heard that Edge Stable would add support for AVIF, did you? We have to modify the misleading note.

Th3S4mur41:
@tats-u you are right, avif support was enabled by default to Edge 116 in the canary channel, but it has not been ported to the stable channel yet.

So, all they did with 116 was remove the necessity to add the [`--enable-features=msEdgeAVIF` argument](https://winaero.com/avif-support-is-now-available-in-microsoft-edge/) to the msedge.exe shortcut.

One other requirement that I'm currently verifying, is that [it looks like avif support also requires to be on one of the Windows Insider channels](https://stackoverflow.com/questions/75459594/why-doesnt-edge-support-avif-images#comment135702302_76481551).

tats-u:
@Th3S4mur41 
> avif support was enabled by default to Edge 116 in the canary channel, 

No, since 114. The flag has been activated since 112 but the display feature was lacked at that time.
I have been monitoring the AVIF feature in Edge Insider (& stable) since then.

Issue #6056, opened 2021-10-15
Add Scoped Styles aka. CSS Scoping aka. @scope


Schweinepriester:
- <https://css-tricks.com/css-is-going-gosh-darned-hog-wild-i-tell-ya-what/#scoping>
- <https://drafts.csswg.org/css-cascade-6/#scoped-styles>
- <https://css-tricks.com/early-days-for-css-scoping/>
- <https://css.oddbird.net/scope/syntax/>

(Not the same as <https://caniuse.com/style-scoped>)

mirisuzanne:
I expect browsers to start prototyping this soon, but I don't think it's landed in any preview releases yet (if that's the criteria).

keithjgrant:
Chrome has a demo of this behind a flag now and it seems like it might be making forward progress again

matthew-dean:
Tracking bug is here: https://chromium-review.googlesource.com/c/chromium/src/+/4161741

Eduruiz:
I'm using chromium-dev on linux and it's already there :)

Schweinepriester:
- https://fullystacked.net/posts/scope-in-css/
- Shipping in Chrome 117: <https://chromestatus.com/feature/5100672734199808>
- Positive position for WebKit: <https://github.com/WebKit/standards-positions/issues/13>
- Firefox:
  - https://github.com/mozilla/standards-positions/issues/472
  - https://bugzilla.mozilla.org/show_bug.cgi?id=1830512

Fyrd:
Now available at https://caniuse.com/css-cascade-scope

yisibl:
It is expected to ship in Chrome 118. https://chromium-review.googlesource.com/c/chromium/src/+/4790561

bramus:
> It is expected to ship in Chrome 118. https://chromium-review.googlesource.com/c/chromium/src/+/4790561

That is correct, as per https://chromestatus.com/feature/5100672734199808

PR #6760, opened 2023-06-19
Add @scope (currently with incomplete chrome support data)

DarkWiiPlayer:
Add support information for the `@scope` rule.

Corresponding issue: #6056

Chrome status: https://chromestatus.com/feature/5100672734199808
Chrome intent to ship: https://groups.google.com/a/chromium.org/g/blink-dev/c/OEfGbd74QnQ/m/KaX-2hhRAAAJ
Spec: https://drafts.csswg.org/css-cascade-6/#scoped-styles

Support information for chrome is incomplete; `@scope` is enabled behind a flag at least as far as 114 (tested locally) and probably a bit earlier than that.

---

<sub>Side note: Hitting enter in a PR title edit box should *not* just create the new PR with an empty description, that's just terrible UX wth<sub>


Fyrd:
Thanks! I just did more testing on this, looks like flagged support was added all the way back in 104 though not seeing it enabled by default in the latest 117 canary.

yisibl:
It is expected to ship in Chrome 118. https://chromium-review.googlesource.com/c/chromium/src/+/4790561

PR #6764, opened 2023-06-22
Update css-display-contents.json

aardrian:
Added links to two more WebKit bugs to existing note; added more roles that are impacted to same existing note.


Fyrd:
Thanks!

PR #6765, opened 2023-06-22
Update link-rel-preconnect.json for Firefox fix in 115

acreskeyMoz:
We fixed rel=preconnect in Fx 115

https://bugzilla.mozilla.org/show_bug.cgi?id=1543990


Fyrd:
Thanks! Don't usually add a new note when a bug is fixed but I have updated the previous note for clarity.

acreskeyMoz:
Appreciated, thank you!

PR #6767, opened 2023-06-26
Fix typo for "Media Queries: Range Syntax"

Ayc0:
I noticed that some of the code was improperly placed in a inline code – [link](https://caniuse.com/css-media-range-syntax):

image


Fyrd:
Thanks!

PR #6770, opened 2023-06-28
Update webauthn to remove note 4 from firefox as support has been added

CADawg:
https://www.mozilla.org/en-US/firefox/114.0/releasenotes/


Fyrd:
Thanks!

Issue #6771, opened 2023-06-29
DOMMatrix data erroneously suggests iOS Safari doesn’t have support


controversial:
On the entry for [DOMMatrix](https://caniuse.com/dommatrix), caniuse data suggests that all iOS Safari since 5 only support the name WebKitCSSMatrix and not the current DOMMatrix name (“Only supports the WebKitCSSMatrix version of the interface, not DOMMatrix”)

The [MDN Data for the same feature](https://caniuse.com/mdn-api_dommatrix) correctly show that iOS Safari has support the DOMMatrix API since iOS 11.

PR #6773, opened 2023-06-30
DOMMatrix: Align Safari on iOS to Safari

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/6771

probably made it's way like this via https://github.com/Fyrd/caniuse/pull/3477 TP only => desktop only => iOS never updated.

Seeing https://caniuse.com/mdn-api_dommatrix really makes me doubt note 4, but oh well, at least more precise with this PR 0:-)


Fyrd:
Thanks!

I checked again and while SVGMatrix maybe _should_ be an alias it doesn't seem to be. For example in Chrome:
Screen Shot 2023-07-02 at 10 50 02 AM
(also https://tests.caniuse.com/dommatrix has a test for this)

PR #6784, opened 2023-07-13
Remove note 3 from Chrome/Edge 110+

HM100:
As according to [this page](https://en.wikipedia.org/wiki/Google_Chrome), there're zero Operating Systems supported by Chrome 110+ that do not support COLR font rendering.


Fyrd:
Thanks!

PR #6788, opened 2023-07-20
Edge Insider supports AVIF via runtime flag

tats-u:
https://www.neowin.net/news/microsoft-edge-is-finally-getting-avif-support-long-after-firefox-google-chrome-and-safari/

At one time Dev and Beta supported it, but I forget when they dropped it.

### DIFF
--- a/features-json/avif.json
+++ b/features-json/avif.json
@@ -79,8 +79,8 @@
       "111":"n",
       "112":"n",
       "113":"n",
-      "114":"n",
-      "115":"n"
+      "114":"n d #6",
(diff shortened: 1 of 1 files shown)


tats-u:
I may have to change: 
Canary only
↓
unstable builds (at least Canary)

tats-u:
114 (just before stable): including Beta & Dev
115: maybe in Canary
116: including Beta & Dev

tats-u:
@Fyrd I rebased to main.

Fyrd:
Looks good, thanks!

PR #6791, opened 2023-08-04
Update Safari srcset usage

bakura10:
Hi :),

Unfortunately, I figured support is still partial for sizes on Safari. Especially, the min/max expressions are not supported (work properly on Chrome and Firefox) which cause Safari to always use the largest image available when finding such expressions. I opened a bug report here: https://bugs.webkit.org/show_bug.cgi?id=259812


Fyrd:
Thanks!

PR #6803, opened 2023-08-17
Update AVIF iOS Info

BleachDev:
The [Safari 16.1 Release Notes](https://developer.apple.com/documentation/safari-release-notes/safari-16_1-release-notes) says:
- Added support for non-animated AVIF on macOS Ventura, and iPadOS 16.
- Added support for AVIF animated image sequences on macOS Ventura, iPadOS 16, and iOS 16.

Though i think they made a typo and swapped the animated and non-animated  entries up because i tested on iOS 16.6 and 17 and animated AVIF images appear to not be supported.


Fyrd:
Confirmed, thanks!

PR #6806, opened 2023-08-18
Opus/CAF: clarify that only CBR is supported

Artoria2e5:
VBR CAF can be produced by ffmpeg and theoretically is spec-compliant, but Safari (and macOS Quicktime Player) will refuse to decode. Unfortunately Opus defaults to (constrained) VBR because it works better that way!


Fyrd:
Thanks!

PR #6815, opened 2023-08-30
Add CSS Anchor Positioning feature

Th3S4mur41:
> The [CSS Working Group](https://www.w3.org/groups/wg/css/) has published a First Public Working Draft of [CSS Anchor Positioning](https://www.w3.org/TR/2023/WD-css-anchor-position-1-20230629/). This allows a positioned element (such as a tooltip or pop-up footnote) to size and position itself relative to one or more "anchor elements" elsewhere on the page. Because the anchoring element could have any size or position on the page, a flexible fallback scheme is also defined whereby a series of rules is tried until one is found that does not result in the positioned element overflowing its containing block.    
> 
> https://www.w3.org/news/2023/first-public-working-draft-css-anchor-positioning/

Support of the feature in Chrome/Edge Canary with `#enable-experimental-web-platform-features` flag in `chrome://flags` set to `true` verified with the CodePens in the following blog post
https://developer.chrome.com/blog/tether-elements-to-each-other-with-css-anchor-positioning/

According to [this comment](https://bugs.chromium.org/p/chromium/issues/detail?id=1309178#c176), the feature is flagged to be released in Chrome 117


Fyrd:
Thanks!

PR #6823, opened 2023-09-07
Support data update for Samsung Internet 22

nilaallj:
Samsung Internet 22 updated from Chromium 110 to 111. This PR updates support data accordingly:
- Support for CSS `color()` function
- Support for CSS `font-variant-alternates` property
- Support for CSS `lab()` and `lch()` functions
- Support for CSS nth-child-of
- Partial support for Container style queries (custom properties only)

Used [Chrome Platform Status](https://chromestatus.com/roadmap) as reference


Fyrd:
Thanks! Yeah I usually test and update when adding a version but missed this time so I appreciate the quick update!

PR #6824, opened 2023-09-08
Removing `display: contents` bug footnote 2 from iOS 16.1+

jensimmons:
Footnote 2 ("Buttons are not accessible with display: contents applied") describes a problem that was fixed in Safari 16.1 on both macOS and iOS. We just tested again to be sure.

The little '2' was dropped for Safari on macOS, but not Safari on iOS. This PR corrects that.


jensimmons:
We've been working hard on a large pile of improvements to `display: contents` for Safari 17.0. We hope we've fixed everything... but we welcome help testing and making sure once Safari 17.0 is out in the world. We are committed to fixing any remaining bugs.

We'll file another issue for Can I Use about `display: contents` once the time comes. Meanwhile, I wanted to clean up the data about the current state. Buttons are accessible with `display: contents` in Safari 16.1–16.6+ on iOS.

Fyrd:
Ah yes, that seems right.

And sounds great! :)

PR #6833, opened 2023-09-18
Safari 17.0 fixes bugs with display: contents

jensimmons:
Safari 17.0 fixes accessibility bugs with `display: contents`.

> Improvements in the experience of elements which are styled `display:contents` include:
> 
> * Making HTML tables and table components available to the accessibility tree.
> * Ensuring elements with the ARIA roles `grid`, `treegrid`, `table`, `row`, `gridcell`, `cell`, `columnheader`, `tree` and `treeitem` are properly exposed in the accessibility tree.
> * Ensuring `aria-checked` works for elements with the ARIA roles `tree` and `treeitem`. 
> * Returning the correct sub-role for lists, including elements with an ARIA role of `directory`.
> * Ensuring elements are not missing their children.
> * Making elements able to return selected accessibility children.
> * Computing the accessible name that rely on labels or captions.
> * Computing the correct accessibility clickpoint for links and headings.
> * Properly exposing lists that have `display:contents` list items.
> * Making `aria-checked` work for `role="treeitem"`.
> * Making `aria-activedescendant` work.
> * Making `aria-grabbed` and `aria-dropeffect` work.
> * Making `aria-flowto` work.
> * Making CSS `speak-as`, AXAccessKey, `aria-owns`, and URL AX APIs work.
> 
> Safari 17.0 also fixes accessibility issues with HTML tables and table components that have `display: flex`, `grid`, `block`, or `inline-block` applied.

https://webkit.org/blog/14445/webkit-features-in-safari-17-0/

This should be everything. We are happy to give this PR some time so folks can test. If you do find more bugs, please do file them at bugs.webkit.org, so we can address them.


Fyrd:
Great to hear, thanks!

PR #6836, opened 2023-09-22
Safari 17.0: AV1 & unprefixed `hyphens` 🎉 

Schweinepriester:
<https://webkit.org/blog/14445/webkit-features-in-safari-17-0/>

Finally got time to read the changelog, here are the two missing things I found:

> ### AV1
>
> Safari 17.0 adds support for AV1 video on devices with hardware decoding support, like [iPhone 15 Pro and iPhone 15 Pro Max](https://www.apple.com/newsroom/2023/09/apple-unveils-iphone-15-pro-and-iphone-15-pro-max/).

🥳

> ### Hyphenation
>
> Supported in Safari with `-webkit-` prefixes since 2011, the `hyphens` and `hyphenate-character` properties are now supported in their unprefixed form.

Have tested neither, but as `hyphens` was in TP before I have little doubt and I'm sure we'll adjust AV1 as we go 0:-)


Fyrd:
Awesome, thanks!

Issue #6827, opened 2023-09-11
Font-size-adjust needs update to new spec


jensimmons:
`font-size-adjust` was added to Can I Use years ago. Since then, two new details have been added to what this feature should do. In 2023, to fully support how `font-size-adjust` is suppose to work, a browser should have support for:

1. basic syntax
2. two-value syntax
3. `from-font` value

The page at https://caniuse.com/font-size-adjust only reflects the first, the basic syntax. 

There is a page coming from MDN BCD data: https://caniuse.com/mdn-css_properties_font-size-adjust_two-values which shows the support for two-value syntax. 

There's nothing available about `from-font` (which is supported in Safari 17.0, and not yet in Firefox). 

It seems like MDN BCD should have a separate data table for each of these three parts of the whole, where the details can be found about which browser added support for each when.

And that Can I Use should evolve the overall feature page to include how browsers are doing overall with `font-size-adjust` as a whole. Which would mean any browser that doesn't support all three parts gets "partial support" designation. 

Does that make sense to you all? If so, I can create the PR. But I wanted to discuss a plan first.

Fyrd:
Agree with all of these! Thanks for bringing this up, happy to receive a PR on the changes.

I see `from-font` is now on BCD as well and found on caniuse:
https://caniuse.com/mdn-css_properties_font-size-adjust_from-font

jensimmons:
Here's the PR: https://github.com/Fyrd/caniuse/pull/6841

jensimmons:
Done. Thanks!

PR #6841, opened 2023-09-25
Update `font-size-adjust` to reflect whole feature

jensimmons:
Updates discussed in [#6827](https://github.com/Fyrd/caniuse/issues/6827). 
- Marking which browsers are missing support for two-value syntax and `from-font` value.
- Marking browsers without all three parts of the feature as having partial support.


jensimmons:
Just added a link to issue for [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1219875).

Fyrd:
Looks good, thanks!

PR #6853, opened 2023-10-07
Update css-display-contents.json

aardrian:
* Confirmed bug with keyboard-inaccessible buttons persists in Safari 17 through TP 180 (Safari 17.4 beta).
* Updated the list of bugs to reference the WebKit bug reporting this from April 2023.


Fyrd:
Thanks!

Issue #6839, opened 2023-09-25
When Edge stopped support for AV1 videos in Windows?


tats-u:
Before the release of Windows 11, Edge used to be able to play AV1 videos by the help of the [AV1 Video Extension](https://apps.microsoft.com/store/detail/9MVZQVXJBQ9V). However, it doesn't seem to be able to play them even with that extension.

![image](https://github.com/Fyrd/caniuse/assets/12870451/d8e5271b-7859-4767-96f4-33a9249f2528)
↑2 MP4+AV1 videos are not played even if the play buttons are pressed.

![image](https://github.com/Fyrd/caniuse/assets/12870451/b522a60e-b37f-43a9-9654-9c320d800220)
↑c.f. WebKit for Windows (Playwright) supports the combination of AV1+AAC+MP4.

I don't know when Edge turned to be unable to play AV1 videos and I can't create a PR.
Can Edge in Windows _10_ play AV1 still now?

Test cases: https://tests.caniuse.com/av1

tats-u:
![image](https://github.com/Fyrd/caniuse/assets/12870451/fec6697e-d5c4-44bb-8fdf-eecb627008d7)

116 ❌

tats-u:
![image](https://github.com/Fyrd/caniuse/assets/12870451/255767db-dde9-4f6a-8790-faa0bfc2f4c6)

![image](https://github.com/Fyrd/caniuse/assets/12870451/9cbbe8a5-be10-4c95-b1ed-4c52057f557b)

Edge 121 Beta now resumes support for AV1 & AVIF, so I changed the title.

PR #6854, opened 2023-10-10
Edge has stopped supporting AV1 completely

tats-u:
#6839

Updated only with known information


Fyrd:
That's unfortunate, thanks for the update!

Issue #6858, opened 2023-10-12
H.265/HEVC on Firefox


tats-u:
120 Nightly + Windows + H/W decoding support (i7 1165G7)

![image](https://github.com/Fyrd/caniuse/assets/12870451/627504b5-8ac9-4fa0-9848-5a004087e01b)

- https://bugzilla.mozilla.org/show_bug.cgi?id=1853448
- https://bugzilla.mozilla.org/show_bug.cgi?id=1857097

tats-u:
It uses the same API as Edge.
Intel CPU requires 11th Gen or later. (Chrome can play with older CPUs thanks to another API)

tats-u:
You can enable it in stable builds since 121

![image](https://github.com/Fyrd/caniuse/assets/12870451/b95adefc-9507-4b9c-a719-0601ac60af30)

![image](https://github.com/Fyrd/caniuse/assets/12870451/e8541047-1887-423f-b0fb-8aa94b94bc4c)

tats-u:
In Windows, enabled by default since 134.

https://bugzilla.mozilla.org/show_bug.cgi?id=1928536

![Image](https://github.com/user-attachments/assets/1727c07b-5367-42d5-aa47-dbaac5f0756d)

10bit HEVC is playable in Windows since 132. (confirmed in 134)

https://codepen.io/danm8675123/pen/abNavqG
https://bugzilla.mozilla.org/show_bug.cgi?id=1866265

Playable in macOS since 136.

https://bugzilla.mozilla.org/show_bug.cgi?id=1924066

Not yet supported in Linux-based OSes.

PR #6861, opened 2023-10-15
Firefox 120 Nightly has started to support HEVC via OS API

tats-u:
#6858

It uses the same API as Edge as of now.

https://bugzilla.mozilla.org/show_bug.cgi?id=1853448
https://bugzilla.mozilla.org/show_bug.cgi?id=1859080
https://github.com/StaZhu/enable-chromium-hevc-hardware-decoding#whats-the-tech-diff-compared-with-edge--safari

### DIFF
--- a/features-json/hevc.json
+++ b/features-json/hevc.json
@@ -9,6 +9,10 @@
       "title":"Firefox support bug (WONTFIX)"
     },
     {
+      "url": "https://bugzilla.mozilla.org/show_bug.cgi?id=1842838",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

tats-u:
@Fyrd Nightly only but no flags required. Should we remove `d` from `n d`? Or should we set the status as "partially supported" instead?

Fyrd:
@tats-u Hmm.. I think "partially supported" is more appropriate, really.

PR #6869, opened 2023-10-24
WebKit status page has been retired, remove webkit_id

mrego:
The WebKit status page has been retired.

The website https://webkit.org/status/ shows now the following message:
> The WebKit Feature status page has been retired.

Pointing people to *MDN* or *Can I use?*, so this shouldn't point people back to the retired page.


mrego:
Note the issue to add support for that was still open #1808, maybe it can be closed after this is merged.

Fyrd:
Thanks for sharing! And disappointing, not but not surprising.

Note that the data here's generated from the site's database though so it's not just a matter of removing it in the repo. It's probably time to remove `ie_id` and `firefox_id` as well though since those have been defunct for a while. So I'll go ahead and take care of removing all three in the near future. Thanks again!

PR #6877, opened 2023-10-29
iOS bug with Screen Wake Lock API

asci:
on iOS it does not work in standalone (aka PWA) mode, only works in the actual browser. Tested on iOS 17. I think it is important to mention because it reduces possible use cases

### DIFF
--- a/features-json/wake-lock.json
+++ b/features-json/wake-lock.json
@@ -23,7 +23,7 @@
   ],
   "bugs":[
     {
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks! Maybe should refer to it as "web app" mode as that's how Apple [prefers to describe it](https://support.apple.com/en-us/HT213583)? Or PWA / "web app" as PWA is probably more common for developers.

asci:
@Fyrd  thanks, updated and added a link to the bug report

PR #6881, opened 2023-10-30
Firefox is enabling CSS `:has()` by default

Schweinepriester:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1858743#c2
- https://phabricator.services.mozilla.com/D192261

Which version is unclear at the moment, but surely will be made clear shortly!


Fyrd:
Thanks!

PR #6882, opened 2023-10-31
AV1 (surely) is also coming to Safari

Schweinepriester:
Per https://www.apple.com/apple-events/ and https://www.apple.com/macbook-pro/specs/ the newest devices will have 

> AV1 decode

capabilities.

image

image

Therefore I'm strongly assuming Safari will have AV1 support.
I selected 17.2 as starting point, assuming the new Macs will ship with that given the time frame.

Also adjusted the note a bit.

~~Naturally I have not tested it, so technically we could wait until somebody can test or confirm for us 0:-)~~

Confirmed below, afaik ready to go.

Fixes https://github.com/Fyrd/caniuse/issues/6884.


Schweinepriester:
@jensimmons Would you be able to confirm AV1 support for Safari on macOS running on M3 family hardware?

jensimmons:
Yes, AV1 support is coming to the new Macs. Support is in Safari 17.0+. The new MacBook Pros and iMacs are shipping in November.

features-json/av1.json
```diff
@@ -590,7 +590,7 @@
     "1":"Can be enabled in Edge on Windows 10 or later by installing the [AV1 Video Extension from the Microsoft Store](https://www.microsoft.com/store/productId/9MVZQVXJBQ9V)",
     "2":"Firefox 65 only supported AV1 on Windows 64-bit",
     "3":"Firefox 66 only supported AV1 on Windows and macOS",
-    "4":"Supported only for devices with hardware decoder i.e. A17 Pro (in e.g. iPhone 15 _Pro_ (Max)) or later"
+    "4":"Supported only for devices with hardware decoder, i.e. A17 Pro (e.g. iPhone 15 Pro) or M3 chip or later"
```
jensimmons, line 593: I'd suggest something like this instead: 

Supported on devices with hardware decoder, including iPhone 15 Pro, M3 iMac, M3 MacBook Pro, etc.

features-json/av1.json
```diff
@@ -370,8 +370,8 @@
       "16.6":"n",
       "17.0":"n #4",
       "17.1":"n #4",
```
jensimmons, line 372: "17.1":"a #4",

features-json/av1.json
```diff
@@ -370,8 +370,8 @@
       "16.6":"n",
       "17.0":"n #4",
```
jensimmons, line 371: "17.0":"a #4",

jensimmons (changes requested):

Elycin:
> Yes, AV1 support is coming to the new Macs. Support is in Safari 17.0+. The new MacBook Pros and iMacs are shipping in November.

@jensimmons, 

Forgive me, this may not be the exact place to ask but I figured it was worth a shot.

I just wanted to confirm -  are there any plans for software decoding support via a future release in Safari/WebKit? I understand AV1 being computationally intensive may have negative impacts on battery life, given the push for hardware accelerated decoding, but is there a plan for backwards compatible support for older devices? (eg: iPhone 14 Pro Max, M1 Max)

From a business standpoint, there is a user base out there that uses the stock browser on their lovely Apple devices, and companies going forward may be more inclined to use AV1 given the royalty-free option - which may unfortunately leave them in the dark. I worry about companies not being able to cater to these specific customers, outside of my own personal use case.

Thanks.

features-json/av1.json
Schweinepriester: Adjusted!

Fyrd:
Thanks!

Issue #6878, opened 2023-10-29
COLRv0 fonts do not appear to be supported in Safari


HM100:
As according to https://pixelambacht.nl/chromacheck/
![IMG_0482](https://github.com/Fyrd/caniuse/assets/30199227/845ed4e4-2fa8-425a-a761-53feab90748d)

rutopio:
It appears to be a bug. I have submitted a PR about it.

Bug report: COLR support vanished between Safari 16 and 17: [WebKit Bugzilla Bug 262223](https://bugs.webkit.org/show_bug.cgi?id=262223)

Schweinepriester:
@rutopio Thank you for linking that bug, I just couldn't find https://github.com/Fyrd/caniuse/issues/1018#issuecomment-1737111439 again 😅
Also mentioning it to have direct link here.

Schweinepriester:
I believe this can be closed as https://github.com/Fyrd/caniuse/pull/6883#event-10839326292 has been merged.

Issue #1018, opened 2015-05-12
Multicolor font support. This is a new OpenType feature that has four different implementations. Cur...


Fyrd:
Multicolor font support. This is a new OpenType feature that has four different implementations. Currently two of them are supported in Firefox, one in IE11 on Windows 8.1+.

See http://pixelambacht.nl/2014/multicolor-fonts/#testing-your-browser

---------
##### This issue was imported from [Google Moderator](https://www.google.com/moderator/)
Moderator votes: __+18__
Issue added by __Roel N__ on __2014-10-26__
To vote this issue up or down, simply include `+1` or `-1` in your comment.

pkuczynski:
Seems like only very few browsers and only on selected platforms/os support this type of fonts. There are also couple of proposals how to achieve those and its not in any way standardised.

* http://caniemoji.com
* https://code.google.com/p/chromium/issues/detail?id=333011
* https://bugzilla.mozilla.org/show_bug.cgi?id=889401

@Fyrd shall we add it or not?

valtlait:
+1

rotsee:
+1

svgeesus:
> Seems like only very few browsers and only on selected platforms/os support this type of fonts. 

True, but support is improving so this is a good candidate for caniuse. Currently Firefox supports SVG and COLR; Safari supports SBIX, Chrome supports nothing (for Webfonts; CBDT/CBLC is supported for installed fonts only) and Edge recently jumped from COLR/CPAL only (Edge <= 13) to supporting SBIX, SVG, and CBDT/CBLC as well with the Windows 10 update and Edge 14.

> There are also couple of proposals how to achieve those and its not in any way standardised.

That is incorrect. There are four ways to do this; three are standardized by ISO and are part of the OpenType v.3 specification. The fourth, SBIX, is undocumented by Apple but has been reverse engineered by third parties.

svgeesus:
Oh and +1

RoelN:
Shameless +1 even though I added this issue :)

tphinney:
+1

Writing this because I wanted to get the latest info from CanIUse, and it wasn't there.

The previously undocumented SBIX flavor was added to OpenType 1.8, launched in October 2016. All four flavors are now formally part of OpenType and standardized by ISO in the Open Font Format.

Ugoku:
+1

Jolg42:
+1 👍

wasd171:
+1

[...]

PR #6883, opened 2023-10-31
Update iOS 17+ support on COLR/CPAL v0

rutopio:
Reference: COLR support vanished between Safari 16 and 17: [WebKit Bugzilla Bug 262223](https://bugs.webkit.org/show_bug.cgi?id=262223)


Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/6878.

features-json/colr.json
```diff
@@ -508,9 +508,9 @@
       "16.4":"y",
       "16.5":"y",
       "16.6":"y",
-      "17.0":"y",
-      "17.1":"y",
-      "17.2":"y"
+      "17.0":"n #4",
+      "17.1":"n",
+      "17.2":"n"
```
Schweinepriester, line 513: I think it makes sense to put the note on all versions starting 17, assuming it's not intended to be removed.
If it turns out to be intended I guess just `n` it is.
But for now:

```suggestion
      "17.0":"n #4",
      "17.1":"n #4",
      "17.2":"n #4"
```

features-json/colr.json
```diff
@@ -590,6 +590,7 @@
     "1":"Supported only on Safari for macOS High Sierra (10.13)+",
     "2":"Supported only on Windows 8.1+",
     "3":"On Windows pre 8.1 and macOS Sierra (10.12) and below falls back to using FreeType: [Chromium bug 882844](https://bugs.chromium.org/p/chromium/issues/detail?id=882844#c3)"
+    "4":"COLR support vanished between Safari 16 and 17: [WebKit Bugzilla Bug 262223](https://bugs.webkit.org/show_bug.cgi?id=262223)"
```
Schweinepriester, line 593:
```suggestion
    "4":"COLRv0 support vanished with Safari 17: [WebKit Bugzilla Bug 262223](https://bugs.webkit.org/show_bug.cgi?id=262223)"
```

Schweinepriester (changes requested):
I have some suggestions - an additional one for which I can't create a code suggestion is to also change (desktop) Safari to `n #4` starting with 17.0 all the way to TP, please.

https://github.com/Fyrd/caniuse/issues/1018#issuecomment-1737111439 appears to show desktop (as well).

rutopio:
I am certain that iOS 17 does not support on my device. 
Currently, I am running macOS Monterey, and COLRv0 works perfectly on it. 
Need additional feedback from users with the latest macOS versions.

Schweinepriester:
Even with Safari 17!?
Funky, okay, then we might need a distinction.

Perhaps we can summon @t3chguy: On which version of macOS did you take the second screenshot in https://github.com/Fyrd/caniuse/issues/1018#issuecomment-1737111439? :)

rutopio:
Sorry for the confusion 😯, I meant that macOS Monterey with Safari 16, and it works perfectly. However, iOS 17 with Safari 17 fail.
That's very possible that both Desktop and Mobile Safari 17 do not support due to the latest Webkit kernel.
Perhaps @t3chguy can provide more details about the desktop side.

t3chguy:
image

Here you go

rutopio:
Thanks!
I've marked both Desktop and Mobile Safari versions above 17 as unsupported.

Fyrd:
Thanks!

features-json/colr.json
```diff
@@ -590,6 +590,7 @@
     "1":"Supported only on Safari for macOS High Sierra (10.13)+",
     "2":"Supported only on Windows 8.1+",
     "3":"On Windows pre 8.1 and macOS Sierra (10.12) and below falls back to using FreeType: [Chromium bug 882844](https://bugs.chromium.org/p/chromium/issues/detail?id=882844#c3)"
```
Fyrd, line 592:
```suggestion
    "3":"On Windows pre 8.1 and macOS Sierra (10.12) and below falls back to using FreeType: [Chromium bug 882844](https://bugs.chromium.org/p/chromium/issues/detail?id=882844#c3)",
```

PR #6886, opened 2023-11-03
Fix broken links to kangax.github.io/compat-table/

generalmimon:
Apparently, the https://github.com/kangax/compat-table repo has been transferred to https://github.com/compat-table/compat-table. Although the URL https://kangax.github.io/compat-table/ is redirected to https://compat-table.github.io/compat-table/, links like https://kangax.github.io/compat-table/es6/ or https://kangax.github.io/compat-table/es5/ are actually broken, so we need to update them.

### DIFF
--- a/features-json/es5.json
+++ b/features-json/es5.json
@@ -5,7 +5,7 @@
   "status":"other",
   "links":[
     {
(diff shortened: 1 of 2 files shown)


Fyrd:
Thanks!

PR #6887, opened 2023-11-04
Update firefox note in  jpegxl.json

shellheim:
Now jpegxl can be enabled in stable version of firefox also (not only in nightly) so the note should be updated.

### DIFF
--- a/features-json/jpegxl.json
+++ b/features-json/jpegxl.json
@@ -592,7 +592,7 @@
   "notes":"",
   "notes_by_num":{
     "1":"Can be enabled via the `enable-jxl` flag.",
-    "2":"Can be enabled via the `image.jxl.enabled` flag in `about:config` in Nightly only.",
+    "2":"Can be enabled via the `image.jxl.enabled` flag in `about:config`.",
     "3":"Can be enabled via the `--enable-features=JXL` runtime flag."
   },
   "usage_perc_y":3.17,


Fyrd:
Thanks!

tats-u:
@shellheim You told a lie.
No task to link libjxl even in stable builds is not linked to <https://bugzilla.mozilla.org/show_bug.cgi?id=1539075>.
You must present evidence if you object.

image
image

tats-u:
For the first place, you must not overwrite the note for past versions just because of the current one.
You must copy its description and create a new note instead.

shellheim:
@tats-u My bad. I saw that the flag `image.jxl.enabled` was added back in the stable channel. I didn't know that it was not yet functional. I was using a fork and tested on that with the flag enabled and that worked so I made a quick PR. I should have tested on the official browser. Sorry for the trouble.

shellheim:
are you considering adding that the libjxl hasn't yet been linked in stable but the option is still in `about:config` in your revert PR (probably as a comment?) for future contributors who see that?

tats-u:
@shellheim 

> I saw that the flag image.jxl.enabled was added back in the stable channel. 

This is a typical pitfall. This does not make any effect in stable builds.

>  I was using a fork

You should provide more info on that.

> the option is still in about:config

I will cover in that pitfall.

shellheim:
I was using librewolf. They used a patch to fix that and that wasn't in official firefox.

tats-u:
https://codeberg.org/librewolf/source/pulls/53
https://codeberg.org/librewolf/source/pulls/47

It was done by either of these PRs.
To raise another example, Edge is a Chromium fork but intentionally disables AV1 & AVIF. (This is because I hate Edge)

shellheim:
Yeah, I knew that. Sorry for the mistake and thanks for taking the time to find the PRs.

PR #6898, opened 2023-11-15
Update `background-clip: text` data and standardize it

yisibl:
1. background-clip: text has been added to the CSS specification, [status](https://www.w3.org/Style/CSS/current-work):
image
2. There are some bugs specific to the Blink and WebKit kernel browsers. here's a list of the common ones to prevent users from falling into a hole.

- https://bugs.chromium.org/p/chromium/issues/detail?id=1352821#c13
- https://bugs.chromium.org/p/chromium/issues/detail?id=1500148


yisibl:
Unadded bugs: 

- https://bugs.chromium.org/p/chromium/issues/detail?id=1450025

Fyrd:
Thanks! This is cool but until it actually appears in a Working Draft specification I think it should still keep the "unoff" status.

yisibl:
> Thanks! This is cool but until it actually appears in a Working Draft specification I think it should still keep the "unoff" status.

@Fyrd Done

Fyrd (approved):

PR #6903, opened 2023-11-17
Remove -moz-outline-radius note

pepelsbey:
The support for `-moz-outline-radius` properties [has been removed](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/88#css) in Firefox 88.


Fyrd:
Thanks! The "notes" entry is still required so I've restored it with empty string value.

PR #6904, opened 2023-11-21
loading=lazy bugs

hartman:
Updates for loading=lazy bugs

- Safari doesn't load lazy elements when you print
- Firefox requires the loading attribute to be defined before the src attribute
- The chrome mem leak got fixed


Fyrd:
Thanks!

PR #6909, opened 2023-11-26
Revert "Update firefox note in  jpegxl.json" (#6887) and mention pitfall that the flag has no effect in the other builds than Nightly

tats-u:
Reverts Fyrd/caniuse#6887

The flag `image.jxl.enabled` is ***still useless*** even in Firefox 120 stable.

image
image

No task to link libjxl even in stable builds is not linked to <https://bugzilla.mozilla.org/show_bug.cgi?id=1539075>.
This means that there have been no changes related to JPEG XL for stable builds.

### DIFF
--- a/features-json/jpegxl.json
+++ b/features-json/jpegxl.json
@@ -594,7 +594,7 @@
   "notes":"",
   "notes_by_num":{
     "1":"Can be enabled via the `enable-jxl` flag.",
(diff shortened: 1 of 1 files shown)


tats-u:
@Fyrd This is now ready for merged

Fyrd (approved):
Thanks!

PR #6915, opened 2023-12-04
COLR0 is fixed in STP 183 + Safari 17.2 (PR to replace 6888)

jensimmons:
Updated the status, and revised the footnote to be a bit more professional/objective.


Fyrd:
Thanks!

PR #6930, opened 2023-12-27
Firefox 121 stable can enable HEVC via flag

tats-u:
2nd phase of #6858 

image

image

I rolled back my Firefox to 120 once but could not play HEVC even with the flag.

10bit or higher colors are not supported.

https://bugzilla.mozilla.org/show_bug.cgi?id=1839107

https://codepen.io/danm8675123/pen/abNavqG

image


Fyrd:
Thanks!

Issue #6928, opened 2023-12-23
iOS Safari fullscreen API support 


lukewarlow:
Caniuse says that iOS Safari 17.2 now fully supports the fullscreen API. However it doesn't appear to be true (iPadOS only restriction still exists). It seemed to be in the betas but not make it to final release.

jensimmons:
This is correct. It was on in Safari 17.2 beta 1 for iOS, but was later put back into Preview (available behind a flag only). 

https://front-end.social/@jensimmons/111495353776714977

Support for Fullscreen API for iOS  is also not on by default in Safari 17.3 beta. We ask that people help beta test by flipping the flag. 

(Fullscreen is still supported on macOS and iPadOS, as it has been for years.)

PR #6875, opened 2023-10-28
Safari 17.2: Fullscreen API on iOS & relaxed CSS Nesting

Schweinepriester:
Via https://github.com/Fyrd/caniuse/issues/6873 & https://developer.apple.com/documentation/safari-release-notes/safari-17_2-release-notes

> Added support for the Fullscreen API on iOS. (58946263)

> Added new relaxed parsing behavior for CSS Nesting. (113475843)

Fixes https://github.com/Fyrd/caniuse/issues/6873.

I think those two are actually all that are required for caniuse-entries 🤔

PR #6931, opened 2023-12-27
Fullscreen API not yet enabled by default for Safari on iOS

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/6928, all info there.

Partial revert of https://github.com/Fyrd/caniuse/pull/6875.

---

Side note to you, Fyrd:
The spec at https://fullscreen.spec.whatwg.org/ calls it `Fullscreen API` as opposed to the entry here with a space as `Full Screen API` - should we align that? 0:-)


Fyrd:
Thanks! And yes, I'm good with renaming it to Fullscreen API. 👍

PR #6924, opened 2023-12-19
Add Edge support for AVIF

captainbrosset:
Edge 121 will have support for AVIF.

PR #6936, opened 2024-01-05
Edge support for avif

slyi:
Edge added avif support in build 121


Schweinepriester:
Duplicate of https://github.com/Fyrd/caniuse/pull/6924 unfortunately I would say, but it seems you have noticed already - thank you for the effort and caring! :)

PR #6939, opened 2024-01-08
Edge 121 (Beta) restores AV1 support

### LINKED ISSUE #6932: Edge 121 (Beta) restores AVIF and AV1 display features without runtime flags
AVIF: #6924 
AV1: #6939

Related: #6839 

image

image

Works on Windows Sandbox & 9th Gen Intel Processors (without hardware decoders), too. This means software decoders built in the plain Chromium are not omitted unlike before.

tats-u:
Closes #6932

Quoted from #6932's description:

> image
>
> image
>
> Works on Windows Sandbox & 9th Gen Intel Processors (without hardware decoders), too. This means software decoders built in the plain Chromium are not omitted unlike before.

As the author of #6924 is a member of the Edge team (https://github.com/Fyrd/caniuse/pull/6924#issuecomment-1876698592), the stable Edge will not be likely to disable AV1 again.


tats-u:
@captainbrosset Would you mind telling us whether AV1 (not AVIF) is going to be also officially available in Edge 121 stable?

captainbrosset:
> @captainbrosset Would you mind telling us whether AV1 (not AVIF) is also going to be officially available in Edge 121 stable, or reviewing the content of this PR?

Yes, both AV1 and AVIF are going to be available in Edge 121.

tats-u:
@captainbrosset Thank you for providing the precious information. Sorry for bothering you again, but I would appreciate if you could providing information on #6839. (Edge had a period of time when it dropped the support for AV1 playback, but it has not been strictly identified. This is why the long note "Edge has stopped supporting AV1 completely at some point prior to version 116 (additional information required)." has remained in [the AV1 page](https://caniuse.com/av1).)

captainbrosset:
By the way, Edge release notes are now up to date with the AV1/AVIF support in 121: https://learn.microsoft.com/en-us/deployedge/microsoft-edge-relnote-beta-channel#version-1210227749-january-8-2024

My understanding of the AV1/AVIF situation is that both formats were never supported in Microsoft Edge until now with version 121.
If support existed for a while, it should be considered more like a bug than anything else. Edge has not been able to add support for these formats in the browser until now for license issues. See https://toot.cafe/@slightlyoff/109899372183448386 and https://stackoverflow.com/questions/75459594/why-doesnt-edge-support-avif-images.

I would advise removing the long note that appears before 116.

tats-u:
> By the way, Edge release notes are now up to date with the AV1/AVIF support in 121

That is good news for us.

> My understanding of the AV1/AVIF situation is that both formats were never supported in Microsoft Edge until now with version 121.

It should have supported AV1 via Media Foundation, and used to supporte AVIF by the **runtime flag named `msEdgeAVIF`**. I wonder what that flag means. Possibly an Easter Egg.

> I would advise removing the long note that appears before 116.

It may be now futile to seek excessive accuracy of past information. I appreciate your opinion.

Fyrd:
Thanks!

PR #6942, opened 2024-01-11
Indicate buggy Safari behaviour for srcset

gdubicki:
as explained in https://github.com/vuejs/core/issues/6391.

Safari bug tracker: https://bugs.webkit.org/show_bug.cgi?id=190031


Fyrd:
Thanks!

PR #6943, opened 2024-01-16
Firefox is intending to ship content-visibility

Schweinepriester:
https://bugzilla.mozilla.org/show_bug.cgi?id=1874874

> Enable content-visibility by default

> I'm proposing next Monday as a target date, so that it happens after the soft code freeze for Firefox 123

also: https://groups.google.com/a/mozilla.org/g/dev-platform/c/kXp-yUvkNKQ/m/mDVXyyrtAAAJ?pli=1


Fyrd:
Thanks!

PR #5875, opened 2021-05-01
Update Safari Vorbis support

kiding:
Related to #5873, Safari 14.1+ on macOS 11.3+ added Vorbis audio format support, without the .ogg support. 

Tested with `<audio controls src="https://upload.wikimedia.org/wikipedia/commons/2/22/Volcano_Lava_Sample.webm"></audio>`.

#### References
* https://webkit.org/blog/11648/new-webkit-features-in-safari-14-1/

PR #6946, opened 2024-01-18
Vorbis is fully supported since Safari 16.0

jensimmons:
Ogg Vorbis is fully supported on macOS 11.3 Big Sur or later. Safari 16.0 also requires Big Sur or later. Therefore, Ogg Vorbis should be marked "fully supported" since Safari 16.0. I also updated the note to be clearer about support.

### DIFF
--- a/features-json/ogg-vorbis.json
+++ b/features-json/ogg-vorbis.json
@@ -353,18 +353,18 @@
       "15.4":"a #1",
       "15.5":"a #1",
       "15.6":"a #1",
-      "16.0":"a #1",
-      "16.1":"a #1",
-      "16.2":"a #1",
-      "16.3":"a #1",
-      "16.4":"a #1",
-      "16.5":"a #1",
-      "16.6":"a #1",
-      "17.0":"a #1",
-      "17.1":"a #1",
-      "17.2":"a #1",
-      "17.3":"a #1",
-      "TP":"a #1"
+      "16.0":"y",
+      "16.1":"y",
+      "16.2":"y",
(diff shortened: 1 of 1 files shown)


jensimmons:
Any clue as to why the tests failed?

Fyrd:
> Any clue as to why the tests failed?

Looks like there's an uppercase `Y` that's causing the failure.

I notice the description change removes the part in the description that mentions that the ogg container format is not supported, which I imagine was at least the case when support was added (see #5875). If support was added for it since then do we know in which version?

jensimmons:
I fixed the "Y". And look a deeper look at the note. I'm not sure what the answer is to your question, so I reverted the note back to its current state. We can update that later if needed. Meanwhile, full support was added in Safari 16.0 and the PR now corrects that.

Fyrd:
I just tested at https://tests.caniuse.com/?feat=ogg-vorbis using Safari 16.4 and it didn't seem to work. Also tried files from https://getsamplefiles.com/sample-audio-files/ogg and they failed as well. Is there a test you know of that works?

jensimmons:
What I understand now is that Ogg and Vorbis are two different features. 
Vorbis is an audio codec that is supported in Safari on Mac. 
And Ogg is a media container format that is not currently supported in Safari. 

One option would be to split this page into two features, since they aren't necessarily required to go together. The Vorbis codec, for example is used in WebM containers.

I created a different PR with the assumption that the page will remain as is, one page for two features, updating Safari's support for Vorbis with newer information.

PR #6948, opened 2024-01-19
Add description to css-stick

reece00:
If the parent element has the properties position: absolute and overflow: visible, and the parent element's parent element has the property overflow: hidden, the relative position seems to become relative to the top-level window.
[Sticky test html.zip](https://github.com/Fyrd/caniuse/files/13988999/Sticky.test.html.zip)

 Google Chrome 120.0.6099.225

Safari on iOS 17.2.1

### DIFF
--- a/features-json/css-sticky.json
+++ b/features-json/css-sticky.json
@@ -35,13 +35,7 @@
     },
     {
       "description":"A parent with overflow set to `auto` will prevent `position: sticky` from working in Safari"
-    },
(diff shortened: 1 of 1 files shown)


Fyrd:
Merging this for the description change only, I feel the bug description is a bit too specific and I notice it doesn't have a bug link. If it's been filed somewhere feel free to re-include. Thanks!

PR #6950, opened 2024-01-24
Updating footnotes for indexeddb.json

jensimmons:
I was struck today when looking up IndexDB that some of the language in the footnotes is quite emotional. This PR contains updates to simply make them a bit more factual and consistent. (No change to the information itself.)


Fyrd:
Hm, I wouldn't describe the language as emotional, I think the words were intended to be descriptive ("complete" and "never" just referring to absolute states vs partial). But I'm fine with these changes to avoid the perception.

PR #6957, opened 2024-02-01
Update css-scrollbar.json as I found one interesting change in Chrome 121.0.0

HM100:
In Chrome 121.0.0 onwards, the webkit-way of styling scrollbars is gone (At least when the standarized variants are used), and instead, the standarized scrollbar-color and scrollbar-width are supported. See https://chromestatus.com/feature/5665308343795712 for more
image
A screenshot is given above for evidence/proof of this change. Disregard the message about Chrome is updating


HM100:
Tested Edge as well, and it had the same change

HM100:
In my tests, on Websites that use both the Webkit and the Standard way of styling scrollbars, the latter is prioritized.

Fyrd:
Cool, thanks!

PR #6964, opened 2024-02-06
Chrome is enabling zstd support by default

paolobarbolini:
It looks like Chrome 123 is planning on enabling zstd support by default. https://chromestatus.com/feature/6186023867908096

I've enabled it on the Desktop side only for now. Let me know if there are any changes I should make.


Schweinepriester (approved):

Fyrd:
Thanks!

Issue #6963, opened 2024-02-06
Server-timing information for Safari is wrong


kamtschatka:
When you look at https://caniuse.com/?search=server-timing, for Safari it says: "Partial support". "Supports the Server Timing reporting in the Web Inspector Network tab, but not the JavaScript API".

I have a Safari 16.6 available to me, where I was able to verify that it is actually already available:
![image](https://github.com/Fyrd/caniuse/assets/22147234/fb33e1fd-021c-4cb7-a7cc-74c3a9a898de)

According to https://developer.mozilla.org/en-US/docs/Web/API/PerformanceServerTiming, this should be available on Safari with version 16.4, but I am not able to verify that.

PR #6965, opened 2024-02-06
Adjust Server Timing for Safari

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/6963.

As mentioned there^ and visible when searching like https://caniuse.com/?search=server-timing, MDN has full support listed since 16.4: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceServerTiming#browser_compatibility

Only the HTTP-header appears to be unclear:
https://caniuse.com/mdn-http_headers_server-timing => https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing#browser_compatibility

https://wpt.fyi/results/server-timing?label=master&label=experimental&aligned= seems close enough for full support

image

A quick test on https://tests.caniuse.com/server-timing also seems promising:

302771201-5a0b160e-5a4a-4b2d-afea-aab5d930567e


Fyrd:
Thanks!

PR #6966, opened 2024-02-07
@scope rule is defined in an official working draft spec

mirisuzanne:
It would also be great if a search for 'scope' would produce this as the top result, instead of the deprecated 2014 feature. Is that possible?


Fyrd:
Thanks! Yes, let me see what I can do about the ordering.

Fyrd:
Okay, https://caniuse.com/css-cascade-scope comes in first now, I also renamed the deprecated feature "Scoped attribute" to help clarify the difference.

PR #6974, opened 2024-02-15
Safari on iOS 17.4 will lose PWA capabilities in the EU

Schweinepriester:
- https://mastodon.social/@owa/111934184525302106
- https://open-web-advocacy.org/blog/apple-on-course-to-break-all-web-apps-in-eu-within-20-days/
- https://bugs.webkit.org/show_bug.cgi?id=268643
- https://twitter.com/firt/status/1757704444991426812
- https://mastodon.social/@tomayac@toot.cafe/111845185569144991
- EDIT 2024-02-26: https://infrequently.org/2024/02/home-screen-advantage/

~~Draft because things can still change, but to~~ start the discussion about which entries and how to mark it. I feel like a region-locked or region-excluded feature warrants no support `n` with a note as a warning to devs. The alternative would be only partial `a` of course.

What do you think, Fyrd?

~~I'd love for this PR to be obsolete!
Any comments @jensimmons?~~


Schweinepriester:
there we have it: https://techcrunch.com/2024/02/15/apple-confirms-its-breaking-iphone-web-apps-in-the-eu-on-purpose/

> Well, it turns out it’s not a bug that broke iPhone web apps, also known as Progressive Web Apps (PWAs), in the EU. Following developer complaints and press reports about how PWAs were no longer functional in the EU after installing the most recent iOS betas, Apple has updated its website to explain why. No surprise, the tech giant is blaming the new EU regulation, the Digital Markets Act, for the change, saying that the complexities involved with the DMA’s requirement to allow different browser engines is the root cause.

Schweinepriester:
For e.g. https://caniuse.com/mdn-api_notification_badge I asked over there: https://github.com/mdn/browser-compat-data/issues/22241

Schweinepriester:
<https://twitter.com/WebReflection/status/1758232341267333456>:

> I mean … even MDN or CanIUse now should have tables showing in which country a Web feature is enabled … this is madness 100%

Fyrd:
Very unfortunate indeed. However I do think "partial support" is still the most accurate description. Though while geo-limited like this it would certainly not ever reach full support.

May be worth linking to an article with explanation in the note.

Schweinepriester:
Mkay, adjusted.

jensimmons:
Suggested revision to the note:

> "7":"Not available in the European Union. See [explanation](https://developer.apple.com/support/dma-and-apps-in-the-eu/#:~:text=Why%20don%27t%20users%20in%20the%20EU%20have%20access%20to%20Home%20Screen%20web%20apps%3F)."

Schweinepriester:
Changing the wording regarding "only in the EU" or "only outside" or something is fine with me, but I strongly believe there should not just the stance of Apple be linked, but also an article contextualizing it.

jensimmons:
There are _many_ articles in the press about this. How are you deciding which one you want to link to?

Schweinepriester:
https://techcrunch.com/2024/02/15/apple-confirms-its-breaking-iphone-web-apps-in-the-eu-on-purpose/ was simply the first I personally encountered. In the end I would leave that decision to Fyrd.

Fyrd:
On further consideration let's just stick to the facts of the "what" rather than the "why". Have reverted back to the original line without links. Thanks for the input!

PR #6977, opened 2024-02-19
Add Chrome, Edge and Opera partial support to css-subgrid.json

Baptistou:
See issue [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1473242&no_tracker_redirect=1) and [here](https://issues.chromium.org/issues/40278914).
CSS subgrid does not work on nested fieldset elements on Chromium.
Issue reproduced today on Chrome, Edge and Opera latest version.

### DIFF
--- a/features-json/css-subgrid.json
+++ b/features-json/css-subgrid.json
@@ -83,14 +83,14 @@
       "111":"n",
       "112":"n",
       "113":"n",
-      "114":"n d #1",
-      "115":"n d #1",
-      "116":"n d #1",
-      "117":"y",
-      "118":"y",
-      "119":"y",
-      "120":"y",
-      "121":"y"
+      "114":"n d #1 #2",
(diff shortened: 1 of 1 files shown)


Fyrd:
Hmm, this seems like it's too specific to warrant partial support, maybe should just be added to the bugs list?

Baptistou:
> Hmm, this seems like it's too specific to warrant partial support, maybe should just be added to the bugs list?

If I compare with `:disabled` attribute for example, I see that IE has [partial support on fieldset](https://caniuse.com/fieldset-disabled) --> [fieldset-disabled.json](https://github.com/Fyrd/caniuse/blob/main/features-json/fieldset-disabled.json).
So we could either modify css-subgrid.json to add partial support on Chromium-based browsers or create a new file `fieldset-subgrid.json` as above. What do you think?

PR #6981, opened 2024-02-22
Update audio.json - iOS is only partial support not a bug - volume control

shalanah:
- The ability to change volume in iOS programmatically using Audio Element (being read-only) isn't a bug on iOS. It's a choice to not fully implement the API. 
- Desktop Safari does support volume control
- Double checked their native iMusic app and they allow for volume change there without having to use physical buttons.


shalanah:
This affects a lot of sites 
- Youtube sound is either on/off (really odd web behavior forced by iOS)
- Volume = 0 doesn't work (must use mute)
- And think about this site where you should be able to mix and match sounds https://www.noisli.com/ completely broken in iOS

Fyrd:
Agree, this seems like a pretty serious deficiency. Thanks!

PR #6986, opened 2024-02-27
Updating Safari's support of Ogg / Vorbis

jensimmons:
This changes a few things:
1) Separates the note about macOS version support from Ogg/Vorbis support, so that the expiration of the macOS requirement can be probably noted. 
2) Adds support for Vorbis to Safari 17.4 on iOS 17.4.

See release note: https://developer.apple.com/documentation/safari-release-notes/safari-17_4-release-notes
> Added support for the Vorbis audio codec on iOS. 
(Which applies to iPadOS and visionOS as well.)


Schweinepriester:
Good changes and happy to see support increasing!

One questions though:
Is it possible to install some or all versions of Safari 16 on macOS Big Sur below version 11.3?
I don't know for certain, but I think it's likely possible, right?
If it is possible, I think all versions for which it's true, should keep note 1 (i.e. Safari 16.x).

jensimmons:
Vorbis support arrived in Safari 14.1 with macOS Big Sur 11.3.

If a user decided to keep running macOS Big Sur 11.3 and never update — but they did know how to update Safari alone, yes, they could update to Safari 15 and Safari 16. 

I'll update the PR.

PR #6995, opened 2024-03-04
Update jpegxl Chromium support bug

swrobel:
The previous one was closed, but is linked to this newer issue


Fyrd:
I think it makes most sense to link to the jpeg xl bug, not one about the ticket itself.

PR #6996, opened 2024-03-04
Safari 17.4 supports @scope

jensimmons:
Safari 17.4 supports CSS Scoping — both @scope (and the expanded purpose of :scope).

https://developer.apple.com/documentation/safari-release-notes/safari-17_4-release-notes


Fyrd:
Thanks!

PR #7003, opened 2024-03-09
Chrome 124: WebSQL flag on desktop expired

bershanskiy:
Chrome 124 on desktop removed WebSQL entirely ([list of sources](https://groups.google.com/a/chromium.org/g/blink-dev/c/fWYb6evVA-w/m/E-xuk__vAAAJ)).


Fyrd:
Thanks!

PR #7006, opened 2024-03-15
Note which `as` values are supported for <link rel=preload>

tunetheweb:
Browsers do not support all `as` types in `<link rel=preload>`.

For Chromium see here: https://chromium.googlesource.com/chromium/src.git/+/refs/heads/main/third_party/blink/renderer/core/loader/preload_helper.cc#176

And here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#browser_compatibility
image

However further testing browsers show ALL BROWSERS all only support a subset of values:
https://resource-loading-examples.glitch.me/preload-types.html

So made this a general comment at the top, rather than a flag for every browser.


tunetheweb:
Have asked for the HTML spec to be updated: https://github.com/whatwg/html/issues/8332#issuecomment-2001980048

And will also look at updating MDN which is also currently wrong.

Fyrd:
Thanks!

PR #7010, opened 2024-03-20
Update fetch.json

ItamarGronich:
Android chrome supports fetch from version 42. Same as desktop chrome
Same for firefox mobile (the same as its desktop versions)


Fyrd:
Only the latest version of these browsers is currently tracked on caniuse, you can't add more via PR, sorry.

ItamarGronich:
@Fyrd 
What do you mean?
why is mobile not tracked?

PR #7013, opened 2024-03-26
Clarify firefox support for css :is()

carlosjeurissen:
After testing :-moz-any in Firefox 4.0, 3.6 and 3.0. The following is concluded:
support for :-moz-any in firefox v4.0 is already correct
:moz-any in firefox 3.6 and 3.0 is unsupported and has been assumed to also be unsupported in v2.0


Fyrd:
Thanks!

PR #7016, opened 2024-04-01
jpegxl.json: macOS/iOS Safari does not support animated JXL yet

simnalamburt:
You can check this with the following image using the Safari browser: [Link of animated JXL sample](https://jpegxl.info/test-page/anim-icos.jxl)

It should be animated like this  but it won't.

###### References
- https://jpegxl.info/test-page/


Fyrd:
Confirmed, thanks!

PR #7018, opened 2024-04-01
PNG no longer a WD

svgeesus:
PNG has been Candidate Recommendation since 21 September 2023

https://www.w3.org/TR/2023/CR-png-3-20230921/


Fyrd:
Thanks!

PR #7021, opened 2024-04-04
Text Wrap Balance support is now two properties — only supporting one should be marked "partial"

jensimmons:
When `text-wrap: balance` was first created, there was just one property, `text-wrap`. Slightly later, the CSS Working Group also created `text-wrap-style` as they defined how `text-wrap` and `white-space` should interplay with each other to handle wrapping. 

This means that the whole feature of Text Wrap Balance now includes `text-wrap: balance` _and_ `text-wrap-style: balance`. The Chromium-based browsers just have support for the shorthand, not the longhand.

You can see this in the WPT test results for Interop 2024: https://wpt.fyi/results/css/css-text/parsing?label=experimental&label=master&aligned&view=interop&q=label%3Ainterop-2024-text-wrap

This PR marks the Chromium browsers as "partial support" with a footnote to explain.


Fyrd:
Thanks!

PR #6937, opened 2024-01-07
Safari on iOS has supported Animated AVIF since v16.1

o-t-w:
https://twitter.com/jensimmons/status/1584608635174686720?lang=en

PR #7036, opened 2024-04-16
Animated avif is supported since Safari 16.1 on iOS and iPad OS.

yashikota:
Animated avif is supported since Safari 16.1 on iOS and iPad OS.   
https://developer.apple.com/documentation/safari-release-notes/safari-16_1-release-notes#New-Features


Fyrd:
Not actually the case, see: https://github.com/Fyrd/caniuse/pull/6937#issuecomment-1900986502
If you can find a test case that proves otherwise, please share!

PR #7046, opened 2024-04-29
Mark Pointer Lock API as prefixed until Firefox 50

foolip:
`document.body.requestPointerLock` and `document.body.mozRequestPointerLock` were evaluated in devtools in Firefox 49 and 50 to confirm that the change happened in Firefox 50.

This is also what the label on the tracking bug suggests: https://bugzilla.mozilla.org/show_bug.cgi?id=991899


Fyrd:
Thanks!

PR #7055, opened 2024-05-10
feat(ch-unit): remove links

MichaelDeBoey:
Both links give a 404


Fyrd:
Fixed the first link, it'd simply moved, replaced the second link with [this one](https://meyerweb.com/eric/thoughts/2018/06/28/what-is-the-css-ch-unit/).

But thanks for bringing this up!

PR #7057, opened 2024-05-16
Chrome is shipping `font-size-adjust` with the two-value syntax 🥳

Schweinepriester:
- https://botsin.space/@intenttoship/112451152841836606
- https://groups.google.com/a/chromium.org/g/blink-dev/c/l2CiHZ8BArs/m/FPQcJCWQAwAJ
- https://chromestatus.com/feature/5720910061371392
- https://chromestatus.com/feature/5170570175447040

> Estimated milestones
Shipping on desktop
127
Shipping on Android
127
Shipping on WebView
127

> Is this feature fully tested by [web-platform-tests](https://chromium.googlesource.com/chromium/src/+/main/docs/testing/web_platform_tests.md)?
Yes
https://wpt.fyi/results/css/css-fonts?label=master&label=experimental&aligned&view=interop&q=label%3Ainterop-2024-font-size-adjust

Not sure if we should update the chrome id from https://chromestatus.com/feature/5720910061371392 to https://chromestatus.com/feature/5170570175447040? 🤔
Or add the latter as an additional link?


Fyrd:
Thanks! We can actually use both IDs, just need to separate with comma. I'll add it.

PR #7062, opened 2024-05-21
Update early Chrome support for WebVTT

foolip:
It was behind a flag in Chrome 18:
https://www.w3.org/community/texttracks/2012/08/23/webvtt-support-in-browsers/

But it was enabled by default together with `<track>` in Chrome 23: https://chromestatus.com/feature/6719115557339136

https://wpt.live/webvtt/rendering/cues-with-video/processing-model/basic.html was tested in Chrome 22 and 23 to confirm.

This contemporary presentation also supports Chrome 23 as the correct version: http://html5videoguide.net/presentations/WebDirSouth2012/


Fyrd:
Thanks!

PR #7068, opened 2024-05-24
Update css-scrollbar.json for Opera browser

Douglas-2024:
Chrome/Edge browsers since version 121 support the three properties "scrollbar-color", "scrollbar-width" and "scrollbar-gutter", therefore they are marked in green as fully supported.

The Opera browser since version 107 supports these three properties, so it should be marked in green as fully supported as well.

See:
https://caniuse.com/?search=scrollbar-color
https://caniuse.com/?search=scrollbar-width
https://caniuse.com/?search=scrollbar-gutter


Fyrd:
Thanks!

PR #7083, opened 2024-06-12
Add X25519Kyber768 json data

### LINKED ISSUE #7072: X25519Kyber768 Hybrid Post-Quantum Key Exchange for HTTPS
Hybrid key exchange in SSL handshake using ML-KEM in addition to default already has significant usage thanks to Google and Cloudflare. It appears to only be supported currently in TLS 1.3 but is intended for TLS 1.2 as well.

olokelo:
X25519Kyber768 is an IETF Internet Draft designed to bring post-quantum asymmetric cryptography to TLS.
It has already been implemented in Chrome, Edge and Firefox for a while but since version 124 it's enabled by default in Chrome and Edge.

[Chromium feature link which includes deployment plan](https://chromestatus.com/feature/5257822742249472)
[Cloudflare Research site dedicated to post-quantum cryptography](https://pq.cloudflareresearch.com/)
[IETF draft status](https://datatracker.ietf.org/doc/draft-tls-westerbaan-xyber768d00/)

I made some tests on Browserstack and it seems like it's also behind a flag starting from Opera 101.
It's supported but not enabled by default on Chrome, Android.
There's no mention of support on Apple platforms nor any other less-popular independent browsers.

Closes #7072.


jannispinter:
I suggest not to implement this PR because it is based on data from a year ago and uses the pre-standard `X25519Kyber786` scheme, which is deprecated already and has been removed by most browsers. I have opened PR #7326 with current data based on the standardized variant `X25519MLKEM768`.

PR #7092, opened 2024-06-14
Appcache support was removed in Safari 15.0

jensimmons:
The final bits of appcache support was removed from WKWebView in iOS 18 beta (and on other platforms). We wrote all about the story of  when and how we've removed appcache step by step over 7 years.  

> **Storage**
> 
> WebKit for Safari 18 beta removes the last bits of support for AppCache.
> 
> When [AppCache](https://www.w3.org/TR/2011/WD-html5-20110525/offline.html#offline) first appeared in 2009, in Safari 4, it held a lot of promise as a tool for caching web pages for use offline. It was imagined as “HTML5 Application Cache” back when HTML itself was being further expanded to handle more use cases for web applications. A developer could create a simple cache manifest file with a list of files to be cached. Its simplicity looked elegant, but there was no mechanism for cache busting, and that made both developing a site and evolving the site over time quite frustrating. AppCache also had security challenges. So new web standards were created to replace it. Today, developers use Service Workers and Cache Storage instead.
> 
> WebKit [deprecated](https://commits.webkit.org/197701@main) AppCache with a warning to the Console in Safari 11.0. Then in 2021, we [removed support](https://commits.webkit.org/236651@main) for AppCache from Safari 15.0, with a few exceptions for third-party users of [WKWebView](https://developer.apple.com/documentation/webkit/wkwebview). Now we are [removing](https://bugs.webkit.org/show_bug.cgi?id=259783) those exceptions. This change to WebKit will only affect the rare web content loaded in older third-party apps that have JavaScript code which relies on the existence of AppCache related interfaces.

From https://webkit.org/blog/15443/news-from-wwdc24-webkit-in-safari-18-beta/#deprecations

Distilling this history down to Can I Use data — we removed support from Safari 15.0.


Fyrd:
Thanks!

PR #7095, opened 2024-06-16
Safari 18 drops JPEG 2000 support

Schweinepriester:
https://webkit.org/blog/15443/news-from-wwdc24-webkit-in-safari-18-beta/

> ### Images
> 
> WebKit for Safari 18 beta removes support for the JPEG2000 image format. Safari was [the only browser](https://caniuse.com/jpeg2000) to ever provide support.
>
> If you've been serving JPEG2000 files using best practices, then your site is using [the `picture` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) to offer multiple file format options to every browser. Safari 18 beta will simply no longer choose JPEG2000, and instead use a file compressed in JPEG XL, AVIF, WebP, HEIC, JPG/JPEG, PNG, or Gif --- choosing the file that's best for each user. Only one image will be downloaded when you use `<picture>`, and the browser does all the heavy lifting.
>
> We have noticed that some Content Deliver Networks (CDN) use User Agent sniffing to provide one file to each UA, offering only JPEG2000 images to Safari --- especially on iPhone and iPad. If you expect this might be happening with your site, we recommend testing in Safari 18 beta on both macOS Sequoia and iOS or iPadOS 18. If you see problems, contact your SaaS provider or change your image delivery settings to ensure your website provides fallback images using industry best practices.
>
> If you notice a broken site, please file an issue at [webcompat.com](https://webcompat.com/).


Fyrd:
RIP JPEG 2000 I guess!

Thanks!

PR #7101, opened 2024-06-26
add css text-box-trim to Chrome

Th3S4mur41:
The feature is implemented behind the `#text-box-trim` flag in Chrome 128 
https://chromestatus.com/feature/5174589850648576


Fyrd:
Thanks!

PR #7106, opened 2024-07-10
Update Chromium support for async clipboard

foolip:
According to https://chromestatus.com/feature/5861289330999296 this feature shipped in M66.

It was behind a flag before that, but confirming the correct versions doesn't seem worth the effort.

Also update Opera to the corresponding Opera version.


Fyrd:
Thanks!

PR #7108, opened 2024-07-12
Treat referrer policy as fully supported even absent strict-origin-when-cross-origin

foolip:
The note is about the default behavior, which is what happens when you don't use the feature (header or attributes).


Fyrd:
Sounds good, thanks!

PR #7111, opened 2024-07-19
Align Chrome/Safari sandbox support with BCD

foolip:
https://github.com/mdn/browser-compat-data/pull/23858 uses both source archeology and previous testing of the relevant browser releases to arrive at Chrome 5 + Safari 5 + iOS 4.


Fyrd:
Thanks!

PR #7117, opened 2024-07-26
Safari 3.1 supported `@font-face`

jensimmons:
Safari added support for `@font-face` in October 2007.  https://webkit.org/blog/124/downloadable-fonts/

Which means Safari 3.1 for Mac had support in March 2008. This PR updates that bit of information.


Fyrd:
Thanks!

Issue #7118, opened 2024-07-26
Samsung Internet 25 supports scrollbar CSS styling without prefixes


Gustavo-Silveira233:
**Page link:**
https://caniuse.com/css-scrollbar

**Current behavior:**
The page mentions that Samsung Internet 25 partially supports scrollbar styling using the webkit prefix. There is also a note:
_Supports scrollbar styling via CSS pseudo-properties. (not on standards track)_

**Bug description:**
If you look at the three pages below:

1. https://caniuse.com/mdn-css_properties_scrollbar-color
2. https://caniuse.com/mdn-css_properties_scrollbar-width
3. https://caniuse.com/mdn-css_properties_scrollbar-gutter

You'll see that Samsung Internet 25 supports all three standard properties, without prefixes.

**Expected behavior:**
Samsung Internet 25 should be marked in green as "✓ Supported".

PR #7119, opened 2024-07-28
Align CSS scrollbar styling to MDN for Samsung Internet 25

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7118.

I have not tested it myself, but relying much like OP of that issue on MDN:
- https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter#browser_compatibility


Fyrd:
Thanks!

PR #7122, opened 2024-07-30
Clarify the lack of support for `touch-action` in macOS Safari (where touch events don't exist)

jensimmons:
It's looking like the upcoming State of CSS 2024 survey will ask about `touch-action`. Which got me looking at the Can I Use page for it. I thought I'd clarify why Safari on macOS doesn't have support. And clarify & update the other footnotes while I was at it (especially since they are describing the past not the present).


jensimmons:
Ah it should be #5, not #1 in all these lines of code.

Fyrd:
Thanks!

Issue #7124, opened 2024-08-05
Stable Safari 17.6 has WebGPU as feature flag.


birkskyum:
I'm not sure exactly what version of Safari it started, but it seems that stable Safari 17.6 on macOS has a WebGPU feature flag which isn't mentioned in the table.

<img width="129" alt="Screenshot 2024-08-05 at 14 14 25" src="https://github.com/user-attachments/assets/92c5976a-6e90-480a-b6f2-fc47e31248ba">

<img width="844" alt="Screenshot 2024-08-05 at 14 14 58" src="https://github.com/user-attachments/assets/dce40cc1-c391-4a9c-af4c-7dafd1e74375">

Ugoku:
17.5 also has this flag

jensimmons:
I'm looking into exactly when this feature flag was added (off by default) to Safari. I know it shipped on by default in Safari Technology Preview 190 (as 'preview') on March 6, 2024.

jensimmons:
It was added in Safari 17.4.
https://bugs.webkit.org/show_bug.cgi?id=265948

jensimmons:
I have an email notification where @birkskyum asked:
> Just out of curiosity, how do you set the flag on iOS?

On iOS, iPadOS and visionOS:
Settings > Apps > Safari > Advanced > Feature Flags

jensimmons:
Thanks for filing this issue!

PR #7127, opened 2024-08-06
WebGPU is supported in Safari 17.4+ behind a flag

jensimmons:
PR for #7124.

Adding the fact WebGPU is available in Safari 17.4+ on both macOS and iOS, after flipping the feature flag. 

Updated the footnote to not say "macOS" so it can be used now for both. Also, replacing "experimental feature" with "feature flag". The word "experimental" is used by Chrome, not Safari. We call it a "feature flag" — which comes in three flavors "testable", "preview", and "stable". (Testable = can be tested by flipping the feature flag. Preview = on in Safari Technology Preview. Stable = on by default for all users.)

### DIFF
--- a/features-json/webgpu.json
+++ b/features-json/webgpu.json
@@ -392,10 +392,10 @@
       "17.1":"n",
       "17.2":"n",
       "17.3":"n",
-      "17.4":"n",
-      "17.5":"n",
-      "17.6":"n",
-      "18.0":"n",
+      "17.4":"n d #2",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

birkskyum:
@jensimmons , thank you for looking into this - just out of curiosity, how do you set the flag on iOS?  I only mentioned macOS in here, because I've never figured how to set any flags on iOS.

Issue #6615, opened 2023-02-15
FileSystemSyncAccessHandle with OPFS: Firefox and Safari links


js-choi:
This is regarding OPFS and the [entry for File System Access](https://caniuse.com/native-filesystem-api).

It would be informative for developers if these links were added to the resources page for File System Access:

* [Firefox support bug for Origin Private File System](https://bugzilla.mozilla.org/show_bug.cgi?id=opfs)
* [WebKit blog post for Origin Private File System](https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/)

In addition, this line:

> * [Firefox position: harmful](https://mozilla.github.io/standards-positions/#file-system-access)

…probably should be changed to something like:

> * [Firefox position: harmful (when not restricted to Origin Private File System)](https://mozilla.github.io/standards-positions/#file-system-access)
> * [Firefox intent to prototype Origin Private File System access](https://groups.google.com/a/mozilla.org/g/dev-platform/c/gc4BxcgkN7E)

## Background
Can I Use currently classifies Origin Private File System (OPFS) under the entry for File System Access (see #6570).

However, this situation is pretty confusing for developers due to unclear distinctions between the [File System Access incubator spec](https://wicg.github.io/file-system-access/), the [File System WHATWG standard](https://fs.spec.whatwg.org/), and the FileSystemSyncAccessHandle API (the latter of which would enable, e.g., SQLite using files from storage). File System Access builds on File System, it is File System and *not* File System Access that defines OPFS, and the FileSystemSyncAccessHandle API is seemingly defined in both File System and File System Access. File System Access also defines systems for allowing websites to non-sandboxed files, along with FileSystemSyncAccessHandle (which, again, is also defined in File System).

[Safari has implemented FileSystemSyncAccessHandle specifically for OPFS files](https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/). The announcement post refers to this as implementing “File System Access with [OPFS]” but does not refer to the File System standard.

[...]

PR #7129, opened 2024-08-14
Clarify that the File System Access API is not supported in Safari/Firefox

BenjaminAster:
Caniuse's [entry for the File System Access API](https://caniuse.com/native-filesystem-api) currently specifies that Firefox/Safari partially support the File System Access API, with a note saying that they "only support the Origin Private File System (OPFS) part". This made sense when the API was originally introduced and everything, including the Origin Private File System, was in a specification with the name "File System Access".

However, since then, the specification has been split into two parts: the [File System API](https://fs.spec.whatwg.org/) (which is basically the OPFS and all browsers support) and the [File System *Access* API](https://wicg.github.io/file-system-access/) (which only Chromium supports). So the definition of the term "File System Access API" has changed over the years and now specifically refers to *only* the part that allows for direct access to files/directories on the user's local file system. Therefore, saying that Firefox/Safari "partially support" the "File System Access API" does not make sense any more. Additionally, the note that "Desktop Chromium browsers currently support basic functionality and will be adding more of the API in the future" is also not true anymore; [since Chromium v105](https://chromestatus.com/feature/6383970247770112), all features in the specification are supported afaik.

Somewhat related issue: #6615


Fyrd:
Yup, this was definitely due for an update. Thanks for the research and updates!

PR #7132, opened 2024-08-23
Update webcodecs support for Firefox desktop

kbrosnan:
WebCodecs are targeting Firefox 130 see https://bugzilla.mozilla.org/show_bug.cgi?id=1908572

Fx Android is not supported yet.


Fyrd:
Confirmed, thanks!

PR #7133, opened 2024-08-26
Mark datalist as partial in Firefox due to missing date and time inputs

stof:
Firefox 110 added support for datalist on range and color inputs, but support is still missing for date and time inputs.

Marking this as partial support is consistent with the MDN data: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist#browser_compatibility


Fyrd:
Thanks!

PR #7135, opened 2024-08-28
chore: update status for webauthn on firefox mobile

taraldr:
Removing `#4`, since using an authenticator with a PIN definitely is supported, see screenshot.

Also adding `#7` since firefox on mobile doesn't support direct attestations due to them having an extra privacy requirement, which they haven't implemented the UI for.

Firefox on mobile working fine with a FIDO 2 authenticator with PIN enabled:
image

### DIFF
--- a/features-json/webauthn.json
+++ b/features-json/webauthn.json
@@ -583,7 +583,7 @@
       "127":"y"
     },
     "and_ff":{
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #7137, opened 2024-08-31
Opera dropped JPEG XL flag back in v96.

goodusername123:
Opera does not know about nor care about JPEG XL so the flag disappeared as soon as Opera landed on Chromium M110 which happened in version 96 of Opera.


Fyrd:
Thanks!

PR #7138, opened 2024-08-31
Theora support removed in Firefox 130.

goodusername123:
OGG video (Theora) is slated for removal in Firefox version 130 following Chrome's/Chromium's decision. Supported for 15 years before removal :(
https://bugzilla.mozilla.org/show_bug.cgi?id=1890370

This project exists as a polyfill: https://github.com/bvibber/ogv.js


Fyrd:
Confirmed, thanks!

PR #7140, opened 2024-09-04
Chrome supports Web Share API since 128

zac1st1k:
Web Share API support backported to Chromium 128

https://chromiumdash.appspot.com/commit/60426ee006d2587433b4c9f80b95dd6a19327ff1

Orignal Chromium 129 commit
https://chromiumdash.appspot.com/commit/52c8632e39b952492b2a9ba9c9c63c3a17b75490


Fyrd:
Thanks! FYI this should have been a lowercase `y`, went ahead and fixed that.

PR #7160, opened 2024-09-23
Mark Safari `createImageBitmap()` bug as fixed

ddbeck:
https://caniuse.com/createimagebitmap shows a partial implementation for Safari for a bug that was fixed in Safari 17. This PR elaborates on the note (linking to the de-duplicated bug) and marks full support on the versions since the fix.

References:

- https://bugs.webkit.org/show_bug.cgi?id=237082
- https://developer.apple.com/documentation/safari-release-notes/safari-17-release-notes#Canvas

Discovered in the course of https://github.com/web-platform-dx/web-features/pull/1707


Schweinepriester:
Woohoo! 🥳 

Do you know if it is also fixed for Safari on iOS? Surely? 
If so, you could also update the `ios_saf` entries :)

ddbeck:
@Schweinepriester I ran the test on iOS and it works there too. 🎉

Fyrd:
Thanks Daniel & @Schweinepriester !

PR #7161, opened 2024-09-24
Update Temporal's Firefox implementation bug

Fly7113:
The linked bug was closed more than a year ago, and a new [META] one has been opened.
In my opinion the newly linked bug gives a better understanding of the status of the implementation efforts going forward in Firefox.


Fyrd:
Looks good, thanks!

PR #7165, opened 2024-09-27
Correcting Safari support of AVIF

jensimmons:
Safari 16.1–16.3 did not support animated AVIF. Safari 16.4 shipped such support to all platforms. Somehow the update for 16.4 was only applied to Safari on macOS and not on iOS. This PR updates the data with the correct information.


jensimmons:
In other words, this PR removes Footnote 2 from Safari on iOS, starting with 16.4 — to make it match Safari on macOS. 

Screenshot 2024-10-01 at 2 10 27 PM

Without this change, developers could look at this and believe animated AVIF is still not supported on iOS, which is untrue.

Fyrd:
Thanks!

PR #7167, opened 2024-09-28
SSE 4.1 required note

blathers16:
WASM  SIMD requires SSE 4.1 to run.  This PR adds a note to this effect.
see https://github.com/bytecodealliance/wasmtime/issues/3809


Fyrd:
Thanks!

PR #7170, opened 2024-10-01
Add note about Style Queries in Safari 18

bramus:
Style queries don’t work on the document element in Safari 18.

Also see https://github.com/mdn/browser-compat-data/pull/24401


jensimmons:
Is there another way to phrase this so that it's clearer to developers what it means?
> The document element cannot be a container

"Document element" is not a phrase commonly used (at least not in the worlds I've been in). Root element? HTML element?

bramus:
I’m using the phrasing which @nt1m provided me in https://github.com/mdn/browser-compat-data/pull/24401#discussion_r1759626878

Happy to update the terminology to “root element” which indeed is more recognizable by authors.

Fyrd:
Thanks! Corrected the link format to use markdown.

PR #7171, opened 2024-10-01
Safari TP 204 supports Cookie Store API

Schweinepriester:
https://webkit.org/blog/15978/release-notes-for-safari-technology-preview-204/

> - Added support for the Cookie Store API. ([283654@main](https://commits.webkit.org/283654@main)) (135969444)

- https://tests.caniuse.com/cookie-store-api
- https://wpt.fyi/results/cookie-store?label=master&label=experimental&aligned=

Haven't tested (yet) myself, but perhaps by the time you look at this WPT has a run with TP 204 🤞


Fyrd:
Hmm, looking at those test results maybe it should be partial support?

Schweinepriester:
While I'm inclined to agree, I don't know where the bar is then?
Which tests are most important, what's the count to pass?

<https://tests.caniuse.com/cookie-store-api> is fully green at least:

image

Lacking a better idea, I'd say let's go for full first, surely somebody will open a PR if there's an aspect important enough to warrant partial…

Or we could ask Jen?

Fyrd:
> Lacking a better idea, I'd say let's go for full first, surely somebody will open a PR if there's an aspect important enough to warrant partial…

Yeah let's just go for this. :)

PR #7175, opened 2024-10-03
`image-orientation` shipped in iOS 13.4

ddbeck:
In the course of https://github.com/web-platform-dx/web-features/pull/1833, I discovered a discrepancy between caniuse and BCD. After some testing in BrowserStack (see https://github.com/mdn/browser-compat-data/pull/24520 for complete detaiils), I found that iOS 13.4, not iOS 14, first shipped support for `image-orientation`.


Fyrd:
Thanks Daniel!

Issue #7178, opened 2024-10-07
`background-attachment` is supported in Safari on iOS


jamesnw:
[`background-attachment`](https://caniuse.com/background-attachment) reports that Safari on iOS does not support `background-attachment:local` due to this bug: https://bugs.webkit.org/show_bug.cgi?id=219324

That bug is resolved, and BCD reports support starting in 15.4.

PR #7180, opened 2024-10-08
Align Safari on iOS to Safari and MDN for `background-attachment`

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7178.

I have not tested, but seems very plausible that the change never made it's way from Safari to Safari on iOS and also MDN indeed reports it as such: https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment#browser_compatibility


Fyrd:
Looks good, thanks!

Issue #7182, opened 2024-10-09
mutation-events not supported since Chrome 127


1jj:
https://caniuse.com/mutation-events are no longer supported by default since Chrome 127 and Edge 127.

https://developer.mozilla.org/en-US/docs/Web/API/MutationEvent
https://developer.chrome.com/blog/mutation-events-deprecation

Furthermore, the spec link is wrong.
https://www.w3.org/TR/uievents/#legacy-mutationevent-events
There are no such events specced here. Actually, there is no valid spec at all.

PR #7183, opened 2024-10-09
Updates to Mutation events in light of deprecation

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7182.

A lot of context there^ already, but for completeness links I came across while digging:
- https://www.w3.org/standards/history/uievents/
   - https://www.w3.org/TR/2024/WD-uievents-20240906/#legacy-mutationevent-events still in the spec
   - https://www.w3.org/TR/2024/WD-uievents-20240907/ removed
- https://chromestatus.com/feature/5083947249172480
- https://github.com/mozilla/standards-positions/issues/807
- https://github.com/WebKit/standards-positions/issues/192

### DIFF
--- a/features-json/mutation-events.json
+++ b/features-json/mutation-events.json
@@ -1,12 +1,12 @@
 {
   "title":"Mutation events",
   "description":"Deprecated mechanism for listening to changes made to the DOM, replaced by Mutation Observers.",
-  "spec":"https://www.w3.org/TR/DOM-Level-3-Events/#legacy-mutationevent-events",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
Also: A new `status` value indicating deprecation would be very nice 🤓

Fyrd:
Nice, thanks!

Fyrd:
> Also: A new `status` value indicating deprecation would be very nice 🤓

For now I've set to "unofficial" which I think is used for other deprecated features as well. But yeah, a separate status would be more accurate.

PR #7188, opened 2024-10-23
Set form validations to partially supported in Firefox Mobile

projecteurlumiere:
Mobile Firefox does not show HTML5 validation error notices. 
It seems to be a long-standing issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1510450


Fyrd:
Confirmed. And wow, that's surprising. Thanks for reporting!

PR #7190, opened 2024-10-24
Safari 15.4 reinstated `navigator.hadrwareConcurrency`

ddbeck:
See https://bugs.webkit.org/show_bug.cgi?id=233381 and https://github.com/mdn/browser-compat-data/pull/24815 for the full details, but the summary is that this returned to Safari.

Discovered in the course of https://github.com/web-platform-dx/web-features/pull/1931#discussion_r1812881017


Fyrd:
Interesting, thanks!

PR #7159, opened 2024-09-22
Chrome 131 supports `currentcolor` in Relative Color & more

Schweinepriester:
- https://chromestatus.com/feature/4755025804132352
- https://issues.chromium.org/issues/325309578
- https://bugzilla.mozilla.org/show_bug.cgi?id=1893966

Not entirely sure about the system color aspect, but seemingly those WPT tests are the only ones: https://wpt.fyi/results/css/css-color?label=experimental&label=master&aligned&view=interop&q=label%3Ainterop-2024-relative-color

PR #7192, opened 2024-10-24
Fix CSS Relative color for Chrome & Firefox

### LINKED ISSUE #7191: Relative Color fixes applied to wrong browser
It appears that https://github.com/Fyrd/caniuse/pull/7159 was meant to be applied to Chrome, but it was instead applied to Firefox. However, it looks like the note can also be removed from Firefox 133- https://bugzilla.mozilla.org/show_bug.cgi?id=1893966

In other words:

- Firefox 131 and 132 SHOULD have a note, and 133 should not.
- Chrome 131+ should NOT have a note.

https://caniuse.com/css-relative-colors

Schweinepriester:
Fixes #7191 (my own fail in #7159) 🙈


Fyrd:
Oof, yeah it's tricky when the version numbers are so similar 😅  Thanks!

PR #7199, opened 2024-10-30
Safari supports HTTP/3 for all users now

### LINKED ISSUE #7114: HTTP/3 usage might be underestimated, what does "partial support" i.e. "Enabled by default for a portion of users" really mean?
See:
https://caniuse.com/http3

Support could be as low as  75.86% or as high as 98% so it matters which and the numbers only depend on availability on Apple products.

E.g., for "tracked mobile" the "22.22%" "partial support" to get to 98.08% depends on how many users of Safari 14 up to 17 have have it enabled.

This "Enabled by default for a portion of users." seems like A/B testing, and wouldn't Apple have enabled for all users be now? Can someone test for me and see if HTTP/3 doesn't work for some version on iOS or macOS? To give me some confidence inn the low (or high) number.

It doesn't seem plausible that Apple enabled first for only some users with 14.4 in Sept 2020 and is still doing it year later with 17.5, so simply wrongly marked/not updated here?

Schweinepriester:
Fixes #7114.

Per https://github.com/Fyrd/caniuse/issues/7114#issuecomment-2448297609

> Safari now supports HTTP3 for all users, as of September 2024, no matter which version of Safari they are running.

For now I would argue for `y` for 18.1+, because that is the first minor version released after 2024-09:

https://developer.apple.com/documentation/safari-release-notes/safari-18_1-release-notes

> Released October 28, 2024 — 18.1 (20619.2.8)

18.0, released in 2024-09, could be either way from the start, but could also have been changed while it already was out, hence me playing it safe with 18.1.

While one could argue for `y` for all versions with the note, I feel like that would not represent the history as a retroactive change, though there's certainly room for debate.


jensimmons:
This PR is not accurate, since HTTP3 support did not ship with Safari 18.0. It was enabled in the same month as Safari 18.0 but it applies to all users on Safari 16.0+. 

This is a weird feature. I talked to three of our engineers to nail down the historical details, and represent them in this PR: https://github.com/Fyrd/caniuse/pull/7200

Schweinepriester:
@jensimmons PR is also fair:
- https://github.com/Fyrd/caniuse/pull/7200
- https://github.com/Fyrd/caniuse/pull/7200/files

You decide Fyrd! :)

Schweinepriester:
… aaand the simultaneous comment, haha :)

jensimmons:
I'm happy about the excitement!! Go HTTP3!

I was going to create a PR like this, the way @Schweinepriester did... just make Safari 18.0 (or he did 18.1) solid green, while the previous versions stayed they way they've been... since, you know, that's literally how every other "Foobar just shipped!" PR works for Can I Use. But then as I discussed this with our engineers, I realized, no actually the day we made this change, we made it for every version of Safari 16.0+. It's impossible for anyone who is running Safari from the last two years of releases to not have this change. So... that's why I wrote the PR at #7200 the way I did.

Fyrd:
Yeah I went with @jensimmons 's PR in this case though certainly both approaches seemed reasonable. Thanks as always @Schweinepriester !

PR #7200, opened 2024-10-30
Updating Safari support for HTTP3

jensimmons:
As of Sept 2024, Safari 16+ supports HTTP3 for all users. 

Previously, a subset of users had support enabled (as long as they had Safari 14 or later (including Big Sur or later for Safari on macOS). There was a feature flag available in Safari 14.0-16.3 that allowed users to force HTTP3 on (it was removed for Safari 16.4+). Unlike other feature flags, having the flag off did not disable HTTP3, it just meant maybe the user got it, maybe they didn't. 

It's complex to communicate all this in a dataset that assumes features ship with a new version of the browser. In this case, the changes are on being made elsewhere, and are not dependent on updating the browser.


jensimmons:
Resolves https://github.com/Fyrd/caniuse/issues/7114

Fyrd:
Thanks! Yeah I think this note is descriptive enough. Won't be long before all anyone cares about is that it's supported in Safari 16+ :)

PR #7204, opened 2024-11-07
Mark `background-clip: text` as fully supported + correct history of Safari support 

jensimmons:
`background-clip: text`  is very well supported, and authors should feel confident in using it to create effects like this [demo](https://codepen.io/jensimmons/pen/XWvxoaj?editors=1100):
Screenshot 2024-11-07 at 12 56 53 PM

However, on Can I Use, the vast majority of implementations of `background-clip: text` are currently marked "partial" because of a minor bug where if the author wraps the text in an inner element, _and_ puts position: relative or absolute on it, the background image doesn't appear. 

However, this's not at all a common situation. I can't think of even one use case for this: 

```
<h1><span>Hello world</span></h1>

h1 { 
  background-image: linear-gradient(red, blue);
  background-clip: text;
  color: transparent; 
}
span { position: relative; }
```

Meanwhile this works fine (see [demo](https://codepen.io/jensimmons/pen/poMxYWx/65d983fa3ebc35def667b0a677051239?editors=1100) of both):

```
<h1>Hello world</h1>

h1 { 
  background-image: linear-gradient(red, blue);
  background-clip: text;
  color: transparent; 
  position: relative;
}
```
Therefore, I do not believe implementations deserve getting a "partial support" designation. Noting the bug is a good idea (so I left the footnote) but Can I Use does not usually withhold full support for a bug on a rare condition.


jensimmons:
Second commit:

I noticed that Safari 3.1 and 3.2 were marked "unknown", while Safari 3.2 was the first browser will support. So I did some historical digging, talking to folks at Apple. 

The feature was committed to WebKit on March 20, 2008 (three days after Safari 3.1 shipped). Which means it was not in Safari 3.1, so I marked that not supported. And since Safari 3.2 was shipped with all changes in Nov 2008, that's the first this feature appeared, so I marked 3.2 as supported.

jensimmons:
Third commit:

Safari unprefixed background-clip in 2010
Can I Use marked Safari as unprefixing in Safari 14, in 2020. That struck me as odd, so I looked into it.

Actually, we unprefixed in Safari 5: https://github.com/WebKit/WebKit/commit/740414aaf941cd54946204d86912fede7bb0b103

I wondered where the Safari 14 date came from. It looks like someone noticed in 2020 that Safari was marked as requiring a prefix. They tested and saw that was wrong, that unprefixed worked. So they just changed the current version of Safari. They did not look into the history. https://github.com/Fyrd/caniuse/issues/6106

Safari 14 is definitely wrong.

jensimmons:
The fourth commit fills in the earliest data for Safari on iOS.

Fyrd:
Yeah, looking at this again I agree this bug doesn't warrant partial support. Thanks for the research & updates!

PR #7211, opened 2024-11-20
input type=week in Safari 18.2

jensimmons:
As seen in the beta release notes, Safari 18.2 adds support for input type=week on iOS. 

https://developer.apple.com/documentation/safari-release-notes/safari-18_2-release-notes#Forms


Fyrd:
Thanks!

PR #7213, opened 2024-11-22
Safari 18.2 supports `auxclick` for Pointer Events

jensimmons:
> Added auxclick event support for Pointer Events

https://developer.apple.com/documentation/safari-release-notes/safari-18_2-release-notes#:~:text=Added%20auxclick%20event%20support%20for%20Pointer%20Events

Which also means Safari Technology Preview does as well, and has for some time.


Fyrd:
Thanks!

PR #7216, opened 2024-12-02
Update css-text-box-trim.json

jensimmons:
Safari 18.2 adds support for `text-box`, `text-box-trim`, and `text-box-edge`. This PR also updates the title of the feature — since the `text-box` shorthand is a better way to use this. And the PR updates the description to explain this includes `text-box-trim`, and `text-box-edge`... and describing more accurately what it does. I also removed the reference to `leading-trim`, since 1) the early draft never shipped in customer-facing browsers, and 2) the name change is very old news.


jensimmons:
Screenshot 2024-12-02 at 2 28 37 PM

Is the "Limited availability across major browsers" box automated? How do we make it go away? This is a feature developers can use today to progressively enhance sites in browser with support. Warning them to not use it is unnecessary.

Fyrd:
Looks good, thanks!

Yes, that's the the Baseline banner that is added automatically. I don't think it's warning users not use a feature, just effectively summarizes the current browser support.

PR #7219, opened 2024-12-11
Chrome 133 supports CSS Text Box

Schweinepriester:
- https://groups.google.com/a/chromium.org/g/blink-dev/c/4NeTpgsi64A/m/hDMkvOs2AQAJ
- https://chromestatus.com/feature/5174589850648576
- WPT also looking good: https://wpt.fyi/results/css/css-inline/text-box-trim?label=experimental&label=master&aligned
- though there appears to be some confusion if it is ready yet: https://github.com/WebKit/standards-positions/issues/432#issuecomment-2536867135


Fyrd:
Thanks!

yisibl:
It is not currently shipped, and we need to judge it based on the configuration in this file.

Though it may eventually catch up with Chrome 133.

`status: "experimental",`
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/runtime_enabled_features.json5;drc=23e91c0259da94bbc2dcdbefecb94151edcc56a7;l=1299

PR #7244, opened 2025-01-23
Update HEVC for Firefox

tats-u:
See https://github.com/Fyrd/caniuse/issues/6858#issuecomment-2609524603 for details.


lilydjwg:
HEVC is [now available on Linux](https://bugzilla.mozilla.org/show_bug.cgi?id=1894818) too. Also, it's [available on Android](https://bugzilla.mozilla.org/show_bug.cgi?id=1937193). The pref [has changed to `media.hevc.enabled`](https://bugzilla.mozilla.org/show_bug.cgi?id=1941945).

tats-u:
You should write the version (137).

tats-u:
The pref should be enabled in all OSes so the change of the name should not be affected.

Fyrd:
While I appreciate the level of detail addressed here I don't think it's necessary to preserve the historical information of partial support behind a feature flag in older versions. As such I'm going to close this is favor of #7276. Thanks!

PR #7256, opened 2025-02-06
Update `inputmode` support on desktop Safari

ddbeck:
This reconciles Safari's desktop support with iOS and mdn/browser-compat-data. While macOS doesn't have a virtual keyboard to change, the reflection API does work and apparently this does change autocomplete behavior.

See https://github.com/web-platform-dx/web-features/pull/2573 for the inspiration for this PR and further discussion on expected behavior.

Thank you!


Fyrd:
Thanks for the information! Sounds reasonable. 👍

PR #7257, opened 2025-02-07
Update hevc.json

CKingX:
Updated HEVC support for Firefox:
* Firefox 134 added HEVC decoding for Windows devices with hardware decoding by default (can also be seen by release notes: https://www.mozilla.org/en-US/firefox/134.0/releasenotes/
* Firefox 136 added support for HEVC decoding for macOS (release notes here: https://www.mozilla.org/en-US/firefox/136.0beta/releasenotes/)

The one thing I am unclear on is 10-bit support on macOS and on Firefox 134. I have Firefox 135 already. I do not want to downgrade and mess anything up, and I can't test this in a VM (because you need hardware passthrough). But Firefox does support 10-bit decoding in Firefox 135.

### DIFF
--- a/features-json/hevc.json
+++ b/features-json/hevc.json
@@ -234,10 +234,10 @@
       "131":"n d #7",
       "132":"n d #7",
       "133":"n d #7",
-      "134":"n d #7",
-      "135":"n d #7",
-      "136":"n d #7",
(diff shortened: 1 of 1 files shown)


CKingX:
Wait never mind, this is a duplicate of #7244

PR #7259, opened 2025-02-11
Magnetometer: entire API is behind a flag in Chromium

ddbeck:
via @foolip in https://github.com/web-platform-dx/web-features/issues/2631

> In Chromium the features is controlled by a flag which is off by default: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/runtime_enabled_features.json5;l=3940-3943;drc=bdbfc851687b28b24b80bccbc4c027bfe05a701b


Fyrd:
Thanks! Looks like this was the result of the initial addition of the Magnetometer feature, seems [I noticed that it needed to be fixed](https://github.com/Fyrd/caniuse/pull/4204#issuecomment-387617275) but failed to actually do so. 😬

PR #7262, opened 2025-02-18
Correct errors in Push API

jensimmons:
Correcting two mistakes in data about support in Safari:

1) "Partial support" on macOS is due to the requirement for macOS Ventura. That requirement no longer applies for Safari 18.0+. In order to apply this note to Safari 16.1-17.7 only, I split footnote 5 ("Only available on macOS 13 Ventura or later and only in Safari itself, not WKWebView nor SFSafariViewController") into two separate footnotes.

2) Footnote 6 currently reads "Requires website to first be [added to the Home Screen](https://caniuse.com/web-app-manifest). Delivered silently, meaning no sound, vibration, haptics or screen wake." First, this footnote is being renumbered as footnote 7 (to make 5 become 5 & 6). Second, the claim "Delivered silently, meaning no sound, vibration, haptics or screen wake" is untrue — so I've simply deleted the sentence. More information about how sound, vibration, haptics or screen wake work with Web Push on iOS is available at https://webkit.org/blog/12945/meet-web-push/.


Fyrd:
Thanks! Looked into it a bit and it seemed that line came from #6688. I believe that was an issue that occurred while the feature was in beta but then got fixed (via this [reddit thread](https://www.reddit.com/r/ios/comments/125v1ft/ios_164_web_push_notifications_sound_vibration/)),  so agree it can be removed. Thanks!

PR #7265, opened 2025-02-23
Edge 132 surely also supports CSS Text Box

Schweinepriester:
- https://caniuse.com/css-text-box-trim
- https://wpt.fyi/results/css?label=master&label=experimental&aligned=&q=text-box (latest versions technically 3 tests less than Chrome, but …)
- https://tests.caniuse.com/css-text-box-trim seems outdated? 😅
- or at least I've found no indication to the contrary


Fyrd:
Yup, looks like the tests need updating. Thanks for the update and for calling this out!

Fyrd:
Test has now been updated (needed the value `trim-both` rather than `both`)

PR #7269, opened 2025-02-27
Add Cross-Document View Transitions

### LINKED ISSUE #7028: View Transitions Module Level 2
The draft for [View Transitions Module Level 2](https://drafts.csswg.org/css-view-transitions-2/) was just made available a week ago, adding support for cross-document navigation.

bramus:
This is an attempt at adding Cross-Document View Transitions. Ideally the existing `view-transitions.json` would be renamed to `same-document-view-transitions.json`, but I assume that would break the existing http://caniuse.com/view-transitions.

This (partially) fixes #7028 as this PR only covers Came-Document View Transitions, whereas L2 includes more than just that.


Fyrd:
Closing now #7290 has been merged.

PR #7273, opened 2025-02-28
Removing redundant note from WebGPU

jensimmons:
I noticed the page for WebGPU says "All major browser engines are working on implementing this spec. All major browser engines are working on implementing this spec."

Hm. The same sentence twice? I came here to fix it. But in JSON, it's only there once. Interesting. Well, this note isn't needed at all in Feb 2025, since seeing feature flags for both WebKit and Gecko browsers tells the same story. So I'm deleting it, expecting that will fix the bug as well.

Screenshot 2025-02-28 at 1 42 08 PM

### DIFF
--- a/features-json/webgpu.json
+++ b/features-json/webgpu.json
@@ -657,6 +657,7 @@
       "3.0-3.1":"n d #1"
     }
   },
+  "notes":"",
   "notes_by_num":{
     "1":"Can be enabled in Firefox with the `dom.webgpu.enabled` flag.",
     "2":"Can be enabled in Safari with the `WebGPU` feature flag.",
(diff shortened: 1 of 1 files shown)


features-json/webgpu.json
```diff
@@ -657,7 +657,6 @@
       "3.0-3.1":"n d #1"
     }
   },
-  "notes":"All major browser engines are working on implementing this spec.",
```
Schweinepriester, line 660: I believe it would need to be

```
"notes":"",
```

to make the pipeline happy :)

(sadly I can't add it as a proper easy to apply suggestion)

Fyrd:
Good catch, thanks!

PR #7277, opened 2025-03-05
Adding 7 more MDN links to CSS features

gregoryhammond:
I found 7 more CSS features that had missing MDN links, so I added them.

I thought it would be better to do this all as one pull request, rather than one pull for each small change.


Fyrd:
Thanks!

PR #7282, opened 2025-03-12
More Safari 18.4: Opus & Cookie Store API

Schweinepriester:
<https://developer.apple.com/documentation/safari-release-notes/safari-18_4-release-notes>

> Added support for the Cookie Store API. (135969444)

&

> Added support for Ogg Opus and Ogg Vorbis. (131407707)

Little doubt regarding Cookie Store API because it was in TP already, but Opus needs testing for sure.  
For easy access: https://tests.caniuse.com/opus


Fyrd:
Thanks!

PR #7283, opened 2025-03-12
Safari's support for Image Capture API

jensimmons:
Image Capture API is supported in Safari 18.4.

This P.R. also removes the "partial support" designation. It was there because of a complaint that Safari’s implementation does not support `grabFrame()`.

Instead Safari supports `takePhoto`, which can do everything `grabFrame` does — plus has an optional PhotoSettings parameter that allows you to apply constraints to the track just while the snapshot is taken, including constraints that aren't appropriate for a live video track, e.g. flash, red-eye reduction, etc. The `takePhoto` method is newer than `grabFrame`, and was created to address its shortcomings.

Safari has the full set of capabilities Image Capture API intends.

### DIFF
--- a/features-json/imagecapture.json
+++ b/features-json/imagecapture.json
@@ -18,7 +18,7 @@
     }
   ],
   "bugs":[
-
+  
   ],
   "categories":[
     "DOM",
--- a/features-json/imagecapture.json
+++ b/features-json/imagecapture.json
@@ -18,7 +18,7 @@
     }
   ],
   "bugs":[
-    Safari does not support `grabFrame()`, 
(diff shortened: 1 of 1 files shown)


Fyrd:
I'd be okay with this if there was a signal that `grabFrame()` was going to be deprecated. However as far as I can tell it's part of the specification and seems to serve the purpose of allowing a more performant way of getting image data while the stream continues without temporarily pausing it as `takePhoto()` allows. Given that I think the partial support remains appropriate.

Fyrd:
Closing as per above comment.

PR #7302, opened 2025-04-04
Mention that some data on caniuse.com is from browser-compat-data

tyilo:
I was trying to programmatically use some data available on caniuse.com, but could not find it in the data.

Providing this link should help people find the data they need.


Fyrd:
Good addition, thanks!

PR #7308, opened 2025-04-14
Align Ambient Light Sensor for Chrome for Android

### LINKED ISSUE #7307: AmbientLight - Chrome for Android is possible to use with flags
https://caniuse.com/ambient-light

It's is same as Desktop Chrome(`n d #2`). but not other browsers.
Seems like minimum support is same as Desktop Chrome version.

Schweinepriester:
Fixes #7307.

I have not tested, but plausible and https://developer.mozilla.org/en-US/docs/Web/API/AmbientLightSensor#browser_compatibility has it listed as well 0:-)


Fyrd:
Can confirm, looks like my tests were out of date so that's probably how this got missed before. Updated now and confirmed, thanks!

Issue #7327, opened 2025-06-02
Add CSS sibling-index() and sibling-count()


Th3S4mur41:
As Chrome intends to ship with Chrome 138
https://groups.google.com/a/chromium.org/g/blink-dev/c/opsqx1cBPyc/m/pJaRwh7UAAAJ 

See also: 
- https://github.com/w3c/csswg-drafts/issues/1026#issuecomment-2923716398
- https://drafts.csswg.org/css-values-5/#tree-counting

Robotic-Brain:
Duplicate of #7324

Fyrd:
Available via MDN BCD: https://caniuse.com/mdn-css_types_sibling-count,mdn-css_types_sibling-index

PR #7324, opened 2025-05-23
Add css-tree-counting-functions

Robotic-Brain:
Add upcoming CSS tree counting functions confirmed to ship in Chrome 138.

I'm pretty sure Chrome 133+ is the only browser supporting this feature right now.
Mozilla Firefox nothing implemented yet: https://bugzilla.mozilla.org/show_bug.cgi?id=1953973
Safari also nothing yet: https://github.com/WebKit/standards-positions/issues/471

Closes #7327

PR #7328, opened 2025-06-02
Add CSS sibling-index() and sibling-count()

Th3S4mur41:
Closes https://github.com/Fyrd/caniuse/issues/7327

Add [Tree Counting Functions](https://drafts.csswg.org/css-values-5/#tree-counting) with support starting in Chrome 138.

Support was verified with small demo in Chrome Canary (139) and Edge Dev (138)


Robotic-Brain:
Duplicate of #7324

Also Chrome 133 already supports it behind experimental flags.

Issue #7336, opened 2025-06-18
mutation-events are history


1jj:
This follows #7182 

- Firefox 140 removes mutation events
- Chrome 127 already did and recently removed the flag (which version?)
- Safari implemented removal (shipped?)

https://bugzilla.mozilla.org/show_bug.cgi?id=1963043
https://github.com/whatwg/dom/issues/305#issuecomment-2902826519

mfreed7:
> * Chrome 127 already did and recently removed the flag (which version?)

Chrome 139 [removed the flag](https://chromiumdash.appspot.com/commit/63b5a7bf0507136eb25b9da446be9ca0fb2c7af1).

PR #7337, opened 2025-06-18
Mutation-events are history

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7336.

As OP there unsure about the version, but just changed last/beta/future 0:-)


Schweinepriester:
https://chromestatus.com/feature/5083947249172480 says up to including chrome 134.

we'd probably also need to align edge & opera…?

Schweinepriester:
> we'd probably also need to align edge & opera…?

… changed the last version; also chromium mobile

1jj:
> https://chromestatus.com/feature/5083947249172480 says up to including chrome 134.

I would not trust this number ...

Fyrd:
Will merge as-is for now, thanks!

Issue #7267, opened 2025-02-25
Add CSS if() function


Schweinepriester:
- https://drafts.csswg.org/css-values-5/#if-notation
- Chrome
  - Intent to ship (136): https://groups.google.com/a/chromium.org/g/blink-dev/c/EOz7NK6Y0cE/m/VsZveC-fAwAJ
  - https://chromestatus.com/feature/6313805904347136?gate=5199452060778496
  - https://issues.chromium.org/issues/346977961
- Firefox
  - Request for position: https://github.com/mozilla/standards-positions/issues/1167
- Safari
  - Request for position: https://github.com/WebKit/standards-positions/issues/453

PR #7345, opened 2025-07-09
Create CSS if() function dataset

mfranzke:
Resolves https://github.com/Fyrd/caniuse/issues/7267

I've created all of the data based on the related issue #7267 and my own research. And I've added the polyfill that I provide myself as Open Source to support this functionality.


Fyrd:
Technically already exists at https://caniuse.com/mdn-css_types_if but I appreciate the extra links and research so happy to include as caniuse feature as well. Thanks!

PR #7348, opened 2025-07-21
Align createImageBitmap with web-features

ddbeck:
This PR proposes letting `createImageBitmap()` be fully supported in Firefox, despite not supporting the `resizeQuality` option.

Over in web-features, it was proposed to mark `createImageBitmap()` as being Baseline and this recently happened. The main justification is that `resizeQuality` is specified as a "[preference](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#:~:text=enumeration%20is%20used%20to%20express%20a%20preference%20for%20the%20interpolation%20quality%20to%20use%20when%20scaling%20images)" or hint to browsers about how to scale the image and does not bind implementers to use a specific scaling algorithm or strategy. The suggestion here is that Firefox ignoring this hint is probably not an impediment to using `createImageBitmap()` in most applications. See https://github.com/web-platform-dx/web-features/issues/3139 for further discussion.

If you merge this PR, then we'd merge TK on web-features, to align the features and reestablish the link between them.


Fyrd:
This seems reasonable to me so will go ahead and merge it. I do see the possibility of support for `resizeQuality` being tracked independently though if it seems important to developers.

PR #7362, opened 2025-08-06
Fix grammar in css-text-box-trim.json

danmooney3:
Fix grammar ("it's" == "it is"; "its" is the possessive form which was originally intended)


Fyrd:
Thanks!

PR #7363, opened 2025-08-07
Firefox supports service workers in private browsing since FF138

sandrosc:
see: https://bugzilla.mozilla.org/show_bug.cgi?id=1320796


Fyrd:
Thanks!

PR #7367, opened 2025-08-11
Firefox supports service workers in private browsing since FF138

sandrosc:
remove outdated bug entry (see #7363)


Fyrd:
Thanks!

PR #7374, opened 2025-08-29
Firefox is letting View Transitions ride the trains in 144

Schweinepriester:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1985809#c2
- https://hg-edge.mozilla.org/integration/autoland/rev/b0fe1cbd4671

Not entirely sure what to do about Level 2/cross-document, perhaps you got a better idea? 🤔


Fyrd:
Thanks! I think it's fine for now to use partial support for level 1 and full for level 2? But I'm not too familiar with the support for level 2.

PR #7382, opened 2025-09-07
Add MDN Glossary entry for Zstandard compression

Schweinepriester:
https://developer.mozilla.org/en-US/docs/Glossary/Zstandard_compression


Fyrd:
Thanks!

Issue #7360, opened 2025-08-06
Safari on macOS 26 already supports ZStandard


skhaz:
See here https://imgur.com/a/17dMlwy

Schweinepriester:
I was about to send a PR because on the first public beta for iPadOS it looked like this using
- https://tests.caniuse.com/zstd
- https://www.daniel.priv.no/tools/zstd-browser-test/

<img width="1210" height="834" alt="Image" src="https://github.com/user-attachments/assets/e4cf3cf7-efff-4b9e-bc83-1e8fff34c7d3" />


but having updated to the second one (23A5308g), it looks less like it 🤔 

<img width="834" height="1210" alt="Image" src="https://github.com/user-attachments/assets/1f0fce6e-af17-456f-9e0e-d4acd2acda3b" />

Haven't found any feature flag in the settings to tinker with.

Also: https://wpt.fyi/results/fetch/content-encoding/zstd?label=master&label=experimental&aligned&q=zstd%20

Sooo looks to me a bit like they're testing it currently (at least on iPadOS, maybe Safari on macOS is further ahead?), which I'm happy about, but I wouldn't update the support table yet 0:-)

Any comments by any chance, @jensimmons? :)

jensimmons:
Digging into this, it looks like it's _not_ included in the beta for Safari 26.0.
 
Stay tuned...

jensimmons:
It think it was? But got disabled for some reason, needing a bit more bake time? Something.

PR #7383, opened 2025-09-09
Mark Zstandard as partial supported for Safari 26+

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7360.

As we got RCs today

- https://www.macrumors.com/2025/09/09/apple-seeds-ios-26-rc/
- https://www.macrumors.com/2025/09/09/apple-seeds-macos-tahoe-rc/

I gave Zstandard on Safari 26 another test and it's the same as since Beta 2 - supported but not advertised/requested:

TODO

Technically I have only tested on iPadOS and assume it's the same across the board.

### DIFF
--- a/features-json/zstd.json
+++ b/features-json/zstd.json
@@ -449,8 +449,8 @@
       "18.3":"n",
       "18.4":"n",
       "18.5-18.6":"n",
-      "26.0":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Nice, thanks!

PR #7392, opened 2025-09-24
Fix link syntax in view transitions 2 note

Schweinepriester:
https://caniuse.com/cross-document-view-transitions

whoops, hehe, my bad 😅


Fyrd:
Thanks!

PR #7393, opened 2025-09-24
WebGPU only being enabled by default on macOS 26 Tahoe or later

### LINKED ISSUE #7390: WebGPU in Safari 26 requires macOS Tahoe to be on by default
WebGPU in Safari 26 requires macOS Tahoe to be on by default. At first I thought it was a bug so I reported https://bugs.webkit.org/show_bug.cgi?id=299237 but there it was confirmed that "navigator.gpu requires macOS Tahoe, iOS 26, visionOS 26 or later".

Maybe it's possible to add a note to https://caniuse.com/?search=webgpu saying it's only "supported" in Tahoe+

Note: It's still possible to enable the WebGPU feature in "Preview" mode, but that is hidden behind developer settings / feature flags.

Schweinepriester:
Fixes #7390, all details there. Didn't test myself, but because of the webkit bug little doubt is left, imho.

### DIFF
--- a/features-json/webgpu.json
+++ b/features-json/webgpu.json
@@ -441,9 +441,9 @@
       "18.3":"n d #2",
       "18.4":"n d #2",
       "18.5-18.6":"n d #2",
-      "26.0":"y",
-      "26.1":"y",
-      "TP":"y"
+      "26.0":"a #7",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #7399, opened 2025-09-30
Safari 18.4 & later supports web fonts on initial letter

jensimmons:
Back in Safari 18.4, a change was made that now allows type with `initial-letter` to be typeset with a web font. 

The change was made here: https://github.com/WebKit/WebKit/commit/68c05597505a5eb712c742bfed10d35be8db2654

I updated the footnotes to be more consistent and clear. And removed footnote 1 from Safari 18.4 and later, since now there is support for web fonts.

### DIFF
--- a/features-json/css-initial-letter.json
+++ b/features-json/css-initial-letter.json
@@ -450,11 +450,11 @@
       "18.1":"a x #1 #2",
       "18.2":"a x #1 #2",
       "18.3":"a x #1 #2",
-      "18.4":"a x #1 #2",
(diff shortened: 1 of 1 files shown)


jensimmons:
For clarification, footnote 2 still applies. And the prefix is still required. But this improvement does have significant impact, making it possible to use the basic functionality of initial letter in Safari, with the prefix, while using web fonts (as most modern sites do).

julietagarciahernandez4-hash:
[url]

jensimmons:
Thank you!!!

Issue #7395, opened 2025-09-25
Temporal shipping in Chrome 144


Schweinepriester:
- https://x.com/robpalmer2/status/1970874959041040831
- https://x.com/robpalmer2/status/1970895314396774625
- https://boajs.dev/blog/2025/09/24/temporal-release
- https://chromestatus.com/feature/5668291307634688

Once Chrome 144 is available in the data, I'll send a PR asap o7

Schweinepriester:
- https://fed.brid.gy/r/https://bsky.app/profile/did:plc:cs3zn3kikj5kwzljlvyr55sg/post/3lzobffomg62t
- https://groups.google.com/a/chromium.org/g/blink-dev/c/s58GKzoQZFg/m/iQ6OBGuLAAAJ

PR #7401, opened 2025-10-05
Chrome 144 is shipping Temporal

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7395, see info there - most important perhaps https://chromestatus.com/feature/5668291307634688.

also removing the redundant(?) note on https://caniuse.com/temporal:

image

how is

> All major browser engines are working on implementing this spec.

edited or the appearance controlled though? 🤔

### DIFF
--- a/features-json/temporal.json
+++ b/features-json/temporal.json
@@ -706,7 +706,7 @@
       "3.0-3.1":"n"
     }
   },
-  "notes":"All browsers are working on implementing this.",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks! Yeah that line is currently part of the baseline-status logic, it was suggested to include that for tracked web-features which should all be features that the major browsers are working on.

PR #7403, opened 2025-10-09
Mark current implementations of JPEG XL as partial for the lack of progressive decoding

Schweinepriester:
As brought to my attention in https://github.com/web-platform-tests/interop/issues/994#issuecomment-3364648793 and discussed there, mark Safari and Firefox (Nightly) as partial, because they lack support for progressive decoding of JPEG XL.
Now, probably there are other features missing, so it will surely be tricky over time as other formats have been, but as progressive decoding is (or was?) such a staple of JPEG, I think for this current step down partial is appropriate…

More than usual, this is still a _serve_ for discussion how to handle this, esp. as I'm still unsure which test (file) to actually test with. For now I've used https://google.github.io/attention-center/. If you decide to merge, would _really_ be good to get something onto https://tests.caniuse.com/jpegxl 🙏

Here's a recording from Firefox Nightly with the flag enabled, but Safari behaves the same:

https://github.com/user-attachments/assets/a608503c-673d-4cc2-b5b2-c6e2892c69e6


goodusername123:
Unsure if this has much bearing on anything but for me personally Firefox nightly not supporting progressive decoding is a bit of a weird situation as it's in that state currently due to intentionally being abandoned. Forks or browsers derived from Firefox have progressive decoding support for JXL such as [Waterfox](https://www.waterfox.net/), [Pale-Moon](https://www.palemoon.org/)/[Basilisk](https://www.basilisk-browser.org/), and [Zen](https://zen-browser.app/) for example.

Schweinepriester:
> Firefox nightly not supporting progressive decoding is a bit of a weird situation

yeah, but I thought with the looming switch to that rust lib, it would nice to track that and thus somewhen remove the note

Fyrd:
I can see an argument be made either way for whether or not this should count as partial support... I guess it does seem like you say a loss over plain JPEG where progressive decoding is just expected to be supported. So I'll merge for now. Thanks!

PR #7404, opened 2025-10-10
WebGPU: Return note 2 for Safari 26+

Schweinepriester:
In light of https://github.com/Fyrd/caniuse/pull/7393 I realized today that note 2, which mentions the feature flag, i.e. how to turn it on, is still relevant for those previous versions of macOS, hence return it :)


Fyrd:
Thanks!

PR #7408, opened 2025-10-30
Firefox is enabling `@scope` by default on all channels

Schweinepriester:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1991105#c2
- https://phabricator.services.mozilla.com/D270674

not sure starting with which version yet, so I went with the highest in the data (but nightly is currently 146)


Fyrd:
Thanks!

PR #7410, opened 2025-10-30
Change values for Chrome hls to yes in versions 142, 143, and 144

blineadam:
Updated values for chrome hls keys 142, 143, and 144 to 'y'.

Added in chromium 142 natively and verified in chrome 142 https://chromium-review.googlesource.com/c/chromium/src/+/7047405


Fyrd:
Confirmed, thanks!

PR #7412, opened 2025-11-03
Firefox removed support for vibration.json in Firefox 129

jensimmons:
Release note: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/129#:~:text=The%20Navigator%2Evibrate%28%29%20method%20has%20been%20removed%20%28Firefox%20bug%201653318%2C%20Firefox%20bug%201900037%29%2E

Issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1900037#c0

It did not work before removal — so at least in Firefox 128 & an unknown number of versions before, it apparently did not do anything. (See issue for more details.) Without knowing when it stopped working, I cannot document that — so this P.R. simply marks 129+ as not supported.


Fyrd:
Confirmed for Firefox on Android as well, so included that. Thanks!

jensimmons:
Oh yes! I should have done mobile. I make that mistake too often — only changing desktop. Thanks!

PR #7414, opened 2025-11-05
iOS 26 and later support <link rel=dns-prefetch>

daytonlowell:
Source https://webkit.org/blog/17333/webkit-features-in-safari-26-0/#networking


Fyrd:
Thanks!

Issue #7422, opened 2025-11-22
Firefox 145 has support for WebGPU on macOS 26+


bes:
The implementation status page for WebGPU https://github.com/gpuweb/gpuweb/wiki/Implementation-Status says that it is enabled on macOS 26 from version 145.

I just tested it on macOS 26 in Firefox 145 and the bevy WebGPU examples, e.g. https://bevy.org/examples-webgpu/3d-rendering/color-grading/ works.

bes:
(On Apple Silicon)

PR #7423, opened 2025-11-22
Firefox is shipping more WebGPU

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7422.

Links from there and more I dug up:
- https://github.com/gpuweb/gpuweb/wiki/Implementation-Status#firefox
- https://www.firefox.com/en-US/firefox/145.0/releasenotes/#:~:text=The-,WebGPU,-DOM%20API%20(
- https://bugzilla.mozilla.org/show_bug.cgi?id=1849913#c3
- https://hg.mozilla.org/mozilla-central/rev/c4ce9cf590ea

I think it would be fair to mark it as full support, given how Chrome is handled, when there is no constraint to the macOS version nor architecture anymore - or that becomes obsolete because Firefox isn't supported on it anymore. I suppose that will be a while.

### DIFF
--- a/features-json/webgpu.json
+++ b/features-json/webgpu.json
@@ -251,10 +251,10 @@
       "142":"a d #1 #6",
       "143":"a d #1 #6",
       "144":"a d #1 #6",
-      "145":"a d #1 #6",
-      "146":"a d #1 #6",
-      "147":"a d #1 #6",
-      "148":"a d #1 #6"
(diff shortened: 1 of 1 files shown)


Fyrd:
> I think it would be fair to mark it as full support, given how Chrome is handled, when there is no constraint to the macOS version nor architecture anymore

Yup, agree with this. Thanks!

Issue #7444, opened 2025-12-30
Wrong Safari support for scrollbar styling


nt1m:
This page is incorrect: https://caniuse.com/css-scrollbar

scrollbar-width/gutter were released in 18.2: https://caniuse.com/?search=scrollbar-gutter , https://caniuse.com/?search=scrollbar-width

scrollbar-color was released in 26.2: https://caniuse.com/?search=scrollbar-color

I would consider maybe just dropping the umbrella item in favor of the individual ones.

PR #7447, opened 2026-01-03
Align Safari support for scrollbar styling

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7444.

> I would consider maybe just dropping the umbrella item in favor of the individual ones.

definitely a consideration imho, but I thought fix it first :)


Fyrd:
Thanks!

PR #7450, opened 2026-01-07
Align Temporal support for Safari TP to MDN

Schweinepriester:
inspired by https://piccalil.li/blog/date-is-out-and-temporal-is-in/ I took another look how it's going and noticed aligning to MDN is possible:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal#browser_compatibility
* https://caniuse.com/mdn-javascript_builtins_temporal

progress! 📈

### DIFF
--- a/features-json/temporal.json
+++ b/features-json/temporal.json
@@ -464,7 +464,7 @@
       "26.1":"n",
       "26.2":"n",
       "26.3":"n",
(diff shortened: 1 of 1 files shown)


Fyrd:
Nice, thanks!

PR #7453, opened 2026-01-11
Update css-marker-pseudo.json for correct Safari TP status

clagnut:
Current Safari TP status incorrect: Safari Tech Preview still only partially supports ::marker (just colour and font, not content)


Fyrd:
Thanks!

PR #7465, opened 2026-01-29
iOS Safari's `background-attachment: fixed` is partially implemented

ddbeck:
This is my best attempt to represent [bug 275247](https://bugs.webkit.org/show_bug.cgi?id=275247), cribbed from https://github.com/mdn/browser-compat-data/pull/28887/.


Fyrd:
Thanks! Oof, yeah I remember iOS Safari struggling with this since forever.

PR #7466, opened 2026-01-31
Update Zstandard support for Safari

ekinnear:
Update Zstandard support for Safari

Zstandard is now supported and enabled by default in Safari 26.3 when running on a corresponding OS version.


Schweinepriester:
exciting, nice! 🎉 

do you happen to have any public source like a changelog, commit or alike?
I recognize you likely know this confidently, but something public would be even nicer :)

WPT: https://wpt.fyi/results/fetch/content-encoding/zstd?label=master&label=experimental&aligned

ekinnear:
> do you happen to have any public source like a changelog, commit or alike?

Great question. 26.3 is in beta at the moment, so anyone can try it out to verify. I don't have a public changelog link yet, but there is often one around the time that the release moves out of beta, so can update if one becomes available. Unfortunately there is no commit link, since this is implemented as part of the platform's HTTP support, rather than in WebKit itself.

ekinnear:
Update: https://webkit.org/blog/17798/webkit-features-for-safari-26-3/

PR #7468, opened 2026-02-09
Added MDN web docs links to details

gregoryhammond:
I added MDN web docs links (both details and summary) within the details element. I put them at the bottom because I figured that was probably the best way for them.


Fyrd:
Thanks!

PR #7476, opened 2026-02-17
min(), max(), & clamp() now supported for `sizes` in Safari 26.4

jensimmons:
https://developer.apple.com/documentation/safari-release-notes/safari-26_4-release-notes#:~:text=Added%20support%20for%20using%20the%20min()%2C%20max()%2C%20and%20clamp()%20math%20functions%20in%20the%20sizes%20attribute%20of%20%3Cimg%3E%20elements.

Removing footnote 4 from Safari 26.4


Fyrd:
Thanks!

PR #7477, opened 2026-02-19
link-icon-svg: update FF for Android support information

mina86:
I’ve tested the feature on Firefox for Android 147 and it works fine.
The icon is shown in tab view and used adding shortcut to desktop.  I’ve
tested with `type=image/svg` and `type=image/svg+xml`—both work.

I suspect this feature has been supported for a while.  A two-year old
[Bug 1821701](https://bugzilla.mozilla.org/show_bug.cgi?id=1821701)
mentions it and tags firefox120 as fixed.

Based on that, marking anything 120 onward as `y`.


Fyrd:
Confirmed, thanks! Only the latest version is tracked so updated the PR appropriately.

PR #7478, opened 2026-02-19
Update JPEG XL support for Chrome Android

ziemek99:
New Chrome version [has just dropped on Android](https://chromereleases.googleblog.com/2026/02/chrome-for-android-update_01493306386.html) and just like the [main Chrome](https://developer.chrome.com/release-notes/145#jpeg_xl_decoding_support_imagejxl_in_blink), it ships JPEG XL behind the flag… I know because I've just checked on my phone 🙂


Fyrd:
Added with the Chrome for Android update, thanks!

PR #7481, opened 2026-02-24
Fixing error in webtransport.json

jensimmons:
We just added support for Safari 26.4 in the JSON file for WebTransport. I made a mistake, not including STP. That's corrected by this PR.


jensimmons:
image
Whoops!

Fyrd:
Thanks!

PR #5789, opened 2021-03-03
Adds workflow file, to run validate against PRs

kennethormandy:
Adds a GitHub Actions workflow file, to run `npm run validate` against Pull Requests.

Sample: https://github.com/kennethormandy/caniuse/runs/2017878290?check_suite_focus=true

I think there used to be CI on the repo, so you might have already decided against this. Feel free to close this if it’s not useful for you.

PR #7496, opened 2026-03-16
Delete .travis.yml

Schweinepriester:
surely it's not used anymore with those old versions of node and [github actions existing](https://github.com/Fyrd/caniuse/blob/main/.github/workflows/main.yml), or…? 😅

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,6 @@
 .gitignore
 .gitattributes
 
-.travis.yml
 validator/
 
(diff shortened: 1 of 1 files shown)


Schweinepriester:
ah, not working anymore: https://github.com/Fyrd/caniuse/pull/5789#issuecomment-791878985

Fyrd:
Good idea, thanks!

PR #7509, opened 2026-04-12
Mozilla position on Web Serial API is neutral

Schweinepriester:
https://mozilla.github.io/standards-positions/#webserial

image

https://github.com/mozilla/standards-positions/issues/336

image

---

originally I arrived here via

- https://mastodon.social/@evilpie@hachyderm.io/116391564494978413
- https://bugzilla.mozilla.org/show_bug.cgi?id=2010930

but looks like it's addon-gated as well currently, so not sure if there's precedent for a note for that case…

### DIFF
--- a/features-json/web-serial.json
+++ b/features-json/web-serial.json
@@ -14,7 +14,7 @@
     },
     {
       "url":"https://mozilla.github.io/standards-positions/#webserial",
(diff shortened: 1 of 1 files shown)


Fyrd:
Thanks!

PR #7519, opened 2026-05-14
Update WebGPU feature support details

quinton-ashley:
It doesn't make any sense to say "Partial support refers to only being enabled by default on macOS 26 Tahoe or later." because Safari 26 isn't available on any other version of macOS, so it's fully supported in Safari 26, not partially supported.


Fyrd:
Per https://developer.apple.com/documentation/safari-release-notes/safari-26-release-notes Safari 26 is also available on macOS Sequoia and macOS Sonoma so the note is still appropriate.

PR #7523, opened 2026-05-22
Firefox 151 and above supports Web Serial

gregstoll:
https://bugzilla.mozilla.org/show_bug.cgi?id=2029625
https://blog.mozilla.org/en/firefox/firefox-web-serial-adafruit/

Thanks!


Fyrd:
Thanks!

PR #7529, opened 2026-06-06
Firefox 151 supports container style queries

### LINKED ISSUE #7528: Firefox 151 supports container style queries
Hi!  Reporting some incomplete/out-of-date feature-support data for Firefox.

caniuse says that Firefox doesn't support container style queries, here:
https://caniuse.com/css-container-queries-style
https://github.com/Fyrd/caniuse/blob/0741550f9cd62d31a3ee600954bd4cef86c47e9a/features-json/css-container-queries-style.json

...but that's out of date -- in fact Firefox enabled the feature in version 151 (current release), here:
https://bugzilla.mozilla.org/show_bug.cgi?id=2030645
Also mentioned in release notes here, if that's a helpful reference:
https://www.firefox.com/en-US/firefox/151.0/releasenotes/

And wpt.fyi shows Firefox release passing nearly all of the interop-2026 tests for this feature:
https://wpt.fyi/results/css/css-conditional/container-queries?label=master&label=stable&aligned&view=interop&q=label%3Ainterop-2026-container-style-queries
(Our only failures are in three tests -- two of the failures are due to incomplete support for an unrelated feature, `attr()`; and the 3rd is due to an edge case, and we recently fixed it in https://bugzilla.mozilla.org/show_bug.cgi?id=2043227 , though that fix hasn't made it to release yet).

For reference, see this related issue in the browser-compat-data repo, for web-feature:container-style-queries, subcategory "Range syntax", where Firefox is incorrectly marked as "Preview" status (but in fact we shipped container-style-queries with range syntax in 151):
https://github.com/mdn/browser-compat-data/issues/29802
(I'm including that link in case browser-compat-data happens to be the source-of-truth for this bit of caniuse data, so that you're aware that the browser-compat-data is incomplete/likely-having-pending-updates for this feature.)

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7528 :)


Fyrd:
Thanks!

PR #7537, opened 2026-06-24
Fix broken resource link from Temporal

### LINKED ISSUE #7535: Broken resource link from Temporal
https://caniuse.com/temporal links to the article https://2ality.com/2021/06/temporal-api.html in its resources section.  This is a 404, and the homepage https://2ality.com/ notes that this may be the case for an extended time period.  The page is available at archive.org (or the link could be removed, or left as-is).

Schweinepriester:
Fixes https://github.com/Fyrd/caniuse/issues/7535

### DIFF
--- a/features-json/temporal.json
+++ b/features-json/temporal.json
@@ -21,8 +21,8 @@
       "title":"WebKit implementation bug"
     },
     {
-      "url":"https://2ality.com/2021/06/temporal-api.html",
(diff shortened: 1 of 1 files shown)


Schweinepriester:
meta:
there are more cases though…
https://github.com/search?q=repo%3AFyrd%2Fcaniuse%202ality&type=code

Fyrd:
Ah yes, thanks for pointing those out! Sad, but will take a look.

PR #7540, opened 2026-06-27
Firefox 154+ is enabling CSS Text Box by default in Nightly only

Schweinepriester:
https://bugzilla.mozilla.org/show_bug.cgi?id=2050138


Fyrd:
Thanks!

PR #7542, opened 2026-07-02
WebGPU is enabled on Linux, for certain configs

captainbrosset:
[This Chromium commit](https://source.chromium.org/chromium/chromium/src/+/0e492e6f38a84951f18d3cb3168b8ed091862738) enabled support for WebGPU on Linux, starting with Chrome 144, at least for certain configs.

@slightlyoff I added a vague note about the partial support, do you think we can be more specific here?


Fyrd:
Thanks!

PR #7546, opened 2026-07-08
update CSS3 `attr()` support, activated in Safari 27

mfranzke:
according to https://bugs.webkit.org/show_bug.cgi?id=26609#:~:text=Enabled%20in%20stable%20as%20of%20https%3A//commits.webkit.org/316672%40main this is now activated in Safari stable (26.6 & 27). Compare to the source code: https://commits.webkit.org/316672@main

~If there might be a 26.6 update previous to the 27 release, I would test for that feature and add another PR if it's already supported in 26.6 than.~
I've tested it in Safari 26.6 and it doesn't work in that release.


Fyrd:
Hmm, I'm not seeing this working yet in the current Safari 27 beta, but will leave this open for now.

mfranzke:
> Hmm, I'm not seeing this working yet in the current Safari 27 beta, but will leave this open for now.

@Fyrd ~I've tested it in Safari 26.6 and it works in that version. I sadly do not have a beta Release of Safari 27 accessible to me, so what would be the preferred way to proceed with this? Should I scope this PR to Safari 26.6 even only, or keep 27 in?~

Schweinepriester:
technically iPadOS, but still not yet active it would appear:

IMG_0013

IMG_0014

mfranzke:
same here with the final Safari 27.0 release, [it doesn't work with the simple test case](https://tests.caniuse.com/?feat=css3-attr). I'll close this for now; as it does work in Safari TP, it might be either just a matter of minor releases or worse (even though that it was merged to the stable channel).

PR #7548, opened 2026-07-10
Return Firefox 154+ is enabling CSS Text Box by default in Nightly only

Schweinepriester:
the changes from https://github.com/Fyrd/caniuse/pull/7540 were also lost 😢


Fyrd:
Oof, sorry about that 😬

PR #7549, opened 2026-07-15
Firefox 154 is shipping CSS Text Box

Schweinepriester:
- https://groups.google.com/a/mozilla.org/g/dev-platform/c/yCAFcvQ-Tkc/m/vUnKuduqBAAJ
- https://bugzilla.mozilla.org/show_bug.cgi?id=2050141

PR #7550, opened 2026-07-15
Fix validation error

Schweinepriester:
as realized here: https://github.com/Fyrd/caniuse/pull/7549#issuecomment-4983440943


Fyrd:
Thanks!

PR #7555, opened 2026-07-20
Update push-api.json

jensimmons:
The legacy Safari Push Notifications is not supported in Safari 26.4 and later.


Fyrd:
Thanks!

PR #7569, opened 2026-08-31
Small improvements to jpegxl.json

goodusername123:
- Switched to using [non-breaking spaces](https://jkorpela.fi/chars/spaces.html#:~:text=NO%2DBREAK%20SPACE) in the main title and description for the spelling of the name of the image format in question (link/resource titles are unchanged but maybe they should be?).
- Mention "JXL" since it's one of the main valid ways to refer to JPEG XL (It's the file extension + some people including core devs of JPEG XL also use it as a preferred term/name). This attempts to somewhat match how some other caniuse pages go about this such as https://caniuse.com/apng and https://caniuse.com/es6 only thing I'm unsure of is is if "(JXL)" should be directly after "JPEG XL" or all the way at the end of the title.
- Add additional relevant resources/links (the "In-depth overview" document might be somewhat redundant since it's already linked on the ["spec"/"Specification" marked page](https://jpeg.org/jpegxl/documentation.html) so feel free to remove the "In-depth overview" resource/link if so).

image showing a before and after for how the main title and description should look after this change:
BeforeAndAfter-jpegxl json-caniuse


Fyrd:
Thanks!

PR #7572, opened 2026-09-11
Safari TP added support for the unprefixed `user-select` CSS property

Schweinepriester:
https://webkit.org/blog/18304/release-notes-for-safari-technology-preview-252/

> -   Added support for the unprefixed `user-select` CSS property. ([319562@main](https://commits.webkit.org/319562@main)) (184964807)


Fyrd:
Thanks!

